PR #18, opened 2025-05-20
feat: add Gmail message sending functionality with MIMEText support

valdezm:
Adds simple sending of email using gmail tool `send_gmail_message`
image

PR #19, opened 2025-05-20
feat: add Gmail message sending functionality with MIMEText support #18

taylorwilsdon:
ref https://github.com/taylorwilsdon/google_workspace_mcp/pull/18


taylorwilsdon:
@valdezm I was totally wrong about the from header, TIL - been a while since I built a send object I guess. Tested and all is looking good, merging - thanks again!

PR #32, opened 2025-05-25
feat: enhance Gmail search with XML structure and web interface links

MariusWilsch:
- Add _generate_gmail_web_url() helper function using #all for universal Gmail access
- Update search_gmail_messages to return XML-structured output with index attributes
- Include clickable Gmail web interface URLs for both messages and threads
- Maintain backward compatibility with existing Message/Thread ID usage
- Improve manual verification capabilities for users when multiple threads have same subject

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -78,6 +78,39 @@ def _generate_gmail_web_url(item_id: str, account_index: int = 0) -> str:
     return f"https://mail.google.com/mail/u/{account_index}/#all/{item_id}"
 
 
+def _format_gmail_results_plain(messages: list, query: str) -> str:
+    """Format Gmail search results in clean, LLM-friendly plain text."""
+    if not messages:
+        return f"No messages found for query: '{query}'"
+
+    lines = [
+        f"Found {len(messages)} messages matching '{query}':",
+        "",
+        "📧 MESSAGES:",
+    ]
+
+    for i, msg in enumerate(messages, 1):
+        message_url = _generate_gmail_web_url(msg["id"])
+        thread_url = _generate_gmail_web_url(msg["threadId"])
+
+        lines.extend([
+            f"  {i}. Message ID: {msg['id']}",
+            f"     Web Link: {message_url}",
+            f"     Thread ID: {msg['threadId']}",
+            f"     Thread Link: {thread_url}",
+            ""
+        ])
+
+    lines.extend([
+        "💡 USAGE:",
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
Tested, I really like this! Great add, thanks - having clickable unsubscribe is awesome:

image

taylorwilsdon:
Not sure why XML though, let me rewrite this for tool calling compatibility, xml is a design anti-pattern for LLM tool calling - we don't need structured markup here and it just adds complexity especially if you ask it to return results in markdown in your system prompt like I do 😃  Hang tight, will have it in 2 seconds

taylorwilsdon:
Ok - set in [3c00274](https://github.com/taylorwilsdon/google_workspace_mcp/pull/32/commits/3c00274e3f666fb47de30bd154869a60522c0d9a) should be identical functionality-wise!

MariusWilsch:
Well I strongly disagree with this one. XML, is the one pattern imo which
allows Claude's performance to be so much better than most other LLM's

See article here:
https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/use-xml-tags

Am Di., 27. Mai 2025 um 05:51 Uhr schrieb Taylor Wilsdon <
***@***.***>:

MariusWilsch:
> Tested, I really like this! Great add, thanks - having clickable unsubscribe is awesome:
> 
> image

But thanks man!

taylorwilsdon:
I've played with that but my understanding is that this is intended for structuring prompts coming in rather than tool calling responses based on how anthropic implemented native tool calling - using a structured input ie xml will help Claude parse instructions clearly and break up information categorically ie context vs task vs constraints. For tool return values, the model doesn't need help parsing its own tool outputs since it generated the calls and knows what to expect in the response, since the tool outputs are typically consumed programmatically or displayed straight shot rather than used as multi-step instructions. Open to being wrong, but I haven't seen any convincing examples of a better (or even different) output with xml vs plaintext in a tool call response.

MariusWilsch:
Okay, that’s a valid argument. I I just see it from this perspective is
that there may be not an apparent instant benefit however in general the AI
Understand it better. This guy is the level of error is so small but this
matter too much anyways. Just happy that’s in there now.

Viele Grüße / Best regards

Marius Wilsch
Co-Founder

Marius Wilsch
------------------------------
Veloxforce OÜ | Järvevana tee 9-40, 11314, Tallinn, Estonia
Registration code: 17029678 | VAT: EE102766828

Issue #58, opened 2025-06-12
tool proposal: doc from template


intval:
I would like to propose a new tool: create copy of document from template with variable interpolation

**Usecase:** 
there is some template for a contract with text like `contract date {{date}}` , I want an agent to copy that template and insert some variables in place of placeholders.

**Proposed Signature**:
```
@server.tool()
@require_multiple_services([
    {"service_type": "drive", "scopes": "drive_read", "param_name": "drive_service"},
    {"service_type": "docs", "scopes": "docs_read", "param_name": "docs_service"}
])
async def copy_template_tool(
    input: "CopyTemplateInput",
    drive_service,
    docs_service
):
    # Your tool implementation here
    pass


class CopyTemplateInput(BaseModel):
    templateId: str = Field(..., description="ID of the template document to copy from.")
    newTitle: str = Field(..., min_length=1, description="Title for the new document.")
    parentFolderId: Optional[str] = Field(None, description="ID of folder where document should be created. If not provided, creates in Drive root.")
    replacements: Optional[Dict[str, str]] = Field(None, description='Key-value pairs for text replacements in the template (e.g., {"{{NAME}}": "John Doe", "{{DATE}}": "2024-01-01"}).')
```

**Implementation consideration:**
1. Most likely reading the document and extracting variables beforehand also needs to be implemented
2. To keep doc structure (e.g. tables, headers, footers) it will need to read it the document in it's raw api structure (collection of json paragraph object)

** This issue **
I'm happy to contribute it, as i'll develop it for myself. Would you be interested in accepting this as contribution ?

taylorwilsdon:
Oooh very interesting! I could see some valuable use cases here, even operating as an LLM driven mail merge that sends the template-filled content as an email or generates a drive file with it filled. Definitely open to this 👍

tabletenniser:
+1 to this, any ETA on when can this be done?  Instead of a completely separate tool, maybe we can just update the existing `create_doc` tool to also take a template?

PR #60, opened 2025-06-15
Two new tools: copy doc and insert template variables + support for service account credentials

intval:
This addresses issuen https://github.com/taylorwilsdon/google_workspace_mcp/issues/58


.gitignore
```diff
@@ -13,11 +14,11 @@ dist/
 .venv/
 venv/
 .idea/
-.vscode/
+mcp_server_debug.log
```
intval, line 16: Dropped vscode so we can commit launch json and mcp.json
the first one is a standard, the second one is more for reference of people who want to use this MCP directly from vscode or cursor

taylorwilsdon:
Thanks for this! Will build and take a look today

.gitignore
```diff
@@ -2,6 +2,7 @@
 __pycache__/
 *.py[cod]
 *.so
+uv.lock
```
taylorwilsdon: per astral uv.lock should be committed to source https://docs.astral.sh/uv/guides/projects/#uvlock

```suggestion
```

.gitignore
```diff
@@ -13,11 +14,11 @@ dist/
...
-.vscode/
+mcp_server_debug.log
 
 # ---- macOS clutter -----------------------------------------------------
 .DS_Store
 
 # ---- Secrets -----------------------------------------------------------
 client_secret.json
-
+.credentials/*.json*
```
taylorwilsdon, line 23: ++++

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Run in Debug Mode",
```
taylorwilsdon, line 5: Love the idea of adding debug mode and including preconfigured json, although I'm thinking it might make more sense to launch as a self contained stdio rather than spinning off a streamable http instance - open to being wrong though!

.vscode/mcp.json
```diff
@@ -0,0 +1,9 @@
+{
+"servers": {
+     "google_workspace": {
+        "type": "stdio",
+        "command": "cmd",
+        "args": ["/c", "cd", "google_workspace_mcp", "&&", "uv", "run", "main.py", "--single-user", "--tools", "docs"]
```
taylorwilsdon: This is specific to your windows installation so we'll need to rewrite - thinking maybe we can just use `uvx workspace-mcp` for general plug & play

auth/google_auth.py
```diff
@@ -56,8 +57,25 @@ def _find_any_credentials(base_dir: str = DEFAULT_CREDENTIALS_DIR) -> Optional[C
         if filename.endswith('.json'):
             filepath = os.path.join(base_dir, filename)
             try:
+                # Check if this is a service account file
+                if "iam.gserviceaccount.com" in filename:
+                    logger.info(f"[single-user] Found service account file: {filepath}")
+                    try:
+                        from google.oauth2 import service_account
```
taylorwilsdon: nit - I try to always avoid inline imports, the performance overhead is negligible and the timebomb risk of code appearing to work at runtime and then crashing when this function is invoked outweighs it. can push a refac up to branch

auth/google_auth.py
```diff
@@ -525,8 +542,12 @@ async def get_authenticated_google_service(
...
-        # Try to get email from credentials if needed for validation
-        if credentials and credentials.id_token:
+        # For service accounts, use the service account email
+        if hasattr(credentials, 'service_account_email'):
+            log_user_email = credentials.service_account_email
+            logger.info(f"[{tool_name}] Using service account: {log_user_email}")
+        # For OAuth2 credentials, try to get email from id_token
+        elif credentials and hasattr(credentials, 'id_token') and credentials.id_token:
             try:
                 import jwt
```
taylorwilsdon: need to fix my own nit here 😄

uv.lock
```diff
@@ -24,75 +24,132 @@ dependencies = [
...
-    { url = "https://files.pythonhosted.org/packages/e9/b0/0200da600134e001d91851ddc797809e2fe0ea72de90e09bec5a2fbdaccb/charset_normalizer-3.4.2-cp313-cp313-win_amd64.whl", hash = "sha256:aa6af9e7d59f9c12b33ae4e9450619cf2488e2bbe9b44030905877f0b2324980", size = 105641, upload_time = "2025-05-02T08:33:18.753Z" },
-    { url = "https://files.pythonhosted.org/packages/20/94/c5790835a017658cbfabd07f3bfb549140c3ac458cfc196323996b10095a/charset_normalizer-3.4.2-py3-none-any.whl", hash = "sha256:7f56930ab0abd1c45cd15be65cc741c28b1c9a34876ce8c17a2fa107810c0af0", size = 52626, upload_time = "2025-05-02T08:34:40.053Z" },
+sdist = { url = "https://files.pythonhosted.org/packages/e4/33/89c2ced2b67d1c2a61c19c6751aa8902d46ce3dacb23600a283619f5a12d/charset_normalizer-3.4.2.tar.gz", hash = "sha256:5baececa9ecba31eff645232d59845c07aa030f0c81ee70184a90d35099a0e63", size = 126367, upload-time = "2025-05-02T08:34:42.01Z" }
```
taylorwilsdon: What version of uv are you using? I've never seen `upload-time` before, only `upload_time` (introduced in v0.6.15) or nothing - curious what else is floating around out there

taylorwilsdon (commented):
Very much appreciate the PR, code looks clean and high quality in general. Have a few todos as I test now, but will get this merged today.

taylorwilsdon:
Hey @intval - I tested it out, looks pretty good to me! Do you imagine folks use the copy functionality without the templated aspect and to leave that standalone, or should we instruct the LLM to trigger the replace text mechanism in step? 
image
image
image

FWIW - the current state can already accomplish the copy without the new function:
image
So I'd say we either link the copy & replace bits together for a chained action or drop the `copy_google_doc` logic since we can accomplish that without it. Let me know what you think - t

.gitignore
taylorwilsdon: ++ love that, updated mcp.json to be platform and path agnostic (looked to be hardcoded for windows cmd + c:\workspacemcp)

.vscode/launch.json
taylorwilsdon: Will leave as streamable http for now

taylorwilsdon:
Hey @intval - just wanted to ping one more time to see if ya had any thoughts on the above before moving forward here!

.vscode/mcp.json
intval: Yes, with a small caveat, 
`cmd /c` is not the `cd c:/`, it's actually an argument to `cmd` to start a command, a bit like `bash -c 'echo hi'`  
`cmd` is windows specific, but i've comitted it intentionally, because it took me a while to figure out how to actually run it on windows. I'd like to keep this info somewhere, though we can also move it readme and just keep here everything from run onwards

uv.lock
intval:
```$
uv --version
uv 0.7.4 (6fbcd09b5 2025-05-15)
```

I'm not coming from the python world, so this all seemed just legit ;)

intval:
Ad merging copy functionality with the templated:  

I think it's better to keep them separate.  
Copying a document seems like a valid operation in it's own stead
I ended up using it on it's own as well.  

I also think that templating can be an op on it's own, since it may be a multi step approach, 
e.g. first, replace this bunch, show the user, then replace another bunch of variables.  

I suppose in order to support both agents and humans, it would make sense to add the third method, that combines the two and keep the existing two as separate instances. This will allow more granular agentic flow, like the one i use:
a. read document content to identify placeholders
b. copy the document
c. maybe append to it more text as append_text
b. obtain via rag values to insert for placeholders read by (a) or freshly added by (c)
d. implement placeholder replacement

auth/google_auth.py
```diff
@@ -56,8 +60,24 @@ def _find_any_credentials(base_dir: str = DEFAULT_CREDENTIALS_DIR) -> Optional[C
...
+                    try:
+                        credentials = service_account.Credentials.from_service_account_file(filepath,scopes=SCOPES)
+                        logger.info(f"[single-user] Successfully loaded service account credentials from {filepath}")
+
+                        # Force refresh to get a token, since by default it's not set and the library considers then credentials to be invalid
+                        credentials.refresh(Request())
+                        return credentials
+                    except Exception as e:
+                        logger.warning(f"[single-user] Error loading service account credentials from {filepath}: {e}")
+                        continue
```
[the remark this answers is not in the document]
intval: That's actually correct, i pondered about it for a moment and decided that it's a tradeoff between this and reading the file from disk twice, and decided that for single user configuration it will do.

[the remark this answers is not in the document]
taylorwilsdon: Hey @intval - hm, when I create a service account key it does stub out an email with iam.gserviceaccount.com in it, but the actual generated keyfile name is `open-webui-444821-b145f5cb467e.json` which would not match this check. A user would have to manually rename the key for this logic to work. Thoughts on splitting the service account PR out from the copy & template bits so we can get that merged and figure out the best approach for service accounts separately?

taylorwilsdon (changes requested):

kapilthakare-cyberpunk:

PR #63, opened 2025-06-16
Feature/calendar ends info added

nacholupotti:
Added the end date as extra info for each event, on the get_events tool response


taylorwilsdon:
Love it, great addition @nacholupotti - small detail that goes a long way! Updated a few little bits (docstring & variable name) but otherwise lgtm, thank you! 
image

taylorwilsdon (approved):

PR #65, opened 2025-06-18
Explain GOOGLE_CLIENT_SECRETS in README.md

telenieko:
Currently the only way to find out about configuring the secrets location is by reading the code at https://github.com/taylorwilsdon/google_workspace_mcp/blob/ec480a2aee3eaa76497d9bb75e9269e0594db428/auth/google_auth.py#L30-L39

So, just add a quick note in the relevant part of the README.md


taylorwilsdon:
ah ++ good callout, thanks!

PR #66, opened 2025-06-18
feat: add support for file attachments in event creation and enhance …

nacholupotti:
PR: Google Calendar & Drive Tools – Attachments, File Uploads

gcalendar (Google Calendar tools)
Event Attachments:
Added support for adding Google Drive file attachments to calendar events.
Accepts attachments as a list or a comma-separated string; robust file ID extraction from all Google Drive URL formats.
Ensures supportsAttachments=True is passed correctly to the API when attachments are added

gdrive (Google Drive tools)
File Upload Improvements:
I added the ability to upload fileUrl instead of sending the content.
This reduces a lot the quantity of tokens the model needs to send (since fileUrl is shorter than content)
Both content and fileUrl are now optional; one of both must be present
If both are provided, fileUrl is preferred.
When uploading from fileUrl, the MIME type is inferred from the HTTP Content-Type header if available and specific.
Type Hints & Validation:
Updated function signatures to use Optional[str] for content and fileUrl for better compatibility with validation frameworks.

Summary:
This PR adds support for Google Drive attachments in calendar events, improves file upload flexibility and MIME type detection in creating drive files.

### DIFF
--- a/gcalendar/calendar_tools.py
+++ b/gcalendar/calendar_tools.py
@@ -97,7 +97,7 @@ async def list_calendars(service, user_google_email: str) -> str:
     logger.info(f"[list_calendars] Invoked. Email: '{user_google_email}'")
 
     calendar_list_response = await asyncio.to_thread(
-        service.calendarList().list().execute
+        lambda: service.calendarList().list().execute()
     )
     items = calendar_list_response.get("items", [])
     if not items:
@@ -169,7 +169,7 @@ async def get_events(
     )
 
     events_result = await asyncio.to_thread(
-        service.events()
+        lambda: service.events()
         .list(
             calendarId=calendar_id,
             timeMin=effective_time_min,
@@ -178,7 +178,7 @@ async def get_events(
             singleEvents=True,
             orderBy="startTime",
         )
-        .execute
+        .execute()
     )
     items = events_result.get("items", [])
     if not items:
@@ -296,7 +296,7 @@ async def create_event(
                 if drive_service:
                     try:
                         file_metadata = await asyncio.to_thread(
-                            drive_service.files().get(fileId=file_id, fields="mimeType,name").execute
+                            lambda: drive_service.files().get(fileId=file_id, fields="mimeType,name").execute()
                         )
                         mime_type = file_metadata.get("mimeType", mime_type)
                         filename = file_metadata.get("name")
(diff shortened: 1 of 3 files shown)


taylorwilsdon:
This looks awesome, thank you for putting this together - in theory the get_events function can already call by event id, but I do think there's value in having an explicit one that is likely faster because it's not executing a query and then extracting a response. Will get this tested and merged, thanks!

gcalendar/calendar_tools.py
```diff
@@ -276,16 +283,55 @@ async def create_event(
...
+                lambda: service.events().insert(
+                    calendarId=calendar_id, body=event_body, supportsAttachments=True
+                ).execute()
+            )
+        else:
+            created_event = await asyncio.to_thread(
+                service.events().insert(calendarId=calendar_id, body=event_body).execute
+            )
         link = created_event.get("htmlLink", "No link available")
-        confirmation_message = f"Successfully created event '{created_event.get('summary', summary)}' for {user_google_email}. Link: {link}"
```
taylorwilsdon:
```
"error": "HTTP error! Status: 500. Message: {\"detail\":{\"message\":\"Unexpected error\",\"error\":\"500: {'message': \\\"Error executing tool create_event: Unexpected error creating event: name 'confirmation_message' is not defined.\\\"}\"}}"
```

taylorwilsdon: will fix

gcalendar/calendar_tools.py
```diff
@@ -276,16 +283,55 @@ async def create_event(
...
+                        try:
+                            file_metadata = await asyncio.to_thread(
+                                drive_service.files().get(fileId=file_id, fields="mimeType").execute
+                            )
+                            mime_type = file_metadata.get("mimeType", mime_type)
+                        except Exception as e:
+                            logger.warning(f"Could not fetch MIME type for file {file_id}: {e}")
+                    event_body["attachments"].append({
+                        "fileUrl": file_url,
+                        "title": f"Drive Attachment ({file_id})",
```
taylorwilsdon: Maybe we want to extract the actual file name rather than Drive Attachment (uuid) 
image

taylorwilsdon:
Ok @nacholupotti made a bunch of changes and merged the exception handler decorator bits but it looks great thanks for putting this together! 

Dropped aiohttp we already have httpx used actual file name instead of placeholder some other little quality of life but good to merge - 

Screenshot 2025-06-18 at 7 24 56 PM

durandom:
shouldnt we also support fetching attachments via `get_events` ?

PR #82, opened 2025-07-03
feat: get user google email from environment var

juansepl:
This PR implements the ability to set user_google_email using an environment variable. Previously, this value might have been hardcoded or required explicit input. Now, if a relevant environment variable is detected, its value will be used as the default for user_google_email. This provides an optional, more convenient, and centralized way to manage the Google email address. It improves configurability and reduces potential for errors from manual entry especially during execution time in the server checking the authentication.


taylorwilsdon:
Ooh I like this a lot. When I'm running single email I bake it into system prompts but there are some clients where that's inelegant at best, really good idea. Let me get this reviewed and we'll get it in 

edit - wait read the actual pr, simple and elegant hah.. We should rip some quick docs and inject it a few other places to make it smoother but I will have this out the door tonight or tomorrow.

PR #84, opened 2025-07-04
Update manifest.json

TamimiGitHub:
When setting the `OAUTHLIB_INSECURE_TRANSPORT`,  application throws the following error
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for StdioServerParameters
env.OAUTHLIB_INSECURE_TRANSPORT
  Input should be a valid string [type=string_type, input_value=1, input_type=int]
    For further information visit https://errors.pydantic.dev/2.11/v/string_type
```

Since the `OAUTHLIB_INSECURE_TRANSPORT` env variable is an int the type should be reflected accordingly

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -112,7 +112,7 @@
       "default": "8000"
     },
     "OAUTHLIB_INSECURE_TRANSPORT": {
-      "type": "string",
+      "type": "int",
       "title": "OAuth Insecure Transport",
       "description": "Allow OAuth over HTTP for development (set to '1' for development only)",
       "required": false,


taylorwilsdon:
Ah good catch - I've been unimpressed with the dxt library so far and don't have it functional but this is a good step. Will review and play with that again today.

taylorwilsdon:
Er, wait `- user_config.OAUTHLIB_INSECURE_TRANSPORT.type: Invalid enum value. Expected 'string' | 'number' | 'boolean' | 'directory' | 'file', received 'int'` - it looks like it wants "number" not "int" here

taylorwilsdon:
Fixed and will upstream, dxt is now working

PR #98, opened 2025-07-12
Add color and own attending information to calendar event

marekkalnik:
Two information I find useful :

- color (for people who use colors as a way to label events)
- attending myself:
   - if multiple invitees, it looks at own response
   - if no invitees, it is "attending" by default

I find that there is a tradeoff between completeness of information and response size, but those two are very useful for my case. Maybe the response should be parametrised to avoid information overload?

### DIFF
--- a/gcalendar/calendar_tools.py
+++ b/gcalendar/calendar_tools.py
@@ -77,7 +77,6 @@ def _correct_time_format_for_api(
     logger.info(f"{param_name} '{time_str}' doesn't need formatting, using as is.")
     return time_str
 
-
 @server.tool()
 @require_google_service("calendar", "calendar_read")
 @handle_http_errors("list_calendars")
@@ -110,6 +109,7 @@ async def list_calendars(service, user_google_email: str) -> str:
     )
     logger.info(f"Successfully listed {len(items)} calendars for {user_google_email}.")
     return text_output
+
 
 
 @server.tool()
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
Clever, love this! Will give it a try today

taylorwilsdon:
Ok, tested it out - my only note is that this only applies on `get_event` not `get_events` so you won't get it when just asking for a summary of your calendar for the day / week etc. I'm also not finding that Claude actually utilizes the information in its response. Do you have a screenshot showing desired output? 

Here's what I get, it tells me the color correctly if I ask specifically but doesn't use it otherwise:
image

marekkalnik:
Hi, nice catch, I didn't realize initially that those could be added to event_list.

As for the color, Google API does not return the textual label for a color. It only returns hex values which is not very useful in my opinion.

The way I solved the problem for myself is putting it in my project prompt in Claude.

image

So I can use it this way:
image

taylorwilsdon:
Hm, I think in general anything that requires a custom system prompt probably isn't a good fit as a base tool - if we can find a way to incorporate it more dynamically into responses so that any arbitrary LLM can do something useful with it I'm totally open to it!

marekkalnik:
Yes, I understand. But this is currently not possible with Google APIs, textual information about event colors is simply not available.

Do you want me to separate the "attending" and "color" features so you can merge only the first one ?

Issue #99, opened 2025-07-14
SSL Error in gmail_tools.py


jihunkim0:
The get_gmail_thread_content function in gmail/gmail_tools.py is failing with the following error:
```
Error executing tool get_gmail_thread_content: An unexpected error occurred in get_gmail_thread_content: [SSL] record layer failure (_ssl.c:2580)
```
This is likely a transient network issue. A retry mechanism with exponential backoff should be implemented to make the function more resilient.

PR #100, opened 2025-07-14
Fix: SSL Error in gmail_tools.py

jihunkim0:
This PR fixes issue #99 by adding a retry mechanism with exponential backoff to the `get_gmail_thread_content` function in `gmail/gmail_tools.py`. This should make the function more resilient to transient SSL errors.


taylorwilsdon:
This looks clean, let me take a look but I think we should be able to get this merged - thanks!

taylorwilsdon:
The approach is directionally good, but I feel like taping this on raises idempotency risks on write operations - we may miss SSL failures arising before the wrapped function body likeduring service construction, and it could mask root-cause context when combined with handle_http_errors. I'm thinking the best approach is to refactor this into the existing handle_http_error decorator. Let me see what I can do tonight @jihunkim0

taylorwilsdon:
Refactored in https://github.com/taylorwilsdon/google_workspace_mcp/pull/106 thanks!

Issue #129, opened 2025-07-28
moving event will lost event info


gyliu513:
- step1: Booked an event at 1:00PM EST with title as test
- step2: Move the event test to 3:00PM EST
- result: event moved successfully, but title lost

taylorwilsdon:
<img width="638" height="378" alt="Image" src="https://github.com/user-attachments/assets/e8299842-5b4e-4dec-9fec-8db55b950694" />

<img width="635" height="417" alt="Image" src="https://github.com/user-attachments/assets/ba41c535-ff70-4993-b420-6d5a59af0f28" />

<img width="550" height="308" alt="Image" src="https://github.com/user-attachments/assets/5936c252-ea30-4a83-9b30-efdfa46edf7c" />

Good catch, I am able to reproduce. Will fix!

taylorwilsdon:
Ships with next release

PR #132, opened 2025-07-28
Keep original event info when update

gyliu513:
Fixed https://github.com/taylorwilsdon/google_workspace_mcp/issues/129

@taylorwilsdon ^^


gyliu513:
Seems you already created a PR, thanks @taylorwilsdon !

PR #144, opened 2025-08-06
feat: Enable USER_GOOGLE_EMAIL configuration in manifest.json for DXT users

murphybread:
Hi Taylor,
First, thank you for creating such a useful project.

While setting up the extension on Claude Desktop using the .dxt file, I noticed that the authentication process couldn't start because it requires the user's email, but there was no field to configure it in the extension settings.

It seems the USER_GOOGLE_EMAIL configuration was missing from manifest.json. Since this is a required variable for the auth flow to identify the user, I've added it to the user_config schema and the mcp_config.env map.

With this change, a field for the user's email now correctly appears in the Claude settings, and the authentication process works as expected. I've tested this by repacking the .dxt with the updated manifest, and it resolves the setup issue.

I thought this might be a helpful fix for others as well. Let me know if you have any feedback.
Thanks!

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -28,6 +28,7 @@
       "env": {
         "GOOGLE_OAUTH_CLIENT_ID": "${user_config.GOOGLE_OAUTH_CLIENT_ID}",
         "GOOGLE_OAUTH_CLIENT_SECRET": "${user_config.GOOGLE_OAUTH_CLIENT_SECRET}",
+        "USER_GOOGLE_EMAIL": "${user_config.USER_GOOGLE_EMAIL}",
         "GOOGLE_OAUTH_REDIRECT_URI": "${user_config.GOOGLE_OAUTH_REDIRECT_URI}",
         "GOOGLE_CLIENT_SECRET_PATH": "${user_config.GOOGLE_CLIENT_SECRET_PATH}",
         "GOOGLE_CLIENT_SECRETS": "${user_config.GOOGLE_CLIENT_SECRETS}",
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
Hey there, good catch! For what it's worth, it's not required so we'll want to swap that to optional (lots of people use multiple accounts in one instances, including myself haha) but seems like a great inclusion and I will repack the dxt after merge. Thanks!

manifest.json
```diff
@@ -123,6 +124,16 @@
         "min_length": 24
       }
     },
+    "USER_GOOGLE_EMAIL": {
+      "type": "string",
+      "title": "User Google Email",
+      "description": "Your Google email address for authentication (e.g., your.email@gmail.com)",
+      "required": true,
```
taylorwilsdon:
```suggestion
      "required": false,
```

manifest.json
```diff
@@ -123,6 +124,16 @@
         "min_length": 24
       }
     },
+    "USER_GOOGLE_EMAIL": {
+      "type": "string",
+      "title": "User Google Email",
+      "description": "Your Google email address for authentication (e.g., your.email@gmail.com)",
```
taylorwilsdon:
```suggestion
      "description": "Optional default email for authentication flows. If set, the LLM won't need to specify your email when calling start_google_auth",
```

murphybread:
Ah, that makes perfect sense. Thank you for the clarification!
My reasoning for setting it to required was from a single-user perspective, but I hadn't considered the multi-account use case you mentioned.
Thanks again for the great open-source project. It's reading my Google Drive and Calendar perfectly now!

manifest.json
```diff
@@ -123,6 +124,16 @@
...
       }
     },
+    "USER_GOOGLE_EMAIL": {
+      "type": "string",
+      "title": "User Google Email",
+      "description": "Optional default email for authentication flows. If set, the LLM won't need to specify your email when calling start_google_auth",
+      "required": false,
+      "sensitive": false,
+      "validation": {
+        "pattern": "^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$"
```
taylorwilsdon:
```suggestion
        "pattern": "^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$"
```

taylorwilsdon: this will fix the validation for emails with dashes in them

taylorwilsdon (changes requested):

taylorwilsdon:
Merged and fixed regex, thank you!

PR #149, opened 2025-08-07
Add reminder support to Google Calendar MCP

benjaminjackson:
## Changes

- Add reminder parameters to create_event and modify_event functions
- Support popup and email reminder methods with configurable timing
- Validate reminder format, method, and timing constraints
- Maximum 5 reminders per event, 0-40320 minutes range

## Implementation Notes

_Note: There appears to be a parameter parsing bug in Claude Code’s MCP integration that prevents complex data structures from being properly converted from XML strings to JSON objects, so we are passing reminders as a JSON string instead of a first-class typed array._

### DIFF
--- a/gcalendar/calendar_tools.py
+++ b/gcalendar/calendar_tools.py
@@ -9,6 +9,7 @@ import logging
 import asyncio
 import re
 import uuid
+import json
 from typing import List, Optional, Dict, Any
 
 from googleapiclient.errors import HttpError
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
Thanks! Looks good, merged in main changes and will get this pushed up today.

benjaminjackson:
Great call on the default handling 👍

PR #151, opened 2025-08-08
attachment listing,download and reading from gmail inbox functionality added 

Mustafa-tech-20:
Added Gmail Attachment Handling Functionality

Implemented listing of email attachments from the Gmail inbox

Added functionality to download attachments

Enabled reading and processing of attachment content


taylorwilsdon:
Love the idea, thanks for the PR! I'm a bit concerned about how much overhead this adds though - it's introducing a huge number of new and in some cases (ie pandas) very heavy dependencies. Also looks like some redundant code. Maybe a refac is in order?

lucasmocellin:
@Mustafa-tech-20 do you intend to refactor this as @taylorwilsdon suggested?

this would be a huge add to this MCP.

PR #156, opened 2025-08-10
fix: handle JSON string input for modify_sheet_values parameter

yairwein:
Fixes type mismatch error preventing MCP clients from calling modify_sheet_values.

## Problem
FastMCP 2.11.1 passes complex nested types (like List[List[str]]) as JSON strings through the MCP protocol, but the schema validation expected the actual Python type. This caused validation failures with error: '[["Task"]]' is not valid under any of the given schemas.

## Solution
- Accept both str and List[List[str]] types for the values parameter
- Parse JSON string input to Python list when received from MCP clients
- Validate structure is a proper 2D array (list of lists)
- Maintain backward compatibility with direct Python list input

## Changes
- Added json import and Union type hint
- Modified values parameter to accept Union[str, List[List[str]]]
- Added JSON parsing logic with proper validation
- Updated docstring to document both input formats

This follows the proven pattern used in calendar_tools.py for handling reminders.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/gsheets/sheets_tools.py
+++ b/gsheets/sheets_tools.py
@@ -6,7 +6,8 @@ This module provides MCP tools for interacting with Google Sheets API.
 
 import logging
 import asyncio
-from typing import List, Optional
+import json
+from typing import List, Optional, Union
 
 
 from auth.service_decorator import require_google_service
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
Looks sensible and well approached, thanks @yairwein!

PR #137, opened 2025-08-02
feat: Full Client to Server OAuth 2.1 Implementation - 06/18 MCP Spec

taylorwilsdon:
### feat(auth): Implement OAuth 2.1 for Multi-User Support

This pull request introduces robust **OAuth 2.1 bearer token authentication**, enabling secure, multi-user support for the MCP server. This is a significant enhancement that allows the server to be used as a backend for web applications and other clients where multiple users need to authenticate with their own Google accounts.

The core of this implementation is a new authentication middleware and an innovative CORS proxy architecture that solves common issues with browser-based Google OAuth flows.

### Key Changes:

*   **OAuth 2.1 Bearer Token Support**: The server now accepts Google OAuth 2.1 access tokens via the `Authorization: Bearer` header, allowing for per-request authentication from multiple users.
*   **New Authentication Middleware**: A new [`auth/auth_info_middleware.py`](auth/auth_info_middleware.py:69) has been added. This middleware intercepts incoming requests, validates the bearer token, and populates the `fastmcp_context` with the authenticated user's identity.
*  FastMCP 2.11.1 RemoteAuthProvider: Leverages the latest FastMCP authentication patterns for standards-compliant OAuth
  2.1
  - Multi-User Sessions: Each user's credentials are isolated and securely managed
  - Immutable Session Binding: Once a session is bound to a user, it cannot be hijacked or reassigned

*   **Innovative CORS Proxy**: To overcome Google's lack of CORS headers on its OAuth endpoints, this PR implements an intelligent proxy for the authorization server discovery and token exchange endpoints. This allows browser-based clients to complete the OAuth flow seamlessly.
*   **Simplified Configuration**: The new OAuth 2.1 flow automatically reuses the existing [`GOOGLE_OAUTH_CLIENT_ID`](README.md:20) and [`GOOGLE_OAUTH_CLIENT_SECRET`](README.md:21) environment variables, requiring no extra configuration to enable.
*   **Updated Documentation**: The [`README.md`](README.md) has been extensively updated to document the new OAuth 2.1 capabilities, including when to use it and how the CORS proxy works.

Solves two critical challenges for browser-based OAuth:
  1. Dynamic Client Registration: Proxies registration requests to work around Google's lack of OAuth 2.1 support
  2. CORS Headers: Adds proper CORS headers to Google's OAuth endpoints, enabling browser-based authentication
Closes https://github.com/taylorwilsdon/google_workspace_mcp/issues/112

[...]

PR #160, opened 2025-08-10
fix: always use email from jwt token

shawnzhu:
I like the new `auth_info_middleware.py`, this is about using the email from jwt token as value of `user_google_email` for tool calls.

In the multi-user environment, the use case is avoiding requests authenticated with JWT token to perform tool call on-behalf of arbitrary user. E.g., user `foo@example.com` get autenticated over jwt, then it could search emails owned by `bar@example.com` if that user has been authenticated already.

### DIFF
--- a/auth/auth_info_middleware.py
+++ b/auth/auth_info_middleware.py
@@ -241,6 +241,18 @@ class AuthInfoMiddleware(Middleware):
         
         try:
             await self._process_request_for_auth(context)
+
+            # if authenticated_user_email exists in fastmcp_context and authenticated_via is jwt_token,
+            # uses authenticated_user_email as the value of the argument "user_google_email"
+            authenticated_user_email = context.fastmcp_context.get_state("authenticated_user_email")
+            authenticated_via = context.fastmcp_context.get_state("authenticated_via")
+
+            if authenticated_user_email and authenticated_via == "jwt_token":
+                # Check if this is a tool call with arguments
+                if hasattr(context, 'message') and hasattr(context.message, "arguments") and "user_google_email" in context.message.arguments:
+                    # Replace user_google_email argument with authenticated email
+                    logger.debug("Replacing user_google_email argument with email from jwt token")
+                    context.message.arguments['user_google_email'] = authenticated_user_email
             
             logger.debug("Passing to next handler")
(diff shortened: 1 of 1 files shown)


shawnzhu:
@taylorwilsdon Thanks for merging #137. I'd like to hear your comments for JWT authenticated user.

taylorwilsdon:
Definitely a logical iteration, appreciate the PR!. Looking at the code, from what I can see this is firing after `_process_request_for_auth` takes place so we're already inspecting session, validating token etc - I think the refact we need overall is to remove user_google_email as a param from the tools entirely when in multi-user mode and basically pre-fill it based on the active bearer token being provided.

shawnzhu:
> I think the refact we need overall is to remove user_google_email as a param from the tools entirely when in multi-user mode and basically pre-fill it based on the active bearer token being provided.

This is what I did before but changing schema will be a break change. This PR is a trade-off to avoid it.

Let me see how much FastMCP supports dynamic tool schema.

shawnzhu:
@taylorwilsdon See #164 as PoC for removing `user_google_email` param when using multi-user mode.

If this approach works, I will close this one and it could merge the capability of `inject_user_email` decorator into `require_google_service` to avoid creating new decorator

taylorwilsdon:
excellent, I'll take a look asap - thanks!

taylorwilsdon:
@shawnzhu I might have an even more direct solve here - hang tight I will push a PR when I'm done with work for the day

taylorwilsdon:
Hey @shawnzhu - give the latest release a try and let me know if you're all set with that - it's rewriting the user google email on the fly in oauth2.1 mode now so all logs and such should match up

shawnzhu:
@taylorwilsdon I've tried the latest release with 1.3.4, it does replace the `user_google_email` by using the email from google's oauth token.

Two thoughts:

1. what about email from JWT token? I'd prefer the JWT token for authentication so that user don't need to access/refresh/store their google oauth token. Which is this PR is about.
2. the `user_google_gmail` argument from tool schema. it should be removed for mutli-user mode when using authentication over OAuth 2.1 or JWT token - I will rebase #164 to remove it by using `require_google_service` decorator

Issue #103, opened 2025-07-17
Enable credentials can be persisted in database


gyliu513:
## Feature Request

It would be helpful if google_workspace_mcp could support persisting Google API credentials (especially service account or user credentials) directly in a database, rather than requiring them to be stored only in file-based or in-memory formats.

## Motivation
In some deployment scenarios—such as multi-tenant or distributed systems—it’s necessary to:

Dynamically store and retrieve credentials tied to specific users or organizations.

Rotate and update credentials programmatically.

Avoid filesystem storage, especially in containerized or serverless environments.

FYI @taylorwilsdon

taylorwilsdon:
Hm, with service account credentials you can just pass as an environment variable but I'm curious to learn more about the use case! My biggest concern would be that I'm not aware of any other MCP that implements a database and that would potentially mean significantly more overhead, but I'm open minded! Are you thinking sqlite or like a full fat relational db like postgres?

gyliu513:
@taylorwilsdon sorry for late, any lightweighted db will works, like sqlite etc., customer can use the db as they want, if we use sqlalchemy, I think this will enable the mcp server can connect to most db that the customer is using.

shawnzhu:
@taylorwilsdon I'd like to have a centralized credential store (e.g., any RDBMS) if running multiple servers, so that it doesn't need to provide a multi-read-write storage for all pods in k8s.

Idea in my mind:

1. a `CredentialStore` as abstraction layer to support different implementation of credential store, however, the current directory based credential store will be the default one so no break changes to any users.
2. implement a `RDBMSCredentialStore` class so that it could configure it to swap the default store.

taylorwilsdon:
I think that's a relatively sensible pattern - doesn't even need to be that complex, could just be an environment variable config (ie a sqlalchemy setup where you can pass a string that's sqlite or postgres or whatever and it's indifferent so long as the orm can handle it) and basically if set, fall into this layer. 

PR welcome!

taylorwilsdon:
Hey there @gyliu513  - @shawnzhu just shipped #161 which should open the door here - love to see where we take this

taylorwilsdon:
Now that we don't even recommend storing credentials at all and instead using OAuth 2.1, I'm going to close this out - feel free to reopen if there's still interest!

PR #161, opened 2025-08-10
feat: abstract credential store

shawnzhu:
Supports #103 

This is just a refactoring to support adding different credential store in next step.

comments are appriciated!


taylorwilsdon:
Dang, that was quick! Thank you - looking forward to digging into this. Great addition.

taylorwilsdon:
Mind adding some quick docs @shawnzhu I'll get this in with the next release

shawnzhu:
add a section in `README.md`

taylorwilsdon:
Tested, seems happy! Let's get this shipped so we can build on it

PR #164, opened 2025-08-11
remove user_google_email parameter for multi-user mode

shawnzhu:
It introduce a new environment variable `MCP_REMOVE_USER_GOOGLE_EMAIL_ARG=1` to remove `user_google_email`
parameter from tool schema under multi-user mode.

the user_email will be pre-filled by `auth_info_middleware.py` from either google oauth token or JWT token.

see #160


taylorwilsdon:
Hey @shawnzhu - give the latest release a try and let me know if you're all set with that - it's rewriting the user google email on the fly in oauth2.1 mode now so all logs and such should match up

shawnzhu:
@taylorwilsdon it's nice to reuse `require_google_service` decorator under #166 . let me rebase this one to focus on removing `user_google_email` arguments for multi-user mode

taylorwilsdon:
> @taylorwilsdon it's nice to reuse `require_google_service` decorator under #166 . let me rebase this one to focus on removing `user_google_email` arguments for multi-user mode

My thought was that since it's now entirely ignored, I'm not sure that does us anything particularly useful but certainly open to dissent. I wonder if in part because I bake my emails into my system prompts in each client I abstract away a pain point that this would basically automate? Maybe all we need to do is just make it an optional param instead of a required one, but I'd love to see what you're cooking up.

shawnzhu:
If user's google email is baked in system prompt AND model picks it up as expected, it won't harm just more effort for a model to do.

When model could not pick it up as expected, it may hallucinate one like `user@example.com` from example within the tool schema, or asking user to provide one because it's required. It won't harm either just unnecessary.

I could add an environment variable `MCP_REMOVE_USER_GOOGLE_EMAIL_ARG` toggle in this PR to make it an option so it won't be a break change.

taylorwilsdon:
This is extremely clean, thank you! I will get it merged. Big picture I don't actually think we even need the env var flag, if you're in multi user mode the param is explicitly ignored so including it only creates confusion.

taylorwilsdon (approved):

PR #168, opened 2025-08-13
feat(gmail): add HTML body extraction with text fallback

Menelaus48:
Summary  
Add HTML body extraction to Gmail message retrieval functions so the server returns HTML content when a plain-text part is not available.

Problem  
Some incoming messages are HTML-only and the previous implementation only extracted `text/plain`, resulting in empty bodies for those messages.

Solution  
- Add a new helper `_extract_message_bodies()` that returns both text and HTML bodies by traversing multipart payloads.  
- Update retrieval functions to prefer `text/plain` and fall back to the HTML body when text is absent:  
  - `get_gmail_message_content()`  
  - `get_gmail_messages_content_batch()`  
- Preserve the existing `_extract_message_body()` wrapper for backward compatibility.  
- Add safe base64 decoding and error handling.  
- Add truncation for very large HTML bodies to avoid oversized outputs (single and batch limits).

Output format  
--- BODY ---  
[HTML Content Converted]  
<!DOCTYPE html>...

Testing  
- HTML-only messages: HTML content is now returned (instead of an empty body).  
- Multipart messages: `text/plain` is preferred when present; `text/html` is provided as fallback.  
- Text-only messages: behavior unchanged.  
- Verified end-to-end for both single-message and batch retrievals against representative HTML-only messages in a live environment.

Compatibility  
- No breaking API changes.  
- No new runtime dependencies.  
- No version bump included in this PR; maintainers can bump the package version per release process.

Notes for maintainers  
- The change is intentionally scoped to `gmail/gmail_tools.py` only.  
- If you prefer a different truncation policy or a `body_format` parameter (e.g., `best|text|html|both|raw`), I can supply a small follow-up PR to add that as an optional enhancement.

I implemented an HTML fallback for Gmail message retrieval and tested both single and batch retrievals in a live environment. Both now return HTML where appropriate and maintain the original plain-text-first behavior. Happy to respond to review comments or adjust truncation policy if you prefer.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -33,6 +33,7 @@ GMAIL_REQUEST_DELAY = 0.1
 def _extract_message_body(payload):
     """
     Helper function to extract plain text body from a Gmail message payload.
+    (Maintained for backward compatibility)
 
     Args:
         payload (dict): The message payload from Gmail API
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
Looks like a sensible, clean change - thanks @Menelaus48! One quick nit pick about the truncation but otherwise we should be able to get this merged in the next release.

taylorwilsdon:
Pulled in main and small refac, will ship this today after I wrap work for the day!

PR #181, opened 2025-08-20
Add markdown export tool for google docs

diunko:
Hey! wdyt about having markdown export in gdocs? 
Found it very useful for me, so thought I'd share it.
Uses Drive API's export_media with markdown mime type - different from get_doc_content which needs Docs API for parsing, so kept them separate. Also handles base64 images by replacing them with placeholders so they don't bloat the context.
Tested with a couple docs so far - basic formatting works fine but haven't tried anything super complex.

### DIFF
--- a/core/tool_tiers.yaml
+++ b/core/tool_tiers.yaml
@@ -43,6 +43,7 @@ docs:
     - get_doc_content
     - create_doc
     - modify_doc_text
+    - get_doc_markdown
   extended:
     - search_docs
     - find_and_replace_doc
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
I prefer markdown to almost anything else, great idea. Will get it reviewed today, thanks @diunko

taylorwilsdon:
Can you refactor that giant image swapping helper function out of docs tools (which is already way too big haha) into the helper file in the same directory? Will get it in! Thanks

kapilthakare-cyberpunk:

Issue #180, opened 2025-08-20
non-string list_tasks parameters fail with *Input validation error* in Claude desktop


ForrestTrepte:
**Describe the bug**
This was working for me previously. I'm not sure what caused it to start happening. I'm now having this problem with versions of the MCP server that previously worked and as well as with the latest version from main. Perhaps something changed in Claude itself?! Or more likely there's something I'm not understanding.

**Startup Logs**
```
⚙️ Active Configuration:
   - GOOGLE_OAUTH_CLIENT_ID: 841300875238-lcjd04ehdtukplaubu0blooml7h47t23.apps.googleusercontent.com
   - GOOGLE_OAUTH_CLIENT_SECRET: GOCS...oMfk
   - USER_GOOGLE_EMAIL: Not Set
   - MCP_SINGLE_USER_MODE: false
   - MCP_ENABLE_OAUTH21: true
   - OAUTHLIB_INSECURE_TRANSPORT: false
   - GOOGLE_CLIENT_SECRET_PATH: Not Set

INFO:auth.scopes:Enabled tools set for scope management: ['gmail', 'drive', 'calendar', 'docs', 'sheets', 'chat', 'forms', 'slides', 'tasks', 'search']
🛠️  Loading 10 tool modules:
   📧 Gmail - Google Gmail API integration
   📁 Drive - Google Drive API integration
   📅 Calendar - Google Calendar API integration
   📄 Docs - Google Docs API integration
   📊 Sheets - Google Sheets API integration
   💬 Chat - Google Chat API integration
   📝 Forms - Google Forms API integration
   🖼️ Slides - Google Slides API integration
   ✓ Tasks - Google Tasks API integration
   🔍 Search - Google Search API integration

📊 Configuration Summary:
   🔧 Services Loaded: 10/10
   📝 Log Level: 20

🔍 Checking credentials directory permissions...
INFO:core.utils:Credentials directory permissions check passed: C:\Users\forre\.google_workspace_mcp\credentials
✅ Credentials directory permissions verified

INFO:core.server:🔌 Transport: streamable-http
--- Logging error ---
Traceback (most recent call last):
  File "C:\Python311\Lib\logging\__init__.py", line 1113, in emit
    stream.write(msg + self.terminator)
  File "C:\Python311\Lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
[...]

PR #184, opened 2025-08-20
enh: Safe logging & Log Formatting

taylorwilsdon:
Closes https://github.com/taylorwilsdon/google_workspace_mcp/issues/180

### DIFF
--- a/core/log_formatter.py
+++ b/core/log_formatter.py
@@ -1,0 +1,142 @@
+"""
+Enhanced Log Formatter for Google Workspace MCP
+
+Provides visually appealing log formatting with emojis and consistent styling
+to match the safe_print output format.
+"""
+import logging
+import re
+
+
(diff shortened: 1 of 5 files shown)


ForrestTrepte (commented):
My three cents:
1. I'd recommend separating safe logging from the parameter type changes in different PRs. So, later, if it's an Anthropic bug that's fixed, the type changes can be cleanly reverted by backing out the parameter type chage PR.
2. Personally, I'd hold off on implementing the type change workaround. Instead, I'd first focus on understanding the Anthropic issue better.
3. This just fixes one parameter. What about all the other non-string parameters in the codebase?

taylorwilsdon:
Tasks is actually the only tool in the whole repo that implements an int type so this is the sole change to make, I'm not having issues with bool types. Happy to hold off though, its a very minor issue (you can still use get list, just not with max results overridden)

PR #187, opened 2025-08-22
Refactor base URL handling and add Docker Compose for local development

abelsr:
This pull request primarily updates how the application's base URL and port are handled throughout the authentication and configuration code, and adds a Docker Compose configuration for local development. The most significant changes involve removing the port from the `base_url` variable, which simplifies URL management and ensures consistency. Additionally, a new `docker-compose.yml` file is introduced to streamline running the service locally.

Configuration and URL handling improvements:

* Removed the port from the `base_url` in `auth/oauth_config.py` and updated its usage in `auth/google_remote_auth_provider.py` to ensure URLs are consistently constructed without appending the port. [[1]](diffhunk://#diff-13f6b130be79ba2fa7350c792f996a8135e9c5538d37dc3e92e02b9119f82902L28-R28) [[2]](diffhunk://#diff-66082450489b1dd7fb3a4cc897721c79ec2ce9248175e0078cfa35770bf12fb8L89-R93)
* Updated the protected resource URL in `auth/oauth_common_handlers.py` to include a trailing slash, aligning with standard URL conventions.

Development environment setup:

* Added a new `docker-compose.yml` file to define a `gws_mcp` service for easier local development and deployment, including port mapping, environment variables, and volume mounting for secrets.

### DIFF
--- a/auth/google_remote_auth_provider.py
+++ b/auth/google_remote_auth_provider.py
@@ -60,11 +60,14 @@ class GoogleRemoteAuthProvider(RemoteAuthProvider):
         if not REMOTEAUTHPROVIDER_AVAILABLE:
             raise ImportError("FastMCP v2.11.1+ required for RemoteAuthProvider")
 
-        # Get configuration from environment
-        self.client_id = os.getenv("GOOGLE_OAUTH_CLIENT_ID")
-        self.client_secret = os.getenv("GOOGLE_OAUTH_CLIENT_SECRET")
-        self.base_url = os.getenv("WORKSPACE_MCP_BASE_URI", "http://localhost")
-        self.port = int(os.getenv("PORT", os.getenv("WORKSPACE_MCP_PORT", 8000)))
+        # Get configuration from OAuth config
+        from auth.oauth_config import get_oauth_config
+        config = get_oauth_config()
+        
+        self.client_id = config.client_id
+        self.client_secret = config.client_secret
+        self.base_url = config.get_oauth_base_url()
+        self.port = config.port
 
         if not self.client_id:
             logger.error(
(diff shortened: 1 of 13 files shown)


auth/google_remote_auth_provider.py
```diff
@@ -86,9 +86,10 @@ def __init__(self):
         # The /mcp/ resource URL is handled in the protected resource metadata endpoint
         super().__init__(
             token_verifier=token_verifier,
-            authorization_servers=[AnyHttpUrl(f"{self.base_url}:{self.port}")],
-            resource_server_url=f"{self.base_url}:{self.port}",
+            authorization_servers=[AnyHttpUrl(f"{self.base_url}")],
```
taylorwilsdon: This might solve your issue specifically but it will break it for everyone else - I think what we'll actually need here is a check that if the `GOOGLE_OAUTH_REDIRECT_URI` env var is set, then drop that in, otherwise use base url + port.

auth/oauth_config.py
```diff
@@ -25,7 +25,7 @@ def __init__(self):
         # Base server configuration
         self.base_uri = os.getenv("WORKSPACE_MCP_BASE_URI", "http://localhost")
         self.port = int(os.getenv("PORT", os.getenv("WORKSPACE_MCP_PORT", "8000")))
-        self.base_url = f"{self.base_uri}:{self.port}"
+        self.base_url = f"{self.base_uri}"
```
taylorwilsdon: Same comment as earlier

taylorwilsdon (changes requested):
Appreciate the PR! We need to change a few things but I will take a run now and we'll get this merged asap.

PR #195, opened 2025-08-29
explain that when show_completed is set, show_hidden should also be set

### LINKED ISSUE #139: feature request: help LLMs to find completed tasks by including hidden tasks
In the Google Tasks GUI, it when tasks are completed they become hidden. When using list_tasks, show_hidden defaults to false. An LLM may be unaware that completed tasks are hidden, causing it to fail to retrieve completed tasks.

From the [Google API docs](https://developers.google.com/workspace/tasks/reference/rest/v1/tasks#Task):
| Note that showHidden must also be True to show tasks completed in first party clients, such as the web UI and Google's mobile apps.

To reproduce:
* Create a Google Tasks list at tasks.google.com
* Check some of them off (they will move into the Completed dropdown at the bottom)
* Ask Claude, using the MCP server, to show all tasks including completed tasks from that list
* Observe that none of the completed tasks are retrieved. Note that show_completed is true, but show_hidden is not set.

I suggest either of these solutions:
1. Change the default value of show_hidden to true and the default value of show_completed to false. Although that's different than the default values in the Google API, it *seems* like this would produce the expected results. The default behavior would be that completed tasks are not shown (matching the current default behavior of the tool). But with this change, when the show_completed parameter is changed to 'true', then completed tasks would be included as one would expect.
2. Give guidance to the llm in the instructions for the show_completed parameter explaining that completed tasks become hidden, so recommend that it probably wants to also set show_hidden to true in order to retrieve completed tasks.

It seems like this is a confusing aspect of the Google API design. The [task documentation](https://developers.google.com/workspace/tasks/reference/rest/v1/tasks#Task) says:
| Flag indicating whether the task is hidden. This is the case if the task had been marked completed when the task list was last cleared. The default is False. This field is read-only.
It's unclear to me exactly what the purpose is of hidden tasks. I'm not sure if tasks are hidden if and only if they are completed. Or perhaps there are other ways tasks can become hidden via the API without being completed.

ForrestTrepte:
fixes #139

add comment from https://developers.google.com/workspace/tasks/reference/rest/v1/tasks/list about the parameter show_completed

This doc comment is provided to the LLM in tool information, helping it know how to show completed tasks.

### DIFF
--- a/gtasks/tasks_tools.py
+++ b/gtasks/tasks_tools.py
@@ -289,7 +289,7 @@ async def list_tasks(
         task_list_id (str): The ID of the task list to retrieve tasks from.
         max_results (Optional[int]): Maximum number of tasks to return. (default: 20, max: 10000).
         page_token (Optional[str]): Token for pagination.
-        show_completed (bool): Whether to include completed tasks (default: True).
+        show_completed (bool): Whether to include completed tasks (default: True). Note that show_hidden must also be true to show tasks completed in first party clients, such as the web UI and Google's mobile apps.
         show_deleted (bool): Whether to include deleted tasks (default: False).
         show_hidden (bool): Whether to include hidden tasks (default: False).
         show_assigned (bool): Whether to include assigned tasks (default: False).


taylorwilsdon:
Sorry for the delay, been out and just getting caught back up. Appreciate this!

PR #206, opened 2025-09-08
chore: update readme with export_doc_to_pdf tool

emanuilo:
@taylorwilsdon I updated the readme with PDF export tool as we agreed.


taylorwilsdon:
Thanks much!

PR #211, opened 2025-09-17
feat(gmail): Add vacation responder tool

klemens-wisser-pcg:
This commit adds the ability to manage Gmail's vacation responder settings.

Two new tools have been added:
- `get_vacation_settings`: Retrieves the current vacation responder settings.
- `set_vacation_settings`: Updates the vacation responder settings.

A new scope, `GMAIL_SETTINGS_BASIC_SCOPE`, has been added to `auth/scopes.py` to support this functionality.


taylorwilsdon:
I'm wondering if a more generic "manage settings" tool that can do all of the above depending on what gets passed to it is a more effective way to accomplish this. We're already at the point of saturating almost all mainstream clients pushing 80-90 tools so ideally we consolidate anything that we can into a single registered tool that accepts arguments rather than an individual tool per task. I think this is valuable and would like to get added!

kapilthakare-cyberpunk:

PR #219, opened 2025-09-28
enh: Legacy OAuth session binding & compatibility with modern methods

taylorwilsdon:
This change is relatively simple but keeps our codebase more secure and consistent. 

  - Bind OAuth v2.1 state tokens to initiating FastMCP sessions, persist them with expiry, and validate/consume on callback to block replay or cross-session misuse.
  - Capture the active FastMCP session ID during login initiation and propagate it through the OAuth callback flow, trimming verbose state logging.
  - In stdio mode, auto-select the sole authenticated user when only one OAuth session exists to streamline headless usage.
  - Better timezone aware handling for token expiry.

### DIFF
--- a/auth/oauth_config.py
+++ b/auth/oauth_config.py
@@ -129,6 +129,8 @@ class OAuthConfig:
         """
         if self.external_url:
             return self.external_url
+        if self.stateless_mode:
+            return "https://accounts.google.com"
         return self.base_url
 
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
This will also fix compatibility with the new oauth 2.1 mode in open-webui shoutout @tjbck for making that happen

PR #234, opened 2025-10-13
chore: Gmail tool enable html message for `draft_gmail_message`

luutuankiet:
Picking up from #224 now that we have html support in `send_gmail_message`.
This PR ports over the argument and docs to `draft_gmail_message` so that agents can send HTML formatted drafts. 

Should be a trivial change as both points to the same internal  func `_prepare_gmail_message` , but do let me know if you need additional change!


taylorwilsdon:
Awesome, looks good - appreciate it!

Issue #236, opened 2025-10-15
Add a directory path option when creating a doc.


kevintech1020:
For now, I can create a doc with a title, but I can't specify the directory. I want the folder to be created if it doesn't exist. File directory check and create feature before creating a doc is helpful.

taylorwilsdon:
Ah that's a good idea, I believe you can specify the parent ID in files.create() calls so it should be trivial. PR welcome or I'll handle it in the next version!

taylorwilsdon:
merged

PR #241, opened 2025-10-18
Feat/docs folder support

yakiv07:
Adds the ability to create Google Docs in specific folders (My Drive or Shared Drives) with automatic folder creation if needed.

### **Key Features**
- ✅ Create docs in folders by ID or name
- ✅ Auto-create folders if they don't exist  
- ✅ Support for Shared Drives (Team Drives)
- ✅ Direct folder placement using Drive API
- ✅ New folder management tools

### **Changes**
- Enhanced `create_doc` function with `folder_id` and `folder_name` parameters
- Added `find_or_create_folder` helper in `drive_helpers.py`
- Added `create_drive_folder` and `search_drive_folders` tools
- Proper Shared Drive support with `supportsAllDrives=True`

### DIFF
--- a/auth/oauth21_session_store.py
+++ b/auth/oauth21_session_store.py
@@ -18,6 +18,38 @@ from google.oauth2.credentials import Credentials
 
 logger = logging.getLogger(__name__)
 
+
+def _normalize_expiry_to_naive_utc(expiry: Optional[Any]) -> Optional[datetime]:
+    """
+    Convert expiry values to timezone-naive UTC datetimes for google-auth compatibility.
+
+    Naive datetime inputs are assumed to already represent UTC and are returned unchanged so that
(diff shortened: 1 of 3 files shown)


taylorwilsdon:
I like this, appreciate the PR! Looks like it would close https://github.com/taylorwilsdon/google_workspace_mcp/issues/236 as well. Little less sure about the value of search within folder, since you should already be able to do that with the base search drive file. We're at the point of saturating small model context windows based on the sheer number of tool descriptions if users don't exclude certain tools or leverage the tool tiers, so I'm trying to balance nice to have functionality against keeping bloat down :D 

Let me know if you're game to drop the search part @yakiv07 happy to get this merged

kapilthakare-cyberpunk:

Issue #243, opened 2025-10-21
How to reference to platform user id in a multiuser deployment


oscmejia:
What is the best way to get a refference to the connected google accounts from the platform that imementes the MCP server, in the context of a multi-user deployment.
When a user connected an account, an the JSON with credentails is created, the platform has no way to detect which account what linked to which user. Is there any functionality for this in place?

taylorwilsdon:
Hi @oscmejia - there is a whole mode specific to this use case. You'll want to set `MCP_ENABLE_OAUTH21=true` and `WORKSPACE_MCP_STATELESS_MODE=true` if you don't want to cache the credential json on the server. If you run in stateful mode (with the json credentials) then it establishes an immutable session binding between the client and the particular credential object, and if that session ends, the user will be prompted to reauthenticate during their next use (at which point their new session will be bound to the credential file)

Check out the multi-user section in the readme doc it should cover just what you need, that's how I personally have it deployed! https://github.com/taylorwilsdon/google_workspace_mcp?tab=readme-ov-file#oauth-21-support-multi-user-bearer-token-authentication

oscmejia:
Hi @taylorwilsdon 
I'm running the MCP with MCP_ENABLE_OAUTH21=true and WORKSPACE_MCP_STATELESS_MODE=true, as instructed, for a mult user use case.

I impemented Google Auth, on my side, and allows users to grant access to scopes per service, not all at once.
I noticed that you impemented Per-tool scope validation (service_decorator.py:247), but before that whgen validating token, you check that all scoped are grantted and if not, rejects with 401.

It will be straighforward to bypass that check with `is_stateless_mode()`, and copes will be validated by the decorator on a per servivice base.

If you agree with this and think is useful, I can submit a PR. Let me know.

oscmejia:
PR: #246

PR #246, opened 2025-10-22
Feat/stateless base scope validation

oscmejia:
feat: Add base scope validation for OAuth 2.1 stateless mode 
- Modify GoogleProvider initialization to use BASE_SCOPES in stateless mode
- Enable dynamic tool filtering based on user's granted scopes
- Per-tool validation handles service-specific scope checks  
- Maintains security while supporting partial scope authorization

This allows users with different scope configurations to use only the Google Workspace tools they have authorized, with clear error messages for unauthorized tools.


thecopy:
I believe this would also solve our issue, we want to manage the authorization outside the MCP server, and simply provide a valid Google token to the MCP server with specific scopes. So want to MCP server capable of everything (all tools enabled) but on a per-token basis this might change.

oscmejia:
for context: https://github.com/taylorwilsdon/google_workspace_mcp/issues/243

PR #249, opened 2025-10-25
Fix: Gmail send_message now respects user_google_email for From header

loserbcc:
## Problem
Emails sent via send_gmail_message were always showing as from a default user account regardless of which account was specified in the user_google_email parameter.

## Root Cause
The _prepare_gmail_message() function was not setting the From header in the MIME message. Gmail API defaults to the primary send-as address when no From header is provided.

## Solution
- Added from_email parameter to _prepare_gmail_message()
- Set From header explicitly: message["From"] = from_email
- Updated send_gmail_message() to pass user_google_email as from_email
- Updated draft_gmail_message() to pass user_google_email as from_email
- Capitalized MIME headers (Subject, From, To, Cc, Bcc) per conventions

## Testing
Verified emails now correctly show sender address for multiple accounts. Each account now properly displays its own address in the From field.

Fixes multi-account email sending functionality.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -149,6 +149,7 @@ def _prepare_gmail_message(
     in_reply_to: Optional[str] = None,
     references: Optional[str] = None,
     body_format: Literal["plain", "html"] = "plain",
+    from_email: Optional[str] = None,
 ) -> tuple[str, Optional[str]]:
     """
     Prepare a Gmail message with threading support.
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
++ I like this idea, I can imagine a lot of scenarios where you might want to send as something different than the underlying address that you log in as. I personally have 3 different send-as addresses validated on mine so I totally see the use case. Will get this shipped today!

loserbcc:
> ++ I like this idea, I can imagine a lot of scenarios 
thaanks dude -- I love your mcp! (i hoped I didn't make ya angry) I'm kinda new at sharing.

PR #251, opened 2025-10-27
Adds 'minimal' and 'metadata' format options for get_gmail_thread_content and get_gmail_threads_content_batch

mutianzhai:
Adds 'minimal' and 'metadata' format options for get_gmail_thread_content and get_gmail_threads_content_batch.

"minimal" - Returns only the `id` and `threadId` of each message.
"metadata" - Returns message metadata such as headers, labels, and snippet, but not the full body.
"full" (The current option) - Returns the full email message data, including headers and body.

### DIFF
--- a/README.md
+++ b/README.md
@@ -724,6 +724,17 @@ cp .env.oauth21 .env
 | `batch_modify_gmail_message_labels` | Complete | Batch modify labels |
 | `start_google_auth` | Complete | Initialize authentication |
 
+**Gmail Message Format Options:**
+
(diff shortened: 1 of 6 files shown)


taylorwilsdon:
It looks like it's a much larger change than just the format options, adding a few tools? Does the message label tool not work for threads? Afaik the parent message for a given thread is what ultimately dictates the label for the whole thing.

mutianzhai:
Sorry I should not have made the pull from main. Regarding the labels, if you search for gmail messages without a label, you'll get messages from the thread that don't have the label (I think labels are localized to the message). So if you want to exclude all messages on a thread that has a label, I think you have to label all the messages in the thread. Just an idea if you want to use.

PR #253, opened 2025-10-27
Fix multi-user credential switching in same session

loserbcc:
## Summary

This PR fixes a critical bug where the MCP server was unable to switch between multiple Google account credentials within the same session. The server would bind to the first authenticated user and ignore the `user_google_email` parameter on subsequent requests.

## Problem

The `get_credentials()` function in `auth/google_auth.py` has two credential lookup paths:
1. **OAuth21SessionStore.get_credentials_by_mcp_session()** (lines 542-590)
2. **load_credentials_from_session()** (lines 623+)

Both paths were returning cached credentials based solely on `session_id`, without validating that the cached credentials matched the requested `user_google_email` parameter. This caused all tool calls within a session to use the first authenticated user's credentials, even when a different `user_google_email` was specified.

## Solution

Added user validation to **both** credential lookup paths:

### Changes to `auth/google_auth.py`

**Path 1: OAuth21SessionStore lookup (lines 546-590)**
- After retrieving credentials by MCP session, validate the cached user matches the requested user
- If mismatch detected, set `credentials = None` to force reload from credential store
- Log warning when bypassing cache due to user mismatch

**Path 2: Session cache lookup (lines 623-648)**  
- After loading credentials from session, validate the cached user matches the requested user
- If mismatch detected, set `credentials = None` to force reload from credential store
- Log warning when bypassing cache due to user mismatch

Both validations use `OAuth21SessionStore.get_user_by_mcp_session()` to determine the cached user email and compare it with the requested `user_google_email` parameter.

## Testing

Verified the fix works correctly:

✅ **Create operations** - Can create tasks on different accounts:
- Created task on user1@example.com
- Created task on user2@example.com
- Each task has unique ID confirming separate accounts

✅ **List operations** - Returns correct data for each account:
- user1@example.com lists only its own tasks
- user2@example.com lists only its own tasks  

✅ **Account switching** - Can alternate between accounts seamlessly without re-authentication

## Impact

This fix enables:
- Multi-user support within the same Claude Code session
- Correct credential selection based on `user_google_email` parameter
- No breaking changes to existing functionality
- Maintains existing security model (session-to-user binding)

## Files Changed

- `auth/google_auth.py` - Modified `get_credentials()` function to validate cached credentials match requested user

## Commits

- `1975ffd` - Initial fix for session cache credential validation
- `6b006c9` - Complete fix for OAuth21SessionStore credential validation

---

### DIFF
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -546,38 +546,48 @@ def get_credentials(
             # Try to get credentials by MCP session
             credentials = store.get_credentials_by_mcp_session(session_id)
             if credentials:
-                logger.info(f"[get_credentials] Found OAuth 2.1 credentials for MCP session {session_id}")
+                # Validate cached credentials match requested user to support multi-user scenarios
+                cached_user = store.get_user_by_mcp_session(session_id)
 
-                # Check scopes
-                if not all(scope in credentials.scopes for scope in required_scopes):
+                if cached_user and user_google_email and cached_user != user_google_email:
                     logger.warning(
-                        f"[get_credentials] OAuth 2.1 credentials lack required scopes. Need: {required_scopes}, Have: {credentials.scopes}"
+                        f"[get_credentials] OAuth21SessionStore has credentials for {cached_user} "
+                        f"but {user_google_email} was requested. Bypassing OAuth 2.1 store to load correct credentials."
                     )
-                    return None
+                    credentials = None  # Force reload from credential store
+                else:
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
OAuth2.1 mode is meant to behave this way specifically, where the client handles the credential flow and manages storage / caching / refresh (and potentially session switching) - binding a single session to a valid authorization token so that you can operate in a multi-user environment without risking inadvertent access. If you want to use multiple accounts in a trusted environment, you just don't set that env var and you can have multiple accounts authenticated and switch between them as desired.

PR #265, opened 2025-11-04
Feat/stateless base scope validation

oscmejia:
feat: Add base scope validation for OAuth 2.1 stateless mode

Modify GoogleProvider initialization to use BASE_SCOPES in stateless mode
Enable dynamic tool filtering based on user's granted scopes
Per-tool validation handles service-specific scope checks
Maintains security while supporting partial scope authorization
This allows users with different scope configurations to use only the Google Workspace tools they have authorized, with clear error messages for unauthorized tools.


taylorwilsdon:
The change conflicts with the existing scope management system in auth/scopes.py:129-154. You have get_current_scopes() which already handles dynamic scope generation based on enabled tools, but you're bypassing this entirely. No implementation shown for handling scope validation failures at the tool level. The comment mentions "clear error messages for unauthorized tools" but that's not here. Open to the idea but this needs work before it could be reviewed, thanks!

PR #270, opened 2025-11-09
feat: Add support for file:// URLs in create_drive_file

yairwein:
Added ability to upload local files to Google Drive using file:// URL scheme in the create_drive_file tool. This enables uploading files from /tmp and other local paths without requiring HTTP/HTTPS hosting.

Changes:
- Added urllib.parse.urlparse and pathlib.Path imports
- Detect file:// URLs and read directly from local filesystem
- Validate file exists before attempting upload
- Updated docstring to reflect support for file://, http://, and https://
- Added error handling for unsupported URL schemes

### DIFF
--- a/README.md
+++ b/README.md
@@ -698,8 +698,9 @@ cp .env.oauth21 .env
 |------|------|-------------|
 | `search_drive_files` | **Core** | Search files with query syntax |
 | `get_drive_file_content` | **Core** | Read file content (Office formats) |
-| `list_drive_items` | Extended | List folder contents |
 | `create_drive_file` | **Core** | Create files or fetch from URLs |
+| `list_drive_items` | Extended | List folder contents |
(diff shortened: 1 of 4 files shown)


taylorwilsdon:
Clever, makes sense to me! Looks like it's failing `uv run ruff check` if you don't mind applying

taylorwilsdon:
I do have one slight concern here @yairwein that this will not work with streamable http endpoints hosted outside the client system, and won't ever work with docker which is an increasingly popular way to deploy. Might need to add some kind of switch/disclaimer?

taylorwilsdon:
Tested, works great with the updates. Thanks!

yairwein:
Thanks for adding @taylorwilsdon - I am actually running this in streamable with a mounted folder to share files and it works great. 
It is a good way to upload files created by agents to drive

PR #280, opened 2025-11-26
fix: gdoc tools fail with Input validation error: 'drive_service' is a required property

owengo:
Fixes error: "Input validation error: 'drive_service' is a required property"  when using the get_doc_content tool.

The patch is generated by codex:

get_doc_content is decorated with @require_multiple_services (gdocs/docs_tools.py (line 90)), but the decorator in auth/service_decorator.py:638+ never strips the injected service params from the wrapper signature. FastMCP builds the tool schema from that signature, so drive_service/docs_service show up as required inputs. When the client calls the tool with only document_id, validation fails with “'drive_service' is a required property”.

Fix: in require_multiple_services, build wrapper_sig without the injected service param names (and remove user_google_email only when OAuth 2.1 is enabled), and use the filtered param list when locating user_google_email/calling _override_oauth21_user_email. After that, the tool schema will only require the real inputs (e.g., document_id, user_google_email in OAuth 2.0).

### DIFF
--- a/auth/service_decorator.py
+++ b/auth/service_decorator.py
@@ -657,17 +657,17 @@ def require_multiple_services(service_configs: List[Dict[str, Any]]):
 
     def decorator(func: Callable) -> Callable:
         original_sig = inspect.signature(func)
+        params = list(original_sig.parameters.values())
 
-        # In OAuth 2.1 mode, remove user_google_email from the signature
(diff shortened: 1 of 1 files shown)


owengo:
Ok didn't see there was already a PR for this issue

PR #293, opened 2025-12-09
Add get_drive_file_download_url tool for downloading Drive files via HTTP

joshed-io:
Implements a new tool that follows the same pattern as Gmail attachment downloads:

## Features
- Downloads files from Google Drive (handles shortcuts automatically)
- For Google native files (Docs/Sheets/Slides), exports to useful formats:
  * Docs → PDF (default) or DOCX
  * Sheets → XLSX (default) or CSV  
  * Slides → PDF (default) or PPTX
- Saves files to AttachmentStorage (reuses existing infrastructure)
- Returns a download URL that works with the existing `/attachments/{file_id}` route
- Files expire after 1 hour (same as Gmail attachments)
- Handles stateless mode gracefully

## Use Case
This enables agents to download binary files from Drive, which was previously not possible with `get_drive_file_content` (text-only). The agent receives a URL it can fetch to download the file, following the same pattern as Gmail attachment downloads.

## Example Usage
```
get_drive_file_download_url(
    user_google_email: "user@example.com",
    file_id: "1a2b3c4d5e6f7g8h9i0j",
    export_format: "pdf"  # optional for Google native files
)
```

Returns a URL like `http://localhost:8000/attachments/{uuid}` that the agent can fetch to download the file.

### DIFF
--- a/gdrive/drive_tools.py
+++ b/gdrive/drive_tools.py
@@ -14,6 +14,7 @@ from pathlib import Path
 from googleapiclient.http import MediaIoBaseDownload, MediaIoBaseUpload
 import io
 import httpx
+import base64
 
 from auth.service_decorator import require_google_service
 from auth.oauth_config import is_stateless_mode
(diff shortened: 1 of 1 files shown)


joshed-io:
@taylorwilsdon Here's a PR that extends the pattern for gmail attachments to Google Drive attachments. Thankfully, we don't have the `attachment` file issue here, at least in my testing. Hitting the download URL directly in the browser gives me the correct file name and extension.

PR #298, opened 2025-12-12
fix: tab title

owengo:
The code to get the tabs titles in a gdoc was not working, all the tabs showed as "Untitled Tab"
This fix gives the correct value for the tabs name and adds also the tabId, which can be useful to generate links  for example, and for later enhancements such as allowing to edit a specific tab.


taylorwilsdon:
You're on a roll, appreciate it!

PR #281, opened 2025-11-27
Allow formatting text and background color in google docs

owengo:
This PR, 100% generated with codex enhances the gdoc tool with the ability to set text color and the background color

PR #300, opened 2025-12-13
enh: Text & Background Color Formatting 

taylorwilsdon:
Closes out https://github.com/taylorwilsdon/google_workspace_mcp/pull/281


taylorwilsdon:
thanks @owengo looks good

PR #301, opened 2025-12-13
feat: ignore toolCallId to make this mcp compatible with n8n

jonzarecki:


taylorwilsdon:
Saw this one before it closed, I'm not a heavy n8n user but would be interested in getting this in if it is necessary to play nice!

PR #306, opened 2025-12-14
fix(uv): Install missing uvicorn dependency

whyredfire:
- Fixes `[ERROR] Unexpected error running server: 'websockets-sansio'` [1]
- Run `uv lock` to freeze dependency versions
- Use multi-stage docker build (512M -> 340M) [2]

[1] https://github.com/jlowin/fastmcp/issues/2299
[2] https://github.com/astral-sh/uv-docker-example/blob/fa761744819c05f4ce207bde1f0a396f6be3915f/multistage.Dockerfile


taylorwilsdon:
Not sure what the idea is here but this is a totally different approach and uvicorn is already a dependency we pull through fastmcp

whyredfire:
> Not sure what the idea is here but this is a totally different approach and uvicorn is already a dependency we pull through fastmcp

Oh right, I did not see that the first reference actually belongs to the FastMCP project itself. Please let me know if you're fine with the Dockerfile change, I'll either rebase or open a different PR for it.

PR #308, opened 2025-12-15
fix: oauth 2.1 broken

owengo:
Integration with Librechat's OAUTH 2.1  does not work anymore after the recent fastmcp upgrades

Here is the analysis and the fix done  by codex:
The patch seems quite convoluted but it has the merit to solve the broken, it was no more possible to use oauth with the MCP server without it

Issue:

The regression happens after the fastmcp upgrade (it’s unpinned: pyproject.toml (line 14), and now resolves to fastmcp 2.14.0 / mcp 1.24.0: uv.lock (line 415), uv.lock (line 817)). With the newer MCP SDK, the built-in /token handler is stricter about how token requests are encoded / where client_id is provided, so LibreChat’s token_exchange_method: default_post flow reaches the Google callback but then fails on POST /token (401), which leaves subsequent POST /mcp calls as 401 invalid_token.

Fix:

Added a compatibility auth provider that only swaps the /token route to accept both JSON and x-www-form-urlencoded token requests, and to backfill client_id/client_secret from Authorization: Basic … when missing, while still enforcing client validation, redirect_uri matching, and PKCE (auth/compat_google_provider.py (line 47), auth/compat_google_provider.py (line 100), auth/compat_google_provider.py (line 239), auth/compat_google_provider.py (line 352)).
Wired this provider in standard OAuth 2.1 mode (core/server.py (line 131)) so external-provider mode is unchanged.
Hardened the service decorators to derive the authenticated email from the validated access-token claims (and to treat “validated access token present” as OAuth 2.1) instead of relying on middleware state that isn’t consistently populated on newer FastMCP (auth/service_decorator.py (line 78), auth/service_decorator.py (line 106), auth/service_decorator.py (line 329)).

### DIFF
--- a/auth/service_decorator.py
+++ b/auth/service_decorator.py
@@ -57,11 +57,17 @@ def _get_auth_context(
     Returns:
         Tuple of (authenticated_user, auth_method, mcp_session_id)
     """
+    ctx = None
+    authenticated_user: Optional[str] = None
+    auth_method: Optional[str] = None
+    mcp_session_id: Optional[str] = None
+
     try:
         ctx = get_context()
-        if not ctx:
-            return None, None, None
+    except Exception as e:
+        logger.debug(f"[{tool_name}] Could not get FastMCP context: {e}")
 
+    if ctx:
         authenticated_user = ctx.get_state("authenticated_user_email")
         auth_method = ctx.get_state("authenticated_via")
         mcp_session_id = ctx.session_id if hasattr(ctx, "session_id") else None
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
The last time I crossed path with Librechat in the wild, the OAuth support was not compliant with the spec and very incomplete. I am unconvinced that FastMCP actually introduces a regression, and I suspect that librechat just does not implement the oauth spec correctly as I have observed the same in the past. FastMCP v2.14 just tightens conformance to the OAuth 2.0 spec (specifically [rfc 6749 §4.1.3](https://datatracker.ietf.org/doc/html/rfc6749)), which requires the token request to be formatted as application/x-www-form-urlencoded and I'm a little concerned adding a huge additional library as as workaround for non-compliant code isn't the right approach.

With that said, there is a ton of opportunity to clean up and improve the litany of auth code in this repo because most of it predates any native oauth support from clients or the MCP spec itself and I welcome the opportunity to improve things! Let me spin up libre and see what's what.  This project is just one of countless that leverages fastmcp (by far the most popular python mcp sdk) oauth capability so I have to think librechat will get their shit together rather than every one of those MCPs adding workarounds.

taylorwilsdon:
@owengo I opened https://github.com/danny-avila/LibreChat/issues/11055 with a proposed solution (that is a 3 line change instead of a 500 loc one here lol)

owengo:
> The last time I crossed path with Librechat in the wild, the OAuth support was not compliant with the spec and very incomplete. I am unconvinced that FastMCP actually introduces a regression, and I suspect that librechat just does not implement the oauth spec correctly as I have observed the same in the past. FastMCP v2.14 just tightens conformance to the OAuth 2.0 spec (specifically [rfc 6749 §4.1.3](https://datatracker.ietf.org/doc/html/rfc6749)), which requires the token request to be formatted as application/x-www-form-urlencoded and I'm a little concerned adding a huge additional library as as workaround for non-compliant code isn't the right approach.
> 
> With that said, there is a ton of opportunity to clean up and improve the litany of auth code in this repo because most of it predates any native oauth support from clients or the MCP spec itself and I welcome the opportunity to improve things! Let me spin up libre and see what's what. This project is just one of countless that leverages fastmcp (by far the most popular python mcp sdk) oauth capability so I have to think librechat will get their shit together rather than every one of those MCPs adding workarounds.

This pr is just a workaround, I don't pretend it's a good thing and I don't expect it to be merged. I agree a client fix for fastmcp 2.14+ compatibility would be much better!

owengo:
I have submitted https://github.com/danny-avila/LibreChat/pull/11067 which fixes the token exchange problem which makes the "compat google provider" useless.
But there is still the detection of the loggued user when tools are called and the service decorator modifications are still needed to recover the client email from the access token

taylorwilsdon:
Looks like librechat has fixed 👍

owengo:
> Looks like librechat has fixed 👍
@taylorwilsdon There is still a problem in gws with the detection of the logged-in user with OAuth2.1, and this patch fixes it

PR #321, opened 2025-12-23
enh: Pluggable storage backends, valkey oauth storage

owengo:
The oauth 2.1 "stateless" mode is not completely stateless, 
In OAuth2.1 (FastMCP GoogleProvider) mode, Google refresh tokens are stored server-side by FastMCP’s OAuth proxy in an encrypted key/value store.
Default location: ${FASTMCP_HOME}/oauth-proxy/ . FASTMCP_HOME defaults to FastMCP’s user data dir (settings.py (line 183), typically ~/.local/share/fastmcp on Linux).

So the access tokens and the refresh tokens are stored locally, which might not be desirable, and if you run several instances in parallel ( k8s deployment etc ), they won't have a shared access to them.

This PR allows to creates a ValKey backed FastMCP client_storage which solves the problem.

Note that the tokens are encrypted like in the file storage, tls is supported, I am using it with AWS' serverless valkey storage version 8.1 and it is working fine.

The code is 100% codex generated.

### DIFF
--- a/.env.oauth21
+++ b/.env.oauth21
@@ -14,3 +14,26 @@ OAUTH2_ENABLE_DEBUG=false
 
 # Legacy Compatibility (recommended during migration)
 OAUTH2_ENABLE_LEGACY_AUTH=true
+
+# ---------------------------------------------------------------------------
+# FastMCP OAuth Proxy storage (OAuth 2.1)
+#
+# By default, FastMCP stores OAuth proxy state on disk under:
+#   ${FASTMCP_HOME}/oauth-proxy
+#
+# To store OAuth proxy state in Valkey instead, configure:
+#
+# WORKSPACE_MCP_OAUTH_PROXY_STORAGE_BACKEND=valkey
+# WORKSPACE_MCP_OAUTH_PROXY_VALKEY_HOST=localhost
+# WORKSPACE_MCP_OAUTH_PROXY_VALKEY_PORT=6379
+# WORKSPACE_MCP_OAUTH_PROXY_VALKEY_USE_TLS=false
+# WORKSPACE_MCP_OAUTH_PROXY_VALKEY_DB=0
+# WORKSPACE_MCP_OAUTH_PROXY_VALKEY_USERNAME=
+# WORKSPACE_MCP_OAUTH_PROXY_VALKEY_PASSWORD=
+# WORKSPACE_MCP_OAUTH_PROXY_VALKEY_REQUEST_TIMEOUT_MS=5000
+# WORKSPACE_MCP_OAUTH_PROXY_VALKEY_CONNECTION_TIMEOUT_MS=10000
+#
+# Encryption:
+# - Values are encrypted with the same Fernet scheme FastMCP uses for its default disk store.
+# - Key material is derived from FASTMCP_SERVER_AUTH_GOOGLE_JWT_SIGNING_KEY if set, otherwise GOOGLE_OAUTH_CLIENT_SECRET.
+# - For stable decryption across client-secret rotations, set FASTMCP_SERVER_AUTH_GOOGLE_JWT_SIGNING_KEY explicitly.
--- a/core/server.py
+++ b/core/server.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import List, Optional
 from importlib import metadata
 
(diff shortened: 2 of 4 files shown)


taylorwilsdon:
Very interesting idea! I actually saw that as of v2.13.0, FastMCP has new support for pluggable storage backends with persistent state. 

From the [release](https://github.com/jlowin/fastmcp/releases/tag/v2.13.0):
* Built on [py-key-value-aio](https://github.com/strawgate/py-key-value), a new library from FastMCP maintainer Bill Easton ([@strawgate](https://github.com/strawgate)), the storage layer provides encrypted disk storage by default, platform-aware token management, and a simple key-value interface for application state. 
* We're excited to bring this elegantly designed library into the FastMCP ecosystem - it's both powerful and remarkably easy to use, including wrappers to add encryption, TTLs, caching, and more to backends ranging from Elasticsearch, Redis, DynamoDB, filesystem, in-memory, and more! OAuth providers now automatically persist tokens across restarts, and developers can store arbitrary state without reaching for external databases. 
* This foundation enables long-running sessions, cached credentials, and stateful applications built on MCP.

Might be preferable to implement that generically rather than a valkey specific implementation so that folks can use their choice of backend. Let me take a look today and see what's what because I think one way or another, this is worth doing.

owengo:
> * Built on [py-key-value-aio](https://github.com/strawgate/py-key-value), a new library from FastMCP maintainer Bill Easton ([@strawgate](https://github.com/strawgate)), the storage layer provides encrypted disk storage by default, platform-aware token management, and a simple key-value interface for application state.
> Might be preferable to implement that generically rather than a valkey specific implementation so that folks can use their choice of backend. Let me take a look today and see what's what because I think one way or another, this is worth doing.

This is exactly what's being used ( ie: fastmcp implementation of client_storage for valkey ). You can implement other storages if you want. Valkey is compatible with redis and it's quite common to have redis or valkey caching.

Anyway the important point, IMHO, is that currently the "stateless oautth2.1" mode is broken, it's not stateless, tokens are stored locallly, horizontal scaling is not possible etc. So if someone wants to run a redundant / distributed / k8s-based google_workspace_mcp MCP server, this patch provides a solution, fully based on FastMCP 2.13+ implementation

The proposed implementation manages most "best pratices" use cases for the valkey storage:
- encryption ( the same used by fastmcp when using local storage ) before sending to valkey ( in case you don't want to trust the backend )
- encryption in transit
- host, port configuration

Further improvement would perhaps manage read-only connections on port 6380 to aws serverless implementation but I don't know if it's supported by fastmcp asof today.

image

image

taylorwilsdon:
> > * Built on [py-key-value-aio](https://github.com/strawgate/py-key-value), a new library from FastMCP maintainer Bill Easton ([@strawgate](https://github.com/strawgate)), the storage layer provides encrypted disk storage by default, platform-aware token management, and a simple key-value interface for application state.
> >   Might be preferable to implement that generically rather than a valkey specific implementation so that folks can use their choice of backend. Let me take a look today and see what's what because I think one way or another, this is worth doing.
> 
> This is exactly what's being used ( ie: fastmcp implementation of client_storage for valkey ). You can implement other storages if you want. Valkey is compatible with redis and it's quite common to have redis or valkey caching.
> 
> Anyway the important point, IMHO, is that currently the "stateless oautth2.1" mode is broken, it's not stateless, tokens are stored locallly, horizontal scaling is not possible etc. So if someone wants to run a redundant / distributed / k8s-based google_workspace_mcp MCP server, this patch provides a solution, fully based on FastMCP 2.13+ implementation
> 
> The proposed implementation manages most "best pratices" use cases for the valkey storage:
> 
> * encryption ( the same used by fastmcp when using local storage ) before sending to valkey ( in case you don't want to trust the backend )
> * encryption in transit
> * host, port configuration
> 
> Further improvement would perhaps manage read-only connections on port 6380 to aws serverless implementation but I don't know if it's supported by fastmcp asof today.
> 
> image image

Yeah that's fantastic, I saw you're using the new mechanism! Just meant offering more options than just valkey since it's plug and play with all the above. Will take a run through today.

taylorwilsdon:
I'm blocked from updating the branch so I need to spin it out into a PR I can modify. Will track in https://github.com/taylorwilsdon/google_workspace_mcp/pull/328

PR #325, opened 2025-12-26
fix: remove invalid supportsAllDrives param from export_media()

usiegj00:
## Summary

The `export_doc_to_pdf` tool was failing with:
```
Error: Failed to export document to PDF: Got an unexpected keyword argument supportsAllDrives
```

## Root Cause

The Google Drive API's `export_media()` method does not accept the `supportsAllDrives` parameter. This parameter is valid for methods like `get()`, `list()`, and `create()`, but not for `export_media()`.

## Before (broken)
```python
request_obj = service.files().export_media(
    fileId=document_id, mimeType="application/pdf", supportsAllDrives=True
)
```

**Result:** `TypeError: Got an unexpected keyword argument supportsAllDrives`

## After (fixed)
```python
# Note: export_media() does not accept supportsAllDrives parameter
# (unlike get(), list(), etc.). The file access is already validated above.
request_obj = service.files().export_media(
    fileId=document_id, mimeType="application/pdf"
)
```

**Result:** PDF exports successfully

## Testing

Tested with Google Docs stored in a shared drive - multiple documents (1.5MB+ each) exported successfully.

The fix works because the file access is already validated by the preceding `files().get()` call (line 1239-1248) which correctly uses `supportsAllDrives=True`.

### DIFF
--- a/gdocs/docs_tools.py
+++ b/gdocs/docs_tools.py
@@ -1259,7 +1259,6 @@ async def export_doc_to_pdf(
 
     # Export the document as PDF
     try:
-        # Note: export_media() does not accept supportsAllDrives parameter
         request_obj = service.files().export_media(
             fileId=document_id, mimeType="application/pdf"
         )


gdocs/docs_tools.py
```diff
@@ -1259,8 +1259,10 @@ async def export_doc_to_pdf(
 
     # Export the document as PDF
     try:
+        # Note: export_media() does not accept supportsAllDrives parameter
```
taylorwilsdon:
```suggestion
```

gdocs/docs_tools.py
```diff
@@ -1259,8 +1259,10 @@ async def export_doc_to_pdf(
...
+        # (unlike get(), list(), etc.). The file access is already validated above.
```
taylorwilsdon:
```suggestion
```

taylorwilsdon:
Thanks!

PR #332, opened 2025-12-30
fix: prioritize user_google_email over session-based credential lookup

tyadon:
## Summary

Fixes multi-account credential lookup in `streamable-http` transport mode. When using multiple Google accounts, API operations were incorrectly using the first authenticated account's credentials regardless of which `user_google_email` was specified.

**This PR contains two fixes:**
1. `auth/google_auth.py` - Fixes the OAuth 2.0 legacy path in `get_credentials()`
2. `auth/service_decorator.py` - Fixes the OAuth 2.1 path in `get_authenticated_google_service_oauth21()`

## Problem

When running with `streamable-http` transport and multiple authenticated Google accounts:

1. First request authenticates user A and binds that user to the MCP session
2. Subsequent requests for user B call API with `user_google_email="user-b@gmail.com"`
3. Credential lookup finds user A's credentials (either via session ID or access token)
4. Returns user A's credentials, ignoring the explicit `user_google_email` parameter
5. All operations silently use user A's account

**Example**: Querying calendar events for `trey@pourmemory.com` returned events from `richardfyadon@gmail.com`.

### Fix 1: `get_credentials()` in `auth/google_auth.py`
Changed to prioritize explicit user lookup:
1. **PRIORITY 1**: If `user_google_email` is specified, look up credentials for that specific user first
2. **PRIORITY 2**: Only fall back to session-based lookup if no `user_google_email` was provided

### Fix 2: `get_authenticated_google_service_oauth21()` in `auth/service_decorator.py`
Changed to prioritize explicit user lookup:
1. **PRIORITY 1**: If `user_google_email` is specified, look up credentials for that specific user first from the session store
2. **PRIORITY 2**: Fall back to access token provider for single-account scenarios
3. **PRIORITY 3**: Fall back to session store validation

Both fixes ensure that when you call any API with `user_google_email="account-b@gmail.com"`, it actually uses account-b's credentials.

## Testing

Tested with 3 Google accounts in streamable-http mode:
- Before fix: Calendar events returned from wrong account
- After fix: Calendar events returned from correct specified account

## Backwards Compatibility

- Single-account usage: No change in behavior (session lookup still works as fallback)
- Multi-account usage: Now correctly uses the specified account

### DIFF
--- a/auth/service_decorator.py
+++ b/auth/service_decorator.py
@@ -213,10 +213,40 @@ async def get_authenticated_google_service_oauth21(
 ) -> tuple[Any, str]:
     """
     OAuth 2.1 authentication using the session store with security validation.
+
+    IMPORTANT: When user_google_email is explicitly provided, we prioritize looking up
+    credentials for that specific user. This enables multi-account support where a single
+    MCP session can access multiple pre-authenticated Google accounts.
     """
     provider = get_auth_provider()
     access_token = get_access_token()
 
+    # PRIORITY 1: If user_google_email is explicitly provided, use it
+    # This is the key fix for multi-account support - the explicitly requested
+    # account takes priority over the session's authenticated account
+    if user_google_email:
+        store = get_oauth21_session_store()
+
+        # Try to get credentials for the explicitly requested user
+        credentials = store.get_credentials(user_google_email)
+        if credentials:
+            logger.info(
+                f"[{tool_name}] Found credentials for explicitly requested user {user_google_email}"
+            )
+
+            # Validate scopes
+            scopes_available = set(credentials.scopes or [])
+            if not all(scope in scopes_available for scope in required_scopes):
+                raise GoogleAuthenticationError(
+                    f"OAuth credentials for {user_google_email} lack required scopes. "
+                    f"Need: {required_scopes}, Have: {sorted(scopes_available)}"
+                )
+
+            service = build(service_name, version, credentials=credentials)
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
I think this is just a misunderstanding of how the MCP works - `user_google_email` isn’t exposed at all in oauth2.1 mode, I think you may be confused - when running oauth2.1 mode, the client sends the authenticated user token with every request and no email value is sent by the LLM. That is the preferable way to use the MCP in general. If you have a client that isn’t capable of handling oauth, then you can run the legacy compatibility mode that starts a minimal oauth callback server and handles the flow, stores the creds on disk and does some basic session handling to associate the user with the cred store file.

That mode is only suitable for environments where you are the only user (even if you have multiple Google accounts, which is why it allows the value to be passed for user Google email). If you store creds and allow any user to pass any arbitrary email along with your request, that creates an enormous security vulnerability.  All multi user environments must use OAuth mode, and when you’re using that mode, the client is handling auth and the server is not doing anything but passing along the token your client includes with the request to google. There are dozens of 1k-20k employee companies using this in production today with their whole user population. If you can share your env var config for the server (with any client ids/secrets redacted) I can help figure out what’s missing or needs to be changed!

tyadon:
Hey, thanks for the explanation. You're right that I was using it differently than intended. I'm a single user with multiple Google accounts (work/personal) wanting to access both from one session. Sounds like OAuth 2.0 mode is the right fit for that use case. I'll stick with that approach. Appreciate the help.

PR #336, opened 2026-01-05
fix: persist OAuth states to disk to survive server restarts

Skeptomenos:
## Summary

OAuth authentication was failing with "Invalid or expired OAuth state parameter" when the MCP server restarted during an OAuth flow. This happened because OAuth states were stored only in memory (`self._oauth_states` dict) and lost on restart.

This is a common issue in stdio MCP mode where the server process may restart between when the user clicks the auth link and when they complete the OAuth flow in the browser.

## Changes

- Add file-based persistence for OAuth states to `~/.google_workspace_mcp/oauth_states.json`
- Load persisted states on `OAuth21SessionStore` initialization
- Clean up expired states on load
- States now survive server restarts

## Testing

Added 8 test cases in `tests/test_oauth_state_persistence.py`:
- `test_get_oauth_states_file_path_with_env_var`
- `test_store_oauth_state_persists_to_disk`
- `test_load_oauth_states_from_disk`
- `test_expired_states_cleaned_on_load`
- `test_validate_and_consume_persists_deletion`
- `test_state_survives_store_recreation` (key test - simulates server restart)
- `test_handles_missing_file_gracefully`
- `test_handles_corrupted_file_gracefully`

All tests pass:
```
============================= 8 passed in 2.43s ===============================
```

## Backward Compatibility

- If the state file doesn't exist, the store works exactly as before (in-memory only)
- Corrupted files are handled gracefully (logged warning, continues with empty state)
- No changes to the public API


taylorwilsdon:
This capability is already available via disk or valkey/redis through the [OAuth Proxy Storage Backends](https://github.com/taylorwilsdon/google_workspace_mcp?tab=readme-ov-file#oauth-proxy-storage-backends) capability.

PR #339, opened 2026-01-05
feat(gmail): Add Date header to message content tools

DixonDs:
This PR adds the `Date` header to the output of `get_gmail_message_content` and `get_gmail_messages_content_batch`.

### Why this change is needed
Currently, these tools return the sender, subject, and body, but omit the timestamp of the email. Including the `Date` header provides essential context, such as knowing when a booking was made or when a correspondence occurred.

### Consistency
This change also brings consistency with the thread retrieval tools (e.g., `get_gmail_thread_content`), which already include the date in their output. Unified behaviour across message and thread tools improves the developer experience and reliability for agents consuming these APIs.

### Changes
- Updated `GMAIL_METADATA_HEADERS` to include `"Date"`.
- Updated `get_gmail_message_content` to format and return the date.
- Updated `get_gmail_messages_content_batch` to format and return the date in both `full` and `metadata` modes.
- Updated docstrings to reflect the new field.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -31,7 +31,7 @@ logger = logging.getLogger(__name__)
 GMAIL_BATCH_SIZE = 25
 GMAIL_REQUEST_DELAY = 0.1
 HTML_BODY_TRUNCATE_LIMIT = 20000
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
Seems like a sensible improvement, thanks!

PR #340, opened 2026-01-06
Added tools for marking emails as read, and archiving emails in batch.

cyzhao:


taylorwilsdon:
Hey there, you can already accomplish this with the existing tools - there's `modify_gmail_message_labels` and `batch_modify_gmail_message_labels` and they work without any special instructions for this specific purpose - thanks!

PR #363, opened 2026-01-16
feat: Auto-detect user_google_email from OAuth credentials

Coldaine:
## Summary

This PR improves the developer experience by automatically detecting the authenticated user's email from stored OAuth credentials, eliminating the need to configure `USER_GOOGLE_EMAIL` for single-account setups.

## Backward Compatibility Approach

The `user_google_email` parameter is **intentionally kept as-is** in all tool signatures (required, same position) to maintain backward compatibility with existing configurations. However, the runtime behavior now:

1. **Auto-detects first** - If only one Google account is authenticated, that email is used automatically (the provided parameter value is ignored)
2. **Falls back to provided value** - For multi-account setups where auto-detection is ambiguous, the provided `user_google_email` parameter is used
3. **Errors helpfully** - If neither works, provides actionable guidance

This means:
- ✅ Existing tool configs continue to work without modification
- ✅ New users don't need to configure `USER_GOOGLE_EMAIL` at all
- ✅ Multi-account users can still specify which account to use

## Why This Approach?

Changing function signatures (making `user_google_email` optional with defaults) would require reordering parameters, which is a **breaking change** for positional argument calls. By keeping signatures unchanged and modifying only the decorator behavior, we achieve the UX improvement while maintaining **full backward compatibility**.

## Changes

- Add lazy email auto-detection in `core/config.py` with thread-safe caching
- Modify `_extract_oauth20_user_email` decorator to prefer auto-detection
- Fix typo in `draft_gmail_message` docstring example

## Test plan

- [ ] Verify auto-detection works with a single credential file
- [ ] Verify provided `user_google_email` is used when multiple accounts exist
- [ ] Verify error message with multiple credential files lists available accounts
- [ ] Verify error message with no credentials guides user to authenticate
- [ ] Verify existing configurations continue to work unchanged

### DIFF
--- a/auth/service_decorator.py
+++ b/auth/service_decorator.py
@@ -315,29 +315,80 @@ def _extract_oauth21_user_email(
 
 
 def _extract_oauth20_user_email(
-    args: tuple, kwargs: dict, wrapper_sig: inspect.Signature
+    args: tuple, kwargs: dict, wrapper_sig: inspect.Signature, func_name: str = ""
 ) -> str:
     """
-    Extract user email for OAuth 2.0 mode from function arguments.
+    Extract user email for OAuth 2.0 mode with automatic detection.
+
+    BACKWARD COMPATIBILITY: The user_google_email parameter is kept for compatibility
+    with existing configurations, but is now effectively optional in single-account
+    setups. When only one Google account is authenticated, the email is auto-detected
+    and the provided parameter value is ignored.
+
+    Detection priority:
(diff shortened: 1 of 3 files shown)


taylorwilsdon:
@Coldaine can you give the current release a try and see if it does what you need already? Token extraction logic was added in an unrelated OAuth patch.

Coldaine:
@taylorwilsdon  Appreciate it, I will give it a try. Apologies, I'd contributed this and forgot to follow up.

Issue #338, opened 2026-01-05
[Question] LLM asking for user's email in single user mode


lfnandoo:
I'm having a issue (quite similar to [301](https://github.com/taylorwilsdon/google_workspace_mcp/issues/310)) that the LLM is asking for my email before calling tools even w/ single mode active and default email given.
I launched my server with this config:
```json
{
        "command": "uvx",
        "args": ["workspace-mcp", "--tools", "drive", "docs", "--single-user"],
        "env": {
            "GOOGLE_OAUTH_CLIENT_ID": "REDACTED",
            "GOOGLE_OAUTH_CLIENT_SECRET": "REDACTED",
            "USER_GOOGLE_EMAIL": "foo@bar.com"
         }
}
```

Looking more deeply, I figured out that the `user_google_email` is still flagged as required in the tool schema:
```json
{
  "name": "get_drive_file_download_url",
  "description": "Gets a download URL for a Google Drive file. The file is prepared and made available via HTTP URL.\n\nFor Google native files (Docs, Sheets, Slides), exports to a useful format:\n• Google Docs → PDF (default) or DOCX if export_format='docx'\n• Google Sheets → XLSX (default) or CSV if export_format='csv'\n• Google Slides → PDF (default) or PPTX if export_format='pptx'\n\nFor other files, downloads the original file format.\n\nArgs:\n    user_google_email: The user's Google email address. Required.\n    file_id: The Google Drive file ID to get a download URL for.\n    export_format: Optional export format for Google native files.\n                  Options: 'pdf', 'docx', 'xlsx', 'csv', 'pptx'.\n                  If not specified, uses sensible defaults (PDF for Docs/Slides, XLSX for Sheets).\n\nReturns:\n    str: Download URL and file metadata. The file is available at the URL for 1 hour.",
  "input_schema": {
    "properties": {
      "user_google_email": {
        "type": "string"
      },
      "file_id": {
        "type": "string"
      },
      "export_format": {
        "anyOf": [
          {
            "type": "string"
          },
          {
            "type": "null"
          }
        ],
        "default": null
      }
    },
    "required": [
      "user_google_email",
      "file_id"
    ],
    "type": "object"
  }
}
```

I was expecting that giving the default email + single user could make either this optional or remove this from schema.

Is it a miss understanding, a bug or a feature request?

1amnotmad:
[...]

PR #370, opened 2026-01-19
Fix: resolve conflict between Single User Mode and OAuth 2.1

jhutar:
- Ensure `--single-user` CLI flag takes precedence over `MCP_ENABLE_OAUTH21=true` in .env
- Early setting of MCP_SINGLE_USER_MODE to influence configuration loading
- Remove `user_google_email` from tool signatures and docstrings in single-user mode
- Allow implicit authentication in single-user mode by relaxing email validation

Original objective: Fix google-workspace-mcp server error 'OAuth 2.1 mode requires an authenticated user' when running in single-user mode. Resolve conflict where MCP_ENABLE_OAUTH21=true in .env blocks --single-user mode, and make user_google_email optional in tool schemas for single-user operation.

Generated-by: Gemini

Note this is related to https://github.com/taylorwilsdon/google_workspace_mcp/issues/338

### DIFF
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -849,11 +849,16 @@ async def get_authenticated_google_service(
         f"[{tool_name}] Attempting to get authenticated {service_name} service. Email: '{user_google_email}', Session: '{session_id}'"
     )
 
-    # Validate email format
-    if not user_google_email or "@" not in user_google_email:
+    # Validate email format (unless single-user mode, where it can be None)
+    is_single_user = os.getenv("MCP_SINGLE_USER_MODE") == "1"
+    if not is_single_user and (not user_google_email or "@" not in user_google_email):
         error_msg = f"Authentication required for {tool_name}. No valid 'user_google_email' provided. Please provide a valid Google email address."
         logger.info(f"[{tool_name}] {error_msg}")
         raise GoogleAuthenticationError(error_msg)
+    elif is_single_user and not user_google_email:
+        logger.debug(
+            f"[{tool_name}] Single-user mode: user_google_email is None, will try to find any credentials."
+        )
 
     credentials = await asyncio.to_thread(
         get_credentials,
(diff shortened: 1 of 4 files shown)


taylorwilsdon:
Hm - single user mode wasn't ever intended to be used alongside oauth 2.1 mode. Does this actually work? I'm not opposed to it at all conceptually, seems like a sensible addition but the logic wasn't designed for this to work that way.

jhutar:
Hello! It works for me with this `.env`:
```
GOOGLE_OAUTH_CLIENT_ID=...
GOOGLE_OAUTH_CLIENT_SECRET=...
GOOGLE_CLIENT_SECRET_PATH=/home/...

WORKSPACE_MCP_BASE_URI=http://localhost
WORKSPACE_MCP_PORT=8000
MCP_ENABLE_OAUTH21=true

USER_GOOGLE_EMAIL=jhutar@redhat.com
```

And this config:

```
{
  "mcpServers": {

    "google-workspace-mcp": {
      "command": "uv",
      "args": [
        "run",
        "main.py",
        "--single-user",
        "--tool-tier", "complete",
        "--tools", "calendar", "docs", "drive", "sheets", "slides", "tasks"
      ],
      "cwd": "/home/jhutar/Checkouts/google_workspace_mcp/"
    }

  }
}
```

But as I said, I kinda failed to understand the auth logic and "best practice" .env, so not saying this is a right thing to do.

taylorwilsdon:
Single user mode doesn't make any sense in the context of OAuth mode. It's just there to handle user_google_email automatically for legacy clients. With OAuth mode, you pass a token with each request and the user in question is extracted from that, not passed by the LLM. Don't ever use the two together (as I mentioned, I didn't realize it would even work)

jhutar:
If I drop patch in this PR and remove "--single-user" from mine config, Gemini-cli complains:
```
MCP tool 'get_doc_content' reported tool error for function call: {"name":"get_doc_content","args":{"document_id":"..."}} with response: [{"functionResponse":{"name":"get_doc_content","response":{"content":[{"type":"text","text":"Error calling tool 'get_doc_content': An unexpected error occurred in get_doc_content: OAuth 2.1 mode requires an authenticated user for get_doc_content, but none was found."}],"isError":true}}}]
```
When I do `/mcp auth google-workspace-mcp`, I get:
```
ℹ Starting OAuth authentication for MCP server 'google-workspace-mcp'...
✕ Failed to authenticate with MCP server 'google-workspace-mcp': Cannot perform dynamic registration without authorization URL
```

jhutar:
If you could please give me exact config I should use, I can instead of this PR create a PR that would make the MCP server fail to start if somebody is trying same misconfiguration as I'm: Single user mode (represented by `--single-user` option) together with OAuth mode (represented by `MCP_ENABLE_OAUTH21=true` variable, or something else?).

taylorwilsdon:
I added a startup check that blocks single user mode when paired with OAuth mode. Pull down the latest version (1.8.1) and remove the single user flag and it should work as expected! From the config you shared, changes needed are:

* remove USER_GOOGLE_EMAIL (doesn't apply to OAuth mode)
* add OAUTHLIB_INSECURE_TRANSPORT=1 since you're running locally (no HTTPS)
* remove GOOGLE_CLIENT_SECRET_PATH since you're declaring client ID and client secret already. That's an either / or - you use the env vars for client ID and secret, or you point it at a file that contains them (not both)

```
GOOGLE_OAUTH_CLIENT_ID=...
GOOGLE_OAUTH_CLIENT_SECRET=...

WORKSPACE_MCP_BASE_URI=http://localhost
WORKSPACE_MCP_PORT=8000
OAUTHLIB_INSECURE_TRANSPORT=1
MCP_ENABLE_OAUTH21=true

```

jhutar:
Awesome, thank you!

PR #427, opened 2026-02-06
fix: validate session user matches requested email in multi-account setups

jeremi:
## Summary

- When multiple Google accounts are configured, `get_credentials()` returns the first account's session credentials regardless of which `user_google_email` is passed
- Adds email validation after both session-based credential lookups (OAuth 2.1 store in Phase 1, and `load_credentials_from_session` in Phase 3a) so mismatched credentials fall through to the file-based store
- Adds 7 unit tests covering multi-user credential resolution scenarios

## Problem

The session-based credential cache in `get_credentials()` binds to whichever account authenticated first and ignores the `user_google_email` parameter on subsequent calls. This means calling with `user_a@example.com` and then `user_b@example.com` returns user A's data for both.

**Root cause:** Two session lookups (`store.get_credentials_by_mcp_session()` at line 573 and `load_credentials_from_session()` at line 656) return credentials by session ID without checking if they match the requested email. The correct file-based fallback that uses `user_google_email` is never reached.

## Fix

After each session-based credential lookup, check if the session's bound user matches the requested `user_google_email`. If they don't match, set `credentials = None` so execution falls through to the file-based store.

## Test plan

- [x] 7 new unit tests in `tests/auth/test_multi_user_credentials.py`
- [x] All 60 existing tests still pass
- [X] Manual verification: install fork and test with two different Google accounts — each should return its own data

### DIFF
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -572,6 +572,17 @@ def get_credentials(
             # Try to get credentials by MCP session
             credentials = store.get_credentials_by_mcp_session(session_id)
             if credentials:
+                # Validate that session credentials match the requested user
+                if user_google_email:
+                    session_user = store.get_user_by_mcp_session(session_id)
+                    if session_user and session_user != user_google_email:
+                        logger.info(
+                            f"[get_credentials] Session {session_id} is bound to {session_user}, "
+                            f"but credentials requested for {user_google_email}. Skipping session credentials."
+                        )
+                        credentials = None
+
+            if credentials:
                 logger.info(
                     f"[get_credentials] Found OAuth 2.1 credentials for MCP session {session_id}"
                 )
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
I think you're mixing up concepts - `user_google_email` (the tool param) is not used at all in OAuth 2.1 mode. It's actually removed from the args dynamically on startup.  In `auth/service_decorator.py,` the `require_google_service` decorator removes `user_google_email` from the public function signature when OAuth 2.1 is enabled. The email is instead extracted from the authenticated context (the verified token). A caller cannot pass in a different email to cause a mismatch.

In OAuth mode, the client handles the oauth flow callback and stores the user's token, then passes it with each request to the MCP server and the active user is determined by whatever the token unpacks to. The session binding is immutable so we can take for granted that `store.get_credentials_by_mcp_session(session_id)` will always return the right credentials for that session.  Past that, you don't need to do any additional validation because simply by virtue of making a call to Google, the token is checked and validated. 

`get_credentials` isn't a standalone function, it's a two step lookup invoked by `get_credentials_by_mcp_session`, which does not accept the user's email as an input - the user email is determined by `self._mcp_session_mapping.get(mcp_session_id)` (which again is immutable). The _mcp_session_mapping is populated at session creation time and is immutable — if anything tries to rebind a session ID to a different user, a ValueError is raised.

The scenario "calling with user_a@example.com then user_b@example.com returns user A's data for both" cannot happen because:
  1. In OAuth 2.1 mode, the caller doesn't choose which email to pass; it comes from the verified token
  2. Each MCP session is immutably bound to one user
  3. Existing validation already rejects cross-user access attempts

PR #432, opened 2026-02-07
fix(auth): add cross-service Drive scopes for docs and sheets tools

Milofax:
## Summary

Several docs and sheets tools internally use the Google Drive API but don't receive the necessary Drive OAuth scopes when configured with `--tools docs sheets` (without explicitly adding `drive`):

- **Docs tools affected:** `search_docs`, `get_doc_content`, `list_docs_in_folder`, `export_doc_to_pdf`
- **Sheets tools affected:** `list_spreadsheets`

These tools use `@require_google_service("drive", ...)` internally, which calls `build("drive", "v3", credentials)`. The credentials work because the service builder doesn't gate on `--tools`, but the OAuth token lacks the required Drive scopes — causing API 403 errors at runtime.

## Fix

Add minimal cross-service Drive scopes to `DOCS_SCOPES` and `SHEETS_SCOPES` in `auth/scopes.py`:

- **docs:** `drive.readonly` (for metadata queries) + `drive.file` (for PDF export)
- **sheets:** `drive.readonly` (for spreadsheet listing)

This follows the existing pattern where `SCRIPT_SCOPES` (appscript) already includes `DRIVE_FILE_SCOPE` as a cross-service dependency (line 127).

Importantly, this does **not** add `drive` (full access) — only the minimal scopes needed. The alternative workaround of adding `--tools drive` exposes 14 full-access Drive tools, which is undesirable from a least-privilege perspective.

## Changes

| File | Change |
|------|--------|
| `auth/scopes.py` | Add `DRIVE_READONLY_SCOPE` + `DRIVE_FILE_SCOPE` to `DOCS_SCOPES`, `DRIVE_READONLY_SCOPE` to `SHEETS_SCOPES`, and corresponding read-only variants |
| `tests/test_scopes.py` | 10 unit tests verifying scope generation for docs, sheets, combined, and read-only modes |

## Test plan

- [x] All 10 new scope tests pass
- [x] All 74 existing unit tests still pass (84 total passed, 8 pre-existing errors in `manual_test.py` fixture)
- [ ] Manual verification: `--tools docs sheets` (without `drive`) can successfully call `search_docs`, `list_spreadsheets`, `export_doc_to_pdf`


taylorwilsdon:
Hm, that's an interesting point. I think on one side, that may be concerning to users who think that they are only doing `x` and instead are getting a scope request that while legitimate is for `y`, but to your point the tools won't all work without it. I'm open to merging this, will need a ruff format but otherwise seems clean thanks!

Milofax:
Thanks for the review! Pushed a ruff format fix (821804c) — the `DOCS_SCOPES` list needed multi-line formatting.

CI shows `action_required` (fork PR approval needed). Once approved, all checks should pass — verified locally with `ruff format --check` and `ruff check`.

taylorwilsdon:
Thanks!

PR #445, opened 2026-02-10
feat(server): add HTTPS/TLS support via --ssl-cert and --ssl-key args

cfdude:
## Summary
- Add `--ssl-cert` and `--ssl-key` CLI args for native HTTPS support on streamable-http transport
- SSL params passed through to uvicorn via FastMCP's `uvicorn_config` parameter
- Auto-detect HTTPS scheme for `WORKSPACE_MCP_BASE_URI` default when SSL is enabled
- Add PM2 ecosystem config (`ecosystem.config.cjs`) for production HTTP/HTTPS deployment
- Fix `list_users()` in credential store to skip non-email files (e.g. `oauth_states.json`) that caused `NoneType` scopes error in single-user mode
- Add `certs/` to `.gitignore` for local mkcert certificates

## Test Plan
- [x] Server starts on HTTPS with `--ssl-cert` and `--ssl-key` flags
- [x] OAuth redirect URI correctly uses `https://localhost` scheme
- [x] Gmail search tool works end-to-end over HTTPS connection
- [x] Single-user mode correctly skips `oauth_states.json` in credential lookup
- [x] Validation rejects mismatched SSL args (cert without key, SSL without streamable-http)


taylorwilsdon:
Hm, the PR description and what's actually there to diverge and a number of conflicts @cfdude - are you thinking of using the fastmcp server as the user facing server and handling all SSL termination there? That seems like a very high risk endeavor, generally the pattern would be using a reverse proxy to serve only the endpoints exposed by the mcp server and setting the external url in the config for the mcp so that it handles the proxy gracefully.

cfdude:
Sorry Taylor, I think I got over jealous and accidentally did a PR to you instead of my origin branch - please close this for now.  I'm still working on this.

PR #460, opened 2026-02-11
fix: FieldInfo not JSON serializable in CLI mode for label functions

### LINKED ISSUE #440: Bug: FieldInfo is not JSON serializable in CLI mode for modify_gmail_message_labels
## Describe the bug

When invoking `modify_gmail_message_labels` or `batch_modify_gmail_message_labels` via `--cli` mode, the call fails with:

```
Error: Exception: An unexpected error occurred in modify_gmail_message_labels: Object of type FieldInfo is not JSON serializable
```

The same call works correctly via the native MCP server protocol.

## To Reproduce

```bash
workspace-mcp --cli modify_gmail_message_labels --args '{
  "message_id": "19bdbee2b1f60be2",
  "remove_label_ids": ["STARRED"]
}'
```

**Result:** `Object of type FieldInfo is not JSON serializable`

**Expected:** `Message labels updated successfully!`

## Root Cause

In `gmail/gmail_tools.py`, the functions `modify_gmail_message_labels` (line ~1820) and `batch_modify_gmail_message_labels` (line ~1876) use Pydantic `Field()` as default values for optional list parameters:

```python
async def modify_gmail_message_labels(
    service,
    user_google_email: str,
    message_id: str,
    add_label_ids: List[str] = Field(
        default=[], description="Label IDs to add to the message."
    ),
    remove_label_ids: List[str] = Field(
        default=[], description="Label IDs to remove from the message."
    ),
) -> str:
```

`Field()` returns a `FieldInfo` metadata object, not `[]`. When FastMCP serves this tool via the MCP protocol, Pydantic's model validation resolves `FieldInfo` into the actual default value — so the MCP server works fine. But when the CLI handler in `core/cli_handler.py` calls `fn(**args)` directly (line 241), Python uses the raw `FieldInfo` object as the default. This `FieldInfo` object then propagates until something tries to serialize or iterate over it, causing the crash.

## Proposed Fix

Replace `Field(default=[], description="...")` with plain `[]` defaults. The parameter descriptions are already documented in the docstring, which FastMCP uses for schema generation.

```diff
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ modify_gmail_message_labels
     service,
     user_google_email: str,
     message_id: str,
-    add_label_ids: List[str] = Field(
-        default=[], description="Label IDs to add to the message."
-    ),
-    remove_label_ids: List[str] = Field(
-        default=[], description="Label IDs to remove from the message."
-    ),
+    add_label_ids: List[str] = [],
+    remove_label_ids: List[str] = [],
 ) -> str:
```

Same change for `batch_modify_gmail_message_labels`.

**Note:** `Field(...)` (required marker, line ~1780) is unaffected — only `Field(default=...)` with optional defaults causes the issue.

## Affected Functions

| Function | File | Line |
|----------|------|------|
| `modify_gmail_message_labels` | `gmail/gmail_tools.py` | ~1820 |
| `batch_modify_gmail_message_labels` | `gmail/gmail_tools.py` | ~1876 |

Happy to submit a PR for this if you'd like.

liuxiaopai-ai:
## Summary

Fixes #440

`modify_gmail_message_labels` and `batch_modify_gmail_message_labels` crash with `Object of type FieldInfo is not JSON serializable` when invoked via `--cli` mode.

## Root Cause

`Field(default=[], description="...")` returns a `FieldInfo` metadata object, not `[]`. When FastMCP serves tools via MCP protocol, Pydantic resolves `FieldInfo` into the actual default — so the MCP server works fine. But when the CLI handler calls `fn(**args)` directly, Python uses the raw `FieldInfo` object as the default, which is not serializable.

## Fix

Replace `Field(default=[], description="...")` with plain `[]` defaults in both functions. The parameter descriptions are already documented in the docstrings, which FastMCP uses for schema generation.

## Changes

- `gmail/gmail_tools.py`: Replace `Field(default=...)` with `[]` for `add_label_ids` and `remove_label_ids` in:
  - `modify_gmail_message_labels` (line ~1832)
  - `batch_modify_gmail_message_labels` (line ~1888)

+4 / -12 lines.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -1829,12 +1829,8 @@ async def modify_gmail_message_labels(
     service,
     user_google_email: str,
     message_id: str,
-    add_label_ids: List[str] = Field(
-        default=[], description="Label IDs to add to the message."
-    ),
-    remove_label_ids: List[str] = Field(
-        default=[], description="Label IDs to remove from the message."
-    ),
+    add_label_ids: List[str] = [],
+    remove_label_ids: List[str] = [],
 ) -> str:
     """
     Adds or removes labels from a Gmail message.
@@ -1885,12 +1881,8 @@ async def batch_modify_gmail_message_labels(
     service,
     user_google_email: str,
     message_ids: List[str],
-    add_label_ids: List[str] = Field(
-        default=[], description="Label IDs to add to messages."
-    ),
-    remove_label_ids: List[str] = Field(
-        default=[], description="Label IDs to remove from messages."
-    ),
+    add_label_ids: List[str] = [],
+    remove_label_ids: List[str] = [],
 ) -> str:
     """
     Adds or removes labels from multiple Gmail messages in a single batch request.


taylorwilsdon:
Appreciate the PR but I beat ya to it this morning :D https://github.com/taylorwilsdon/google_workspace_mcp/pull/459

PR #462, opened 2026-02-12
fix(gmail): normalize Body params in send/draft tools

Manas-Taneja:
## Summary
Fixes Gmail compose tool failures where `send_gmail_message` and `draft_gmail_message` crash with errors like:
- `object of type 'Body' has no len()`
- `'Body' object has no attribute 'lower'`

These occur when FastAPI `Body(...)` default objects are passed through runtime invocation instead of concrete values.

## Changes
- Added defensive input normalization in `gmail/gmail_tools.py`:
  - `_unwrap_body_param(...)`
  - `_normalize_compose_inputs(...)`
- Applied normalization + validation in:
  - `send_gmail_message`
  - `draft_gmail_message`

## Tests
- Added regression tests in:
  - `tests/gmail/test_gmail_compose_input_normalization.py`
- Local run:
  - `pytest -q tests/gmail/test_gmail_compose_input_normalization.py` (3 passed)

## Notes
- This is a backward-compatible hardening fix.
- No behavior change for normal typed invocations; only normalizes problematic Body-param runtime objects.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -20,6 +20,7 @@ from email import encoders
 from email.utils import formataddr
 
 from fastapi import Body as BodyParam
+from fastapi.params import Body as FastApiBody
 from pydantic import Field
 
 from auth.service_decorator import require_google_service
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
Hm, this was introduced as a fix to a CLI bug reported. Let me take a look and see if this is actually what we need as a workaround, thanks.

taylorwilsdon:
Hm, I'm not able to reproduce via cli or via MCP - can you share a bit more information about the issue you're having @Manas-Taneja?

The CLI shim is not ideal and should probably be revisited but I'm not having any problems with actual usage. 

```
❯ draft a gmail message to redacted@redacted.com with subject "test" and
   body "test2"

⏺ workspace - draft_gmail_message (MCP)(subject: "test", body: "test2", to:
                                       "redacted@redacted.com")
  ⎿  {
       "result": "Draft created! Draft ID: r4958835616771357914"
     }

⏺ Draft created successfully. The draft (ID: r4958835616771357914) is now in
  your Gmail drafts folder, addressed to redacted@redacted.com with subject
  "test" and body "test2". You can review and send it from Gmail whenever you're
   ready.
```

taylorwilsdon:
I've got a new PR here - https://github.com/taylorwilsdon/google_workspace_mcp/pull/467 that cleans up that bodyparams logic and continues to work in CLI and MCP mode. If you have any issues, please open a github issue with details and logs!

Manas-Taneja:
Thanks for taking a look.

I can’t share my fork (private) and I didn’t save the original trace, but the failure mode is: these tools can be invoked directly (outside FastAPI request parsing / outside the CLI arg normalizer), so the default values (BodyParam(...)) can flow into runtime code as fastapi.params.Body objects.

That makes a couple spots blow up:
- _prepare_gmail_message: subject.lower() / body_format.lower() (so 'Body' object has no attribute 'lower')
- send_gmail_message logging / return strings: len(attachments) (so object of type 'Body' has no len())

Minimal repro (no auth/MCP needed):
```py
from fastapi import Body
from gmail.gmail_tools import _prepare_gmail_message

_prepare_gmail_message(subject=Body("test"), body="hi", body_format=Body("plain"))
```

Manas-Taneja:
> I've got a new PR here - #467 that cleans up that bodyparams logic and continues to work in CLI and MCP mode. If you have any issues, please open a github issue with details and logs!

I'll take a look, thanks

PR #502, opened 2026-02-23
fix: use RFC 2231 encoding for non-ASCII attachment filenames

### LINKED ISSUE #500: draft_gmail_message: Non-ASCII characters in attachment filenames result in "noname" display in Gmail
## Bug Description

When creating a Gmail draft with `draft_gmail_message` and attaching files whose names contain non-ASCII characters (e.g. German umlauts like ü, ö, ä, ß), Gmail displays "noname" instead of the actual filename.

## Steps to Reproduce

1. Create a Gmail draft with three PDF attachments:
   - `Prüfbericht Q1.pdf` (contains `ü`)
   - `Statusbericht Projekt.pdf` (ASCII only)
   - `Kostenübersicht März.pdf` (contains `ü` and `ä`)

2. Open the draft in Gmail web UI

## Expected Behavior

All three attachments should show their correct filenames.

## Actual Behavior

- `Statusbericht Projekt.pdf` → **displayed correctly** ✅
- `Prüfbericht Q1.pdf` → **displayed as "noname"** ❌
- `Kostenübersicht März.pdf` → **displayed as "noname"** ❌

The pattern is clear: only filenames containing non-ASCII characters (ü, ä) are affected. The files themselves are attached correctly (correct sizes), but the filename metadata is lost.

## Likely Cause

The `Content-Disposition` header in the MIME message likely does not properly encode non-ASCII filenames using RFC 2231 (`filename*=utf-8''...`). When non-ASCII bytes appear in the header without proper encoding, Gmail falls back to "noname".

Python's `email.mime` module handles this correctly when using `add_header()`:
```python
att.add_header('Content-Disposition', 'attachment', filename='Prüfbericht.pdf')
# Automatically generates: filename*=utf-8''Pr%C3%BCfbericht.pdf
```

## Environment

- workspace-mcp version: latest via `uvx workspace-mcp`
- OS: macOS
- Client: Claude Code CLI
- Tool tier: extended

## Workaround

Using ASCII-only filenames works (e.g. `Pruefbericht.pdf` instead of `Prüfbericht.pdf`).

For **Claude Code** users, a PreToolUse hook can automatically catch this before the API call is made. The hook:

1. Inspects attachment filenames in `tool_input` via stdin JSON
2. Blocks the call if non-ASCII characters are detected
3. Suggests the correct ASCII transliteration (ä→ae, ö→oe, ü→ue, ß→ss)
4. Allows the call if an explicit ASCII `filename` field is set

Alternatively, simply set an explicit `filename` parameter with ASCII-only characters for each attachment.

stakeswky:
## Summary

Fixes #500 — Gmail displays "noname" for attachments whose filenames contain non-ASCII characters (e.g. `Prüfbericht Q1.pdf`).

## Root Cause

`Content-Disposition` was constructed via string formatting:

```python
part.add_header('Content-Disposition', f'attachment; filename="{safe_filename}"')
```

This embeds raw non-ASCII bytes directly in the MIME header. Gmail cannot parse them and falls back to "noname".

## Fix

Pass the filename as a keyword argument to `add_header()`:

```python
part.add_header('Content-Disposition', 'attachment', filename=safe_filename)
```

Python's `email` library then automatically applies [RFC 2231](https://datatracker.ietf.org/doc/html/rfc2231) encoding for non-ASCII filenames:

```
Content-Disposition: attachment; filename*=utf-8''Pr%C3%BCfbericht%20Q1.pdf
```

ASCII-only filenames remain unchanged (`filename="Statusbericht Projekt.pdf"`).

## Verification

```python
from email.mime.base import MIMEBase
from email import encoders

part = MIMEBase('application', 'pdf')
part.set_payload(b'test')
encoders.encode_base64(part)
part.add_header('Content-Disposition', 'attachment', filename='Prüfbericht Q1.pdf')
print(part['Content-Disposition'])
# → attachment; filename*=utf-8''Pr%C3%BCfbericht%20Q1.pdf  ✅
```

The backslash/quote escaping that was previously applied is no longer needed — `add_header()` handles all quoting and encoding internally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Attachments with non-ASCII filenames now display correctly instead of appearing unnamed.
  * Filename handling improved: non-ASCII characters preserved, CR/LF/null chars removed, backslash/quote escaping avoided.
  * A sensible default filename ("attachment") is used when none is provided and encoding is handled properly for safe display.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -337,7 +337,7 @@ def _prepare_gmail_message(
                     .replace("\r", "")
                     .replace("\n", "")
                     .replace("\x00", "")
-                )
+                ) or "attachment"
 
                 part.add_header(
                     "Content-Disposition", "attachment", filename=safe_filename


taylorwilsdon:
Lgtm better than the copilot auto pr mess haha thanks!

stakeswky:
Thanks for the review! Happy to make any adjustments if needed.

PR #528, opened 2026-03-01
enh: Docs Checkbox Handling

taylorwilsdon:
Checkboxes in Google Docs are read as plain bullet items, so the LLM doesn't know they're checkboxes and writes literal [x]/DONE text. The fix is in docs_markdown.py - detect checklist items and render them as - [ ]/- [x] in markdown.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive checklist support for Google Docs to Markdown conversion. Lists are now properly detected as checklists and rendered with [x] for completed items and [ ] for pending items. Formatting maintains correct indentation throughout nested lists, and completed items render cleanly without redundant strikethrough in the converted output.

### DIFF
--- a/gdocs/docs_markdown.py
+++ b/gdocs/docs_markdown.py
@@ -5,6 +5,7 @@ Converts Google Docs API JSON responses to clean Markdown, preserving:
 - Headings (H1-H6, Title, Subtitle)
 - Bold, italic, strikethrough, code, links
 - Ordered and unordered lists with nesting
+- Checklists with checked/unchecked state
 - Tables with header row separators
 """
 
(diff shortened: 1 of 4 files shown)


taylorwilsdon:
Works nicely

dustinsmith1024:
So I have been trying this out and it seems like it wont read if someone actually checks the box in Google. It works if you tell the mcp to check a box then it read that. However I am hoping to have people click the boxes then read it back out. 

Is that happening for others?

taylorwilsdon:
It should properly handle for that when exported to markdown @dustinsmith1024 but I'm wondering if maybe that's not the case for just reading the doc? Can you try retrieve as markdown and see if you get the same behavior?

PR #556, opened 2026-03-11
Auto-open browser for OAuth and add login_hint

### LINKED ISSUE #693: Feature Request: login_hint support for multi-account OAuth 2.1 deployments
I manage multiple Google accounts (e.g., personal@gmail.com, work@business.com, volunteer@org.com) and I want all of them accessible via mcp, ideally from a single Claude instance. 

However, I keep each account in its own browser profile, to maintain easier separation when personally working in them. For example, I don't authenticate work@business.com in my personal@gmail.com browser profile.

Initially, I'm able to successful add accounts to my Claude session one at a time and authenticate them successfully, because I know which profile to use and which credentials to provide. 

But eventually after my initial authentication with each profile, I get prompted to re-authenticate (e.g., for a fresh mcp call or after restarting Claude). At that point, there's no indication about which credentials I should be providing.

Do you have a suggestion for that workflow? And/or any recommended workaround? 
Below is what Claude and I have come up with as an idea.

Thank you for this project!!!

---
ENVIRONMENT
- Transport: streamable-http
- OAuth mode: OAuth 2.1 (MCP_ENABLE_OAUTH21=true, WORKSPACE_MCP_STATELESS_MODE=true)
- Deployment: self-hosted behind nginx reverse proxy
- Client: Claude Desktop via mcp-remote

PROBLEM
When running a single workspace-mcp instance serving multiple Google accounts in OAuth 2.1 multi-user mode, the OAuth consent URL that opens in the browser contains no hint about which Google account should be used. 

The consent URL looks like:
https://mcp.domain.com/consent?txn_id=<opaque_token>
There is no way to tell from the URL — or anywhere in the browser flow — which account is being requested. If you're signed into multiple Google accounts, you're guessing.

One practical workaround is to configure multiple Claude Desktop MCP entries pointing at the same server with unique query parameters (e.g. ?account=personal, ?account=work, ?account=org) so that mcp-remote treats them as distinct servers with separate token caches:
json"workspace-mcp-personal": {
  "command": "npx",
  "args": ["-y", "mcp-remote", "https://mcp.domain.com/mcp?account=personal"]
},
"workspace-mcp-work": {
  "command": "npx",
  "args": ["-y", "mcp-remote", "https://mcp.domain.com/mcp?account=work"]
},
"workspace-mcp-org": {
  "command": "npx",
  "args": ["-y", "mcp-remote", "https://mcp.domain.com/mcp?account=org"]
}

This works for Claude's routing — it can select the right named server per task — but the OAuth consent screen still gives the user zero indication of which Google account to authenticate with. The ?account= query parameter is ignored by the server; it exists purely as a cache key trick for mcp-remote.

PROPOSED SOLUTION
Google's OAuth 2.0 authorize endpoint supports a [login_hint parameter](https://developers.google.com/identity/openid-connect/openid-connect#authenticationuriparameters) that pre-fills the email address on the consent screen and, if that account already has an active session, skips the account picker entirely.

A new environment variable — something like OAUTH_LOGIN_HINT — would allow operators to specify the expected Google account email per deployment. The server would append &login_hint=<email> when constructing the redirect URL to Google's authorize endpoint.

Example config:
yamlenvironment:
  - MCP_ENABLE_OAUTH21=true
  - WORKSPACE_MCP_STATELESS_MODE=true
  - WORKSPACE_EXTERNAL_URL=https://mcp.domain.com
  - OAUTH_LOGIN_HINT=personal@gmail.com

This requires no changes to the OAuth flow itself — it's a single parameter appended to the existing Google authorize redirect. It would be especially valuable for self-hosted multi-account deployments where the operator knows which Google account each server instance is intended for, and wants to guide the user to authenticate with the correct one.

REFERENCES
Google OAuth login_hint docs: https://developers.google.com/identity/openid-connect/openid-connect#authenticationuriparameters
Semi-Related issue #372 (multi-account stdio)
Semi-Related PR #652 (single-user credential selection fix)

usiegj00:
Closes: #693 and #764

## Summary
- Auto-open the OAuth authorization URL in the browser via webbrowser.open() instead of displaying an impossibly long URL that wraps and breaks in terminal environments (especially Claude Code CLI)
- Add login_hint parameter to the OAuth URL when user_google_email is provided, so Google pre-selects the correct account
- Simplify the auth message when the browser opens successfully; fall back to showing the URL if it cannot

## Motivation
The OAuth authorization URL includes dozens of scopes and gets extremely long. In terminal-based MCP clients like Claude Code, the URL wraps with spaces injected at line breaks, making it impossible to copy/paste. Auto-opening the browser solves this entirely.

The login_hint parameter is a standard Google OAuth feature that pre-selects the user account on the consent screen, reducing friction when users have multiple Google accounts.

## Test plan
- [ ] Verify webbrowser.open() launches browser on macOS/Linux/Windows
- [ ] Verify fallback message displays URL when browser cannot open (e.g., headless server)
- [ ] Verify login_hint pre-selects correct account when email is provided
- [ ] Verify auth flow still works end-to-end after changes

## Summary by CodeRabbit

* **New Features**
  * Conditional browser auto-open for local/stdio OAuth sessions; falls back to a manual URL when auto-open fails.
  * Uses provided email as a login hint and shows the authenticated email on the callback, then retries the original tool call.

* **Documentation**
  * Revised first-time authentication guide explaining transport-specific consent flows and updated walkthroughs.

* **Bug Fixes**
  * Clearer user instructions and improved logging when automatic browser opening is unavailable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/taylorwilsdon/google_workspace_mcp/pull/556?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


taylorwilsdon:
Made quite a few changes but it works very well (gosh its been a long time since I've used legacy mode)

PR #559, opened 2026-03-11
fix: escape HTML in OAuth callback responses to prevent XSS

joshjacobson:
## Summary

Three functions in `auth/oauth_responses.py` interpolate dynamic strings directly into HTML via f-strings without escaping:

- `create_error_response()` — `error_message` can contain the Google `error` query parameter from the OAuth callback URL
- `create_success_response()` — `user_display` comes from the authenticated email
- `create_server_error_response()` — `error_detail` comes from exception `str(e)`

An attacker could craft a malicious OAuth redirect with a `?error=<script>...</script>` parameter, achieving reflected XSS in the user's browser during the OAuth flow.

## Changes

- Import `html.escape`
- Apply `html_escape()` to all three dynamic values before HTML interpolation

## Test plan

- [ ] Verify OAuth success page still displays the user email correctly
- [ ] Verify OAuth error page still displays error messages correctly
- [ ] Verify that `<script>alert(1)</script>` in an error message is rendered as escaped text, not executed
- [ ] Verify special characters in email addresses (e.g., `+`, `&`) display correctly

Found during a security review while evaluating this project for use with Claude Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * OAuth HTML responses now escape dynamic content to prevent HTML/script injection and ensure safe display of error, server-error, and success messages.
  * Success view falls back to a friendly default when no user display name is available.

* **Tests**
  * Added tests validating escaped output, correct fallback display, and expected HTTP status codes for error, server-error, and success responses.


taylorwilsdon:
@copilot resolve the merge conflicts in this pull request

PR #578, opened 2026-03-15
feat(gmail): auto-populate In-Reply-To/References for reply drafts

### LINKED ISSUE #555: draft_gmail_message: auto-populate In-Reply-To/References when thread_id is provided
When creating a reply draft with `thread_id` but without `in_reply_to`/`references`, Gmail's web UI sometimes fails to render the draft inline in the thread view. The draft appears in the Drafts list but is invisible when opening the conversation — a "ghost draft" that can't be edited from the thread.

This happens because Gmail's thread view uses `In-Reply-To` and `References` headers to determine where to render a draft in the reply tree. With only `thread_id`, the draft is associated at the data layer but has no render position in the UI.

**Current behavior:** Caller must manually fetch the thread, extract the last message's `Message-ID` header, build the `References` chain, and pass both `in_reply_to` and `references` alongside `thread_id`.

**Proposed:** When `thread_id` is provided but `in_reply_to`/`references` are omitted, automatically:
1. Fetch the thread via `threads.get`
2. Get the last message's `Message-ID` header → set as `In-Reply-To`
3. Build `References` from the last message's existing `References` + its `Message-ID`

This is a one-API-call lookup that makes threaded drafts reliably render in Gmail's web UI without requiring caller-side header management.

**Context:** Issue #294 reported this as a bug and was closed as "can't reproduce" — the behavior is intermittent and more common in longer threads (13+ messages). Auto-populating the headers would eliminate the issue entirely.

Bortlesboat:
## Summary
- Fixes #555 — reply drafts created with `thread_id` but without explicit `in_reply_to`/`references` now auto-populate these headers
- Eliminates "ghost drafts" that appear in the Drafts list but not inline in the thread view
- Respects caller-provided values — only fills missing headers

## How it works
When `draft_gmail_message` is called with a `thread_id` but without `in_reply_to` or `references`:
1. Fetches the thread via `threads().get()` with `format="metadata"` (lightweight, only Message-ID headers)
2. Collects all Message-IDs in thread order
3. Sets `In-Reply-To` to the last message's Message-ID
4. Sets `References` to the full chain of Message-IDs (per RFC 2822)

Gracefully degrades — if the fetch fails or no Message-IDs are found, the draft is created without the headers (existing behavior).

## Test plan
- [ ] Create a reply draft with `thread_id` only — verify In-Reply-To/References are populated
- [ ] Create a reply draft with explicit `in_reply_to` — verify caller value is preserved
- [ ] Create a new draft (no `thread_id`) — verify no change in behavior
- [ ] Verify draft appears inline in Gmail web UI thread view

## Summary by CodeRabbit

* **New Features**
  * Auto-populates conversation threading headers (In-Reply-To/References) when drafting replies to existing Gmail threads, using thread message IDs and any existing headers to improve reply context.
* **Bug Fixes**
  * Gracefully handles failures when fetching thread data (logs error and preserves draft behavior), preventing draft disruption.
* **Tests**
  * Added comprehensive async tests for draft creation covering attachments, signatures, threading behavior, and error cases.

### DIFF
--- a/tests/gmail/test_draft_gmail_message.py
+++ b/tests/gmail/test_draft_gmail_message.py
@@ -143,6 +143,15 @@ async def test_draft_gmail_message_autofills_reply_headers_from_thread():
         include_signature=False,
     )
 
+    # Verify threads().get() was called with correct parameters
+    thread_get_kwargs = (
+        mock_service.users.return_value.threads.return_value.get.call_args.kwargs
+    )
+    assert thread_get_kwargs["userId"] == "me"
+    assert thread_get_kwargs["id"] == "thread123"
+    assert thread_get_kwargs["format"] == "metadata"
+    assert "Message-ID" in thread_get_kwargs["metadataHeaders"]
+
     assert "Draft created! Draft ID: draft_reply" in result
 
     create_kwargs = (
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
You're on a tear, thanks!

taylorwilsdon:
Cleaned up a bit, will ship with next release

PR #580, opened 2026-03-15
feat(sheets): expose cell notes in read_sheet_values

### LINKED ISSUE #517: Expose cell notes in read_sheet_values
The issue is read_sheet_values only returns flat cell values. There's no way to retrieve cell notes through any existing operation.
After a bit of (AI) research, I think the Sheets API exposes notes as a top-level note field on CellData, available via spreadsheets.get with includeGridData=True.

Having access to the notes would be super useful. 

P.S. I can give it a try, though being on the JVM, I've never coded in Python :)

Bortlesboat:
## Summary
- Fixes #517 — adds `include_notes` parameter to `read_sheet_values` to expose Google Sheets cell notes
- Default `False` preserves existing behavior with zero extra API calls
- Follows the same pattern as existing `include_hyperlinks` parameter

## Implementation
- `include_notes: bool = False` parameter on `read_sheet_values`
- When `True`, fetches notes via `spreadsheets.get` with `includeGridData=True` and `fields` scoped to only `note` (minimal payload)
- Three new helpers in `sheets_helpers.py`:
  - `_extract_cell_notes_from_grid()` — extracts `[{"cell": "A1", "note": "..."}]` from grid data
  - `_fetch_sheet_notes()` — API call with tight range bounds and 5000-cell cap
  - `_format_sheet_notes_section()` — formats notes into readable output (max 25 entries)

## Test plan
- [ ] Call `read_sheet_values` without `include_notes` — verify unchanged behavior, no extra API call
- [ ] Call with `include_notes=True` on a sheet with notes — verify notes appear in output
- [ ] Call with `include_notes=True` on a sheet without notes — verify graceful empty response
- [ ] Verify large ranges respect the 5000-cell guard rail

## Summary by CodeRabbit

* **New Features**
  * Retrieve and display cell notes from Google Sheets when reading sheet values (optional).
  * Notes appear in a human-readable section with truncation and an overflow summary for large sets.
  * Metadata retrieval (hyperlinks/notes) consolidated for more efficient single-request fetches; a cell-cap prevents overly large requests.
  * Reading sheet values gains an optional parameter to include notes in output.

### DIFF
--- a/gsheets/sheets_helpers.py
+++ b/gsheets/sheets_helpers.py
@@ -991,9 +991,7 @@ async def _fetch_grid_metadata(
         )
         return "", ""
 
-    cell_count = _a1_range_cell_count(tight_range) or sum(
-        len(row) for row in values
-    )
+    cell_count = _a1_range_cell_count(tight_range) or sum(len(row) for row in values)
     if cell_count > MAX_GRID_METADATA_CELLS:
         logger.info(
             "[read_sheet_values] Skipping grid metadata fetch for large range "


taylorwilsdon:
Appreciate this!

PR #581, opened 2026-03-15
feat(gmail): support quoted reply content and signature positioning

### LINKED ISSUE #547: draft_gmail_message: support quoted reply content and signature positioning
## Problem

When creating reply drafts with `draft_gmail_message` using `thread_id` + `in_reply_to` + `references`, the tool handles threading metadata correctly but the body is treated as a raw passthrough. The caller must manually construct the full reply body including quoted content from the original message.

The `include_signature` parameter (added in #530) appends the signature to the end of whatever body is passed. For new messages this works fine, but for replies it creates a problem: if the caller includes quoted content in the body, the signature ends up *after* the quoted text instead of between the new text and the quoted text — which is the universal email convention.

Currently, building a proper reply requires:
1. Fetching the thread with `get_gmail_thread_content`
2. Formatting the quoted content yourself
3. Manually constructing: `[new reply text] + [signature] + [quoted content]`
4. Setting `include_signature: false` to prevent a duplicate signature at the end

## Proposed Solution

Add a `quote_replied_message` parameter (default `false`) to `draft_gmail_message`:

```python
quote_replied_message: Annotated[
    bool,
    Field(
        description="When replying (thread_id provided), automatically append the replied-to message as quoted content below the signature. Defaults to false.",
    ),
] = False,
```

When `quote_replied_message=True` and a `thread_id` + `in_reply_to` are provided:
1. Fetch the message identified by `in_reply_to`
2. Format it as quoted content (matching Gmail's native reply format)
3. Compose the body as: `[caller's body]` + `[signature if include_signature=True]` + `[quoted message]`

This follows the same pattern as `include_signature` — a boolean the caller controls, defaulting to backward-compatible behavior.

## Why This Belongs in the Server

The server already has all the pieces:
- It fetches message content (`get_gmail_message_content`)
- It fetches and appends signatures (`_get_send_as_signature_html`, `_append_signature_to_body`)
- It handles threading metadata

Composing these together for replies is a natural extension. Without it, every MCP consumer that creates reply drafts must independently implement quoted content formatting.

Bortlesboat:
## Summary
- Fixes #547 — adds `quote_original` parameter to `draft_gmail_message` for proper reply formatting
- When enabled, automatically fetches the original message and constructs a Gmail-style quoted reply
- Signature is positioned correctly between the reply body and the quoted original

## Implementation
- `quote_original: bool = False` parameter (backward compatible)
- `_fetch_original_for_quote()` — fetches thread, finds target message by Message-ID or falls back to last message, extracts sender/date/body
- `_build_quoted_reply_body()` — assembles reply in correct order:
  1. Reply body
  2. Signature (if enabled)
  3. Attribution line: "On {date}, {sender} wrote:"
  4. Quoted original (`<blockquote>` for HTML, `>` prefix for plain text)
- Graceful fallback — if original can't be fetched, falls back to simple body + signature

## Test plan
- [ ] Create reply draft with `quote_original=True` — verify quoted content appears with attribution
- [ ] Verify HTML replies use `<blockquote>` styling
- [ ] Verify plain text replies use `>` prefix
- [ ] Verify signature appears between reply body and quoted content
- [ ] Verify `quote_original=False` (default) — unchanged behavior
- [ ] Verify without `thread_id` — `quote_original` is ignored gracefully

## Summary by CodeRabbit

* **New Features**
  * Draft replies can now include the original message quoted inline; users can toggle quoting when composing within a thread.
  * Quoting supports both plain-text and HTML formats and preserves user signatures when included.
  * When enabled, drafts automatically assemble the reply with the quoted original for clearer conversation context.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -1528,6 +1528,9 @@ async def draft_gmail_message(
               - 'mime_type' (optional): MIME type (defaults to 'application/octet-stream')
         include_signature (bool): Whether to append Gmail signature HTML from send-as settings.
             If unavailable (e.g., missing gmail.settings.basic scope), the draft is still created without signature.
+        quote_original (bool): Whether to include the original message as a quoted reply.
+            Requires thread_id to be provided. When enabled, fetches the original message
+            and appends it below the signature. Defaults to False.
 
     Returns:
         str: Confirmation message with the created draft's ID.


taylorwilsdon:
Hot damn, you've been on a a roll. Appreciate these!

PR #582, opened 2026-03-16
Fix: wire WORKSPACE_MCP_STATELESS_MODE to FastMCP's stateless_http parameter

vdemay:
## Problem

`WORKSPACE_MCP_STATELESS_MODE=true` is already read, validated (requires `MCP_ENABLE_OAUTH21=true`), and used to configure the OAuth layer — but was never passed to `server.run()`. FastMCP therefore always runs in **stateful mode**, keeping session state in memory.

On pod/process restart (e.g. Kubernetes rolling update, crash loop) all in-memory sessions are lost. Clients that reuse their `mcp-session-id` receive:

```json
{"jsonrpc":"2.0","id":"server-error","error":{"code":-32600,"message":"Session not found"}}
```

## Fix

Pass `stateless_http=is_stateless_mode()` to `server.run()` so FastMCP drops session tracking when stateless mode is configured — matching what the env var already advertises.

`is_stateless_mode()` is already imported at the top of `main.py`, so this is a one-line behavioural fix.

## Before / After

```python
# Before
server.run(transport="streamable-http", host=host, port=port)

# After
server.run(
    transport="streamable-http",
    host=host,
    port=port,
    stateless_http=is_stateless_mode(),
)
```

## Test plan

- [ ] Start server with `WORKSPACE_MCP_STATELESS_MODE=false` (default) — stateful behaviour unchanged
- [ ] Start server with `WORKSPACE_MCP_STATELESS_MODE=true` + `MCP_ENABLE_OAUTH21=true` — restart server, confirm subsequent MCP requests succeed without "Session not found"

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed server startup configuration for streamable-http transport to ensure stateless mode is properly enabled during initialization.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -36,7 +36,12 @@ logger = logging.getLogger(__name__)
 GMAIL_BATCH_SIZE = 25
 GMAIL_REQUEST_DELAY = 0.1
 HTML_BODY_TRUNCATE_LIMIT = 20000
-GMAIL_METADATA_HEADERS = ["Subject", "From", "To", "Cc", "Message-ID", "Date"]
+
+GMAIL_METADATA_HEADERS = [
+    "Subject", "From", "To", "Cc",
+    "Message-ID", "In-Reply-To", "References",
+    "Date"
+]
 LOW_VALUE_TEXT_PLACEHOLDERS = (
     "your client does not support html",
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
Ah, nice. This was not a thing in the past. Love to see it.

PR #598, opened 2026-03-21
Fix FastMCP 3.1.1 OAuth proxy bugs blocking Claude Code auth

drewgillson:
## Summary

- Adds runtime workarounds for two FastMCP 3.1.1 bugs that completely block OAuth 2.1 authentication for Claude Code (and any MCP client using dynamic localhost ports)
- Monkey-patches redirect URI validation to be RFC 8252 §7.3 compliant (loopback addresses accept any port)
- Pre-creates FileTreeStore directories for URL-based CIMD client IDs that otherwise cause 500 errors

## Context

FastMCP 3.1.1 introduced two regressions in the OAuth proxy:

1. **Redirect URI port matching** — Claude Code's CIMD document lists `http://localhost/callback` but the client uses ephemeral ports like `http://localhost:51687/callback`. FastMCP does strict port matching and rejects it with a 400. Per RFC 8252 §7.3, localhost redirects should match regardless of port.

2. **FileTreeStore path handling** — The `FileTreeStore` uses the client_id URL (`https://claude.ai/oauth/claude-code-client-metadata`) directly as a filesystem key, creating a path like `.fastmcp/oauth-proxy/mcp-oauth-proxy-clients/https:/claude.ai/oauth/`. It doesn't create intermediate directories, so writes fail with `FileNotFoundError` (500 ISE).

Both workarounds are applied at import time in `core/server.py` and are designed to be safely removed once the upstream issue is resolved.

Upstream issue: https://github.com/PrefectHQ/fastmcp/issues/3574

## Test plan

- [ ] Start server with `MCP_ENABLE_OAUTH21=true` and `streamable-http` transport
- [ ] From a new Claude Code project directory, add the server: `claude mcp add --transport http workspace-mcp http://localhost:8005/mcp`
- [ ] Authenticate via `/mcp` → Authenticate — should complete the Google OAuth flow without 400 or 500 errors
- [ ] Verify existing authenticated sessions continue to work

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth redirect URI validation to better support local development and testing environments, including flexible port handling for loopback connections.
  * Enhanced OAuth proxy filesystem storage initialization to prevent common file system-related failures and improve reliability during authentication workflows.

### DIFF
--- a/core/server.py
+++ b/core/server.py
@@ -3,6 +3,8 @@ import logging
 import os
 from typing import List, Optional
 from importlib import metadata
+from pathlib import Path
+from urllib.parse import urlparse
 
 from fastapi.responses import HTMLResponse, JSONResponse, FileResponse
 from starlette.applications import Starlette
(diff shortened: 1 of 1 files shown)


taylorwilsdon:
Hey there, I think there may be a misunderstanding at play here? Claude Code is one of my primary clients and we have thousands, possibly tens of thousands of users running Workspace MCP through it, this is the first I've heard of any issue (and cannot reproduce with the current version of CC & Workspace MCP as of this comment). I'm wondering if you're missing the `"OAUTHLIB_INSECURE_TRANSPORT": "1"` flag for local, non HTTPS usage?

drewgillson:
Hey Taylor, thanks for taking a look! You're right that most users wouldn't hit this — it's specific to the `streamable-http` + `MCP_ENABLE_OAUTH21=true` + `FileTreeStore` (disk) storage backend combination, which most stdio users wouldn't encounter.

I did test with `OAUTHLIB_INSECURE_TRANSPORT=1` added to `.env` and can confirm both issues still reproduce. That flag controls whether `oauthlib` allows `http://` URLs for the upstream Google OAuth leg, but these errors occur earlier in FastMCP's own OAuth proxy layer before `oauthlib` is involved.

Here's a minimal repro with `OAUTHLIB_INSECURE_TRANSPORT=1` set, on a clean install of the current `main` + FastMCP 3.1.1:

**Bug 1 — FileTreeStore can't handle URL-based client IDs (500):**
```
curl http://localhost:8005/authorize?response_type=code&client_id=https%3A%2F%2Fclaude.ai%2Foauth%2Fclaude-code-client-metadata&...
→ Internal Server Error

# Error: FileNotFoundError at .fastmcp/oauth-proxy/mcp-oauth-proxy-clients/https:/claude.ai/oauth/tmp....tmp
```

**Bug 2 — Redirect URI port matching (400), after manually creating the directory:**
```
mkdir -p ~/.fastmcp/oauth-proxy/mcp-oauth-proxy-clients/https:/claude.ai/oauth
curl "http://localhost:8005/authorize?...&redirect_uri=http%3A%2F%2Flocalhost%3A55555%2Fcallback..."
→ {"error":"invalid_request","error_description":"Redirect URI 'http://localhost:55555/callback' does not match CIMD redirect_uris."}
```

Claude Code's CIMD document (`https://claude.ai/oauth/claude-code-client-metadata`) lists `http://localhost/callback` (port 80), but Claude Code uses ephemeral high ports. FastMCP does strict port matching against CIMD, which violates RFC 8252 §7.3 for loopback redirects.

Both bugs are in FastMCP itself (tracked at https://github.com/PrefectHQ/fastmcp/issues/3574), so this PR is just a runtime workaround. Totally understand if you'd prefer to wait for the upstream fix rather than carry the monkey-patch — just wanted to flag for visibility since it blocks OAuth 2.1 auth for any new Claude Code session.

drewgillson:
Dependency versions for reference:

- **FastMCP**: 3.1.1
- **MCP SDK**: 1.26.0
- **Claude Code**: 2.1.81
- **Python**: 3.13.12
- **workspace-mcp**: 1.14.3 (current `main` as of 5a2074e)

lkchao-majority:
Hi i confirm to have same error, it was working fine previously with older version.

taylorwilsdon:
Hm, and you're using the FileTreeStore integration specifically @lkchao-majority? 

Can you try running with normal in memory storage, or a backend like Valkey?

lkchao-majority:
@taylorwilsdon Hi that's correct when using memory it is working fine.
Same if i created the folder `/home/app/.fastmcp/oauth-proxy/mcp-oauth-proxy-clients/https:/claude.ai/oauth/` in advance.

Also thanks for sharing that project, it has been really useful for some of my colleagues.

taylorwilsdon:
Thanks for confirming @lkchao-majority - that's very useful metadata. I'm hesitant to merge this huge monkey patch into the core server but I don't want folks broken. Let's see if FastMCP acks the issue.

taylorwilsdon:
Looks like upstream fix was already merged - https://github.com/PrefectHQ/fastmcp/pull/3589 will bump version here as soon as it ships

pnancarrow:
> Looks like upstream fix was already merged - [PrefectHQ/fastmcp#3589](https://github.com/PrefectHQ/fastmcp/pull/3589) will bump version here as soon as it ships

@taylorwilsdon it looks like that was released. would love to see this resolved!

taylorwilsdon:
It's live in the codebase @pnancarrow, will release a new build to pypi this AM.

PR #606, opened 2026-03-24
fix: resolve OAuth port conflict on re-auth in stdio mode

### LINKED ISSUE #546: Bug: Multiple stdio processes cause OAuth callback server port conflict and state mismatch
**Describe the bug**
When Claude Desktop runs workspace-mcp in stdio mode, it spawns multiple processes 
within seconds of each other (visible via different PIDs in the debug log). This 
triggers two cascading bugs:

*Bug 1 – Port conflict:* The second process tries to bind the OAuth callback port 
on startup but the first process already holds it. The second process sets 
is_running = False, and all subsequent tool calls that require authentication fail:

  "Cannot initiate OAuth flow - callback server unavailable 
   (Port 8000 is already in use on localhost. Cannot start minimal OAuth server.)"

*Bug 2 – OAuth state mismatch:* Even if Bug 1 is resolved, the OAuth state 
(_oauth_states dict in OAuth21SessionStore) lives only in the memory of the process 
that generated the auth URL. When Google redirects the callback to localhost:<port>, 
it hits the other process's server, which has no knowledge of that state:

  "Invalid or expired OAuth state parameter"

**Startup Logs**
*Process A (PID 23084) — starts successfully:*
2026-03-06 09:06:59,960 - core.server - INFO - 23084 - MainThread - Transport: stdio
2026-03-06 09:07:00,043 - auth.oauth_callback_server - INFO - 23084 - Creating minimal OAuth server instance for http://localhost:7777
2026-03-06 09:07:00,562 - auth.oauth_callback_server - INFO - 23084 - Minimal OAuth server started on localhost:7777
2026-03-06 09:07:00,562 - auth.oauth_callback_server - INFO - 23084 - Minimal OAuth server successfully started on http://localhost:7777

*Process B (PID 22820) — starts 15 seconds later, port already taken:*
2026-03-06 09:07:15,746 - core.server - INFO - 22820 - MainThread - Transport: stdio
2026-03-06 09:07:15,747 - auth.oauth_callback_server - INFO - 22820 - Creating minimal OAuth server instance for http://localhost:7777
2026-03-06 09:07:15,843 - auth.oauth_callback_server - ERROR - 22820 - Port 7777 is already in use on localhost. Cannot start minimal OAuth server.

*Tool call on Process B then fails:*
2026-03-06 09:07:24,309 - auth.service_decorator - ERROR - 22820 - [list_calendars] 
  GoogleAuthenticationError: Cannot initiate OAuth flow - callback server unavailable 
  (Port 7777 is already in use on localhost. Cannot start minimal OAuth server.)

**To Reproduce**
1. Configure workspace-mcp as an MCP server in Claude Desktop (stdio transport)
2. Start Claude Desktop — two workspace-mcp processes start within ~15 seconds
3. Open a conversation and trigger any tool call requiring Google authentication 
   (e.g. list_calendars)
4. See error: "Cannot initiate OAuth flow - callback server unavailable 
   (Port XXXX is already in use)"
5. If port conflict is patched: click the auth URL, complete Google OAuth in the 
   browser, observe "Invalid or expired OAuth state parameter" on the callback page

**Expected behavior**
Authentication should succeed regardless of how many workspace-mcp processes Claude 
Desktop spawns. The OAuth callback server and state should be shared or tolerant of 
multiple concurrent instances.

**Screenshots**
Reproducible via debug log (enable via LOG_LEVEL=DEBUG in config).

- OS: Windows 11
- Container: n/a
- Version: v1.14.2

**Additional context**
Both bugs only affect stdio transport mode with multiple process instances. 
Streamable-HTTP mode is not affected.

Root causes:
- Bug 1: MinimalOAuthServer.start() in auth/oauth_callback_server.py treats any 
  OSError on socket.bind() as fatal, even when the port is held by another 
  workspace-mcp instance that is actively serving.
- Bug 2: OAuth21SessionStore._oauth_states is an in-memory dict with no 
  cross-process sharing.

Workarounds applied locally:
- Bug 1: After bind failure, test connect_ex() to the port — if it responds, 
  treat as success (reuse existing server).
- Bug 2: Persist OAuth states to a shared JSON file in the credentials directory 
  (~/.google_workspace_mcp/credentials/.oauth_states.json) so any process can 
  validate the callback.

Note: I'm not a developer. I discovered and diagnosed these bugs with the help of 
Claude Cowork (Anthropic's AI desktop assistant), which also identified the root 
causes and implemented the local workarounds. I'm happy to share the patched files 
if that helps.

Alexheath21:
## Summary

- Fixes the OAuth port conflict bug where `MinimalOAuthServer` fails with "port already in use" when re-authenticating against its own process in stdio mode
- `start()` now detects a live server thread even when `is_running` is stale (set to `False` by an exception handler while the thread still holds the port)
- The socket bind fallback probes `is_actually_running()` before erroring, recovering when the port is held by our own server
- `stop()` no longer short-circuits on `is_running=False`, ensuring proper cleanup of the uvicorn server and thread

Fixes #546

## Root Cause

In stdio mode, `MinimalOAuthServer` starts a uvicorn server on a daemon thread. When the server thread encounters an error, it sets `is_running = False` but the thread and port binding persist. On the next re-auth attempt:

1. `start()` sees `is_running=False` and skips the "already running" check
2. The socket bind check fails because the port is held by the previous server thread in our own process
3. `start()` returns `"Port 8000 is already in use"` — blocking all subsequent MCP calls until restart

## Fix Details

Three targeted changes to `auth/oauth_callback_server.py`:

1. **`start()` — thread-alive check before `is_running` flag**: Checks `server_thread.is_alive()` first, which catches the stale-flag scenario. If the thread is alive and the port responds, reuses the server.

2. **`start()` — bind failure recovery**: When the socket bind fails with `EADDRINUSE`, probes `is_actually_running()` before giving up. If the port is responding (our own server), marks it as running and succeeds.

3. **`stop()` — unconditional cleanup**: Removed the `if not self.is_running: return` guard so `stop()` always signals `should_exit` and joins the thread, preventing zombie server threads.

## Test Plan

- [x] Added `test_start_reuses_server_when_thread_alive_and_port_responding` — reproduces the exact re-auth bug scenario
- [x] Added `test_start_recovers_when_port_in_use_by_own_server` — covers the bind-failure recovery path
- [x] Added `test_stop_cleans_up_even_when_is_running_false` — verifies stop() works with stale flag
- [x] All 4 existing tests continue to pass
- [x] Full test suite: 309 passed (2 pre-existing failures unrelated to this change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth server startup reliability by better detecting and reusing running instances when the internal state flag is stale.
  * Enhanced port-binding failure handling to recover when a server instance already occupies the port.
  * Strengthened server shutdown to complete cleanly even with inconsistent internal state.

* **Tests**
  * Added comprehensive test coverage for OAuth server lifecycle edge cases and error recovery scenarios.

### DIFF
--- a/auth/oauth_callback_server.py
+++ b/auth/oauth_callback_server.py
@@ -197,7 +197,24 @@ class MinimalOAuthServer:
         Returns:
             Tuple of (success: bool, error_message: str)
         """
-        if self.is_running:
+        # Check if this instance already has a running server — whether is_running
+        # is set or not.  The flag can be stale (e.g. set to False by an exception
+        # handler while the thread is still alive and holding the port).
+        if self.server_thread and self.server_thread.is_alive():
+            if self.is_actually_running():
+                logger.info(
+                    "Minimal OAuth server thread is still alive and port is responding, reusing"
+                )
+                self.is_running = True
+                return True, ""
+            else:
+                logger.warning(
+                    "Minimal OAuth server thread is alive but port is not responding. Restarting."
+                )
+                self.stop()
+                self.server = None
+                self.server_thread = None
+        elif self.is_running:
             if self.is_actually_running():
                 logger.info("Minimal OAuth server is already running")
                 return True, ""
@@ -221,6 +238,14 @@ class MinimalOAuthServer:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
                 s.bind((hostname, self.port))
         except OSError:
+            # Before failing, check if the port is held by a previous server
+            # instance in our own process that we can still use for callbacks.
+            if self.is_actually_running():
+                logger.info(
+                    f"Port {self.port} is in use by our own OAuth server, reusing"
+                )
+                self.is_running = True
+                return True, ""
             error_msg = f"Port {self.port} is already in use on {hostname}. Cannot start minimal OAuth server."
             logger.error(error_msg)
             return False, error_msg
@@ -268,10 +293,12 @@ class MinimalOAuthServer:
         return False, error_msg
 
     def stop(self):
-        """Stop the minimal OAuth server."""
-        if not self.is_running:
-            return
+        """Stop the minimal OAuth server.
 
+        Attempts to shut down the uvicorn server and join its thread regardless
+        of the ``is_running`` flag — the flag can be stale when an exception in
+        the server thread resets it while the thread is still alive.
+        """
         try:
             if self.server:
                 if hasattr(self.server, "should_exit"):
@@ -285,6 +312,7 @@ class MinimalOAuthServer:
 
         except Exception as e:
             logger.error(f"Error stopping minimal OAuth server: {e}", exc_info=True)
+            self.is_running = False
 
 
 # Global instance for stdio mode
--- a/tests/auth/test_oauth_callback_server.py
+++ b/tests/auth/test_oauth_callback_server.py
@@ -131,3 +131,100 @@ def test_ensure_oauth_callback_skips_start_when_other_instance_owns_port(monkeyp
     assert error == ""
     assert len(_PortInUseServer.instances) == 1
     assert _PortInUseServer.instances[0].start_calls == 0
+
+
+class _AliveThread:
+    def is_alive(self):
+        return True
+
+
+def test_start_reuses_server_when_thread_alive_and_port_responding():
+    """Reproduce the re-auth bug: is_running is False but the server thread is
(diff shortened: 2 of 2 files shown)


taylorwilsdon:
There was a PR that the author closed looking to do something similar in https://github.com/taylorwilsdon/google_workspace_mcp/pull/590

Will take a look. Some of the coderabbit feedback looks actionable before merge!

PR #639, opened 2026-03-31
feat(calendar): add rsvp_event tool for non-organizer RSVP responses

thagale:
## What
Adds a dedicated `rsvp_event` tool for accepting, declining, or marking tentative on calendar event invitations.

## Why
The existing `manage_event` tool uses `events.update()` (PUT), which requires organizer permission and overwrites the full event body. It cannot be used to update an attendee's own `responseStatus`. This adds `events.patch()` with only the attendees array — the correct approach per the Google Calendar API docs for non-organizer RSVP.

## Changes
- `gcalendar/calendar_tools.py`: adds `_rsvp_event_impl` and `rsvp_event` tool
- `core/tool_tiers.yaml`: registers `rsvp_event` under `calendar.core`

## Behavior
- Validates response is one of `accepted`, `declined`, `tentative`, `needsAction`
- Guards against attempting to RSVP as the organizer
- Raises clearly if the authenticated user isn't in the attendee list
- Supports optional `comment` and `send_updates` control
- Registered at `calendar.core` tier (same as `manage_event`)

## Release Notes

* **New Features**
  * Enhanced calendar event management with new comprehensive RSVP capabilities. Users can now respond directly to calendar invitations with their attendance status (accept, decline, or tentative), optionally include personal comments with their responses, and customize notification settings to control whether other event attendees receive updates about their RSVP decision.

### DIFF
--- a/core/tool_tiers.yaml
+++ b/core/tool_tiers.yaml
@@ -44,7 +44,6 @@ calendar:
     - list_calendars
     - get_events
     - manage_event
-    - rsvp_event
   extended:
     - query_freebusy
     - manage_out_of_office
--- a/gcalendar/calendar_tools.py
+++ b/gcalendar/calendar_tools.py
@@ -1177,6 +1177,12 @@ async def _rsvp_event_impl(
             f"Invalid response '{response}'. Must be one of: {sorted(valid_responses)}"
         )
 
+    valid_send_updates = {"all", "externalOnly", "none"}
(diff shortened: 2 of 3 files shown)


taylorwilsdon:
@coderabbitai review

taylorwilsdon:
Refactored this into existing tools, way too specific to expose a new tool. We're already way over the number of tools a server should be exposing as-is.

PR #669, opened 2026-04-08
fix(auth): propagate valid scopes to CIMD manager for OAuth 2.1

### LINKED ISSUE #668: CIMD clients (Claude Code) get invalid_scope error with limited TOOLS config
## Bug Description

When running the server with OAuth 2.1 enabled (`MCP_ENABLE_OAUTH21=true`) and a limited tool set (e.g. `TOOLS="drive"`), CIMD clients like Claude Code fail to authenticate with:

```
Authentication Error
invalid_scope: Client was not registered with scope https://www.googleapis.com/auth/drive
```

The error originates from the MCP server itself (not from Google), as the server redirects back to the client callback with the error before ever reaching Google's authorization server.

## Root Cause

Claude Code's [CIMD document](https://claude.ai/oauth/claude-code-client-metadata) does not include a `scope` field. When FastMCP's `CIMDClientManager` resolves the client, it falls back to `self.default_scope` (`cimd.py:754`):

```python
scope=cimd_doc.scope or self.default_scope
```

`default_scope` is initialized from `OAuthProxy.required_scopes` (`proxy.py:363`):

```python
self._default_scope_str: str = " ".join(self.required_scopes or [])
```

Since `required_scopes` is set to `BASE_SCOPES` (openid, email, profile) in `server.py`, the CIMD client is registered with only base scopes, not the tool-specific scopes (e.g. drive).

The existing fix in commit `8b21589` (`client_registration_options.default_scopes = provider_valid_scopes`) correctly handles the MCP SDK's standard DCR registration handler, but the CIMD flow in FastMCP's OAuthProxy bypasses it entirely, using its own `default_scope` attribute instead.

## Steps to Reproduce

1. Set `MCP_ENABLE_OAUTH21=true` and `TOOLS="drive"` in `.env`
2. Run the server via Docker with `--transport streamable-http`
3. Connect from Claude Code
4. The OAuth flow starts but immediately returns `invalid_scope`

## Environment

- google_workspace_mcp v1.18.0
- FastMCP 3.2.0
- Claude Code as MCP client

nicolasalvemini-hue:
## Summary

- Fixes #668 — CIMD clients (e.g. Claude Code) were registered with only `BASE_SCOPES` (openid, email, profile), causing `invalid_scope` errors when requesting tool-specific scopes
- Propagates `provider_valid_scopes` to the CIMD manager's `default_scope` so that CIMD clients without a `scope` field in their metadata document are registered with all enabled tool scopes

## Root Cause

The existing fix (`client_registration_options.default_scopes`) from commit `8b21589` applies to the MCP SDK's standard DCR registration handler. However, the CIMD flow in FastMCP's `CIMDClientManager` bypasses that handler entirely and uses its own `default_scope` attribute, which was only initialized with `required_scopes` (BASE_SCOPES).

## Test plan

- [ ] Run server with `MCP_ENABLE_OAUTH21=true` and `TOOLS="drive"`
- [ ] Connect from Claude Code — OAuth flow should complete successfully
- [ ] Verify tool-specific scopes (e.g. `drive`) are included in the authorization request to Google
- [ ] Verify existing DCR registration flow still works

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OAuth provider scope configuration to properly apply enabled scopes during provider initialization, ensuring correct scope settings are propagated throughout the authentication system.

### DIFF
--- a/core/server.py
+++ b/core/server.py
@@ -490,6 +490,11 @@ def configure_server_for_http():
                     provider.client_registration_options.default_scopes = (
                         provider_valid_scopes
                     )
+                # Propagate valid scopes to the CIMD manager so that CIMD
+                # clients (e.g. Claude Code) are registered with all enabled
+                # tool scopes, not just BASE_SCOPES.
+                if hasattr(provider, '_cimd_manager') and provider._cimd_manager is not None:
+                    provider._cimd_manager.default_scope = " ".join(provider_valid_scopes)
                 # Enable protocol-level auth
                 server.auth = provider
                 logger.info(


taylorwilsdon:
Hm - I'm not sure I follow. Can you explain the use case? Afaik FastMCP hasn't implemented CIMD support yet.

PR #672, opened 2026-04-10
feat: inject quotaUser for per-user quota attribution

alex-modrongroup:
## Summary

Pass `quotaUser=<user-email>` on every Google API call so Google attributes quota to the per-user bucket instead of the project's global bucket.

Without this, multi-tenant deployments through one OAuth client share a single global quota and hit 429s earlier than necessary. This is a [documented Google recommendation](https://cloud.google.com/apis/docs/capping-api-usage#quotauser) for multi-tenant integrators.

## Changes

- Added `_inject_quota_user()` helper in `auth/service_decorator.py` that wraps the HTTP transport to append `quotaUser` to every request URL
- Applied at all 4 `build()` call sites: service account, OAuth 2.1 (2 paths), and OAuth 2.0
- Zero behavioral change for single-user deployments (quotaUser is ignored when there's only one user)

## Attribution

Portions derived from [Mail-0/Zero](https://github.com/Mail-0/Zero) — Copyright (c) 2025 Zero Email, MIT License  
Original TypeScript source: `apps/server/src/lib/driver/google.ts:579-581` (`getQuotaUser()`)

## Testing

- No breaking changes to existing API surface
- `quotaUser` is a standard Google API parameter — no additional scopes required
- Verified: parameter is URL-encoded to handle special characters in email addresses

## Release Notes

* **Chores**
  * Enhanced API request handling to properly associate requests with individual users for improved quota tracking and resource management.

### DIFF
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -1195,7 +1195,10 @@ async def get_authenticated_google_service(
         raise GoogleAuthenticationError(auth_response)
 
     try:
+        from auth.service_decorator import _inject_quota_user
+
         service = build(service_name, version, credentials=credentials)
+        service = _inject_quota_user(service, user_google_email)
         log_user_email = user_google_email
 
         # Try to get email from credentials if needed for validation
--- a/auth/service_decorator.py
+++ b/auth/service_decorator.py
@@ -2,6 +2,7 @@ import inspect
 import json
 import logging
 import os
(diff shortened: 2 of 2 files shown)


alex-modrongroup:
Closing — filed prematurely without following contribution guidelines. Will refile after proper review.

PR #699, opened 2026-04-17
Add Gmail draft lifecycle management tools

future3OOO:
## Summary

This PR adds Gmail draft update and delete support, completing the draft lifecycle exposed by the Gmail integration.

Existing draft support allows creating drafts, but not modifying or deleting them. That leaves no supported way to revise an existing draft in place or remove a stale generated draft.

This PR adds:
- `update_gmail_draft`
- `delete_gmail_draft`

Draft creation and draft update use the same message-building path so aliases, signatures, attachments, threading headers, quoted replies, and reply-subject fallback stay consistent.

## Behavior

`update_gmail_draft` is replacement-based: it rebuilds the draft MIME message and preserves omitted existing draft fields where applicable.

- omit supported fields to preserve existing values
- pass an empty string to clear supported string fields
- pass an empty list to clear attachments
- whitespace-only optional update strings are treated as clear
- update preserves omitted inline related parts, attached `message/rfc822` parts, and preserved unnamed attachment parts
- URL attachments are resolved during update
- malformed raw MIME fails closed with `UserInputError`
- caller-supplied unnamed attachments remain rejected
- CID-bearing preserved parts are only treated as inline when disposition is `inline`

For draft creation, blank `from_email`, `in_reply_to`, and `references` inputs continue to behave like omission where fallback behavior is expected.

## Scope / diff

- 8 files changed, +1070 / -38
- runtime: `gmail/gmail_tools.py` (+442 / -36)
- tests: 3 files, +590 lines
- docs + tier wiring: 4 files, +40 / -2

## Tooling and exposure

- `draft_gmail_message`, `update_gmail_draft`, and `delete_gmail_draft` use `GMAIL_COMPOSE_SCOPE`
- the draft lifecycle tools are exposed in the Gmail `extended` tool tier
- this PR does not change send endpoint exposure

## Pros

- Completes the missing draft lifecycle surface.
- Allows partial draft updates without dropping preserved sender, threading, or attachment state.
- Keeps create and update behavior aligned through one message-building path.

## Cons / tradeoffs

- Gmail draft updates remain replacement-based, so correctness depends on rebuilding MIME accurately from preserved state.
- Shared message-building logic affects both draft creation and draft update.
- `delete_gmail_draft` is destructive and depends on the caller passing the intended `draft_id`.

## Validation

Verified on this branch with:

```bash
uv sync --group dev
uv run ruff check .
uv run ruff format --check .
uv run pytest
```

Result: all passed (`787 passed`).

Also verified with a live MCP run: create -> find -> update -> confirm update -> delete -> confirm cleanup against the configured Gmail account.

## GitNexus usage

GitNexus was run directly against this branch worktree (`gitnexus analyze .`) and then used from the Hermes MCP integration to inspect the Gmail draft flow and shared helper surfaces involved in draft creation, update, and attachment handling before finalizing the branch.

## Summary by CodeRabbit

* **New Features**
  * Added update_gmail_draft (Extended tier) to replace an existing draft by ID
  * Added delete_gmail_draft (Extended tier) to remove a draft by ID

* **Documentation**
  * Updated Gmail tool references and examples to include new draft management capabilities

* **Bug Fixes / Reliability**
  * Improved attachment URL resolution and safer error reporting for failed URL attachments

* **Tests**
  * Added tests covering the new draft operations and tier inclusion

### DIFF
--- a/README.md
+++ b/README.md
@@ -748,6 +748,8 @@ cp .env.oauth21 .env
 | <sub>`manage_gmail_label`</sub> | <sub>Extended</sub> | <sub>Create/update/delete labels</sub> |
 | <sub>`manage_gmail_filter`</sub> | <sub>Extended</sub> | <sub>Create or delete Gmail filters</sub> |
 | <sub>`draft_gmail_message`</sub> | <sub>Extended</sub> | <sub>Create drafts</sub> |
+| <sub>`update_gmail_draft`</sub> | <sub>Extended</sub> | <sub>Replace draft content</sub> |
+| <sub>`delete_gmail_draft`</sub> | <sub>Extended</sub> | <sub>Delete drafts</sub> |
 | <sub>`get_gmail_threads_content_batch`</sub> | <sub>Complete</sub> | <sub>Batch retrieve thread content</sub> |
 | <sub>`batch_modify_gmail_message_labels`</sub> | <sub>Complete</sub> | <sub>Batch modify labels</sub> |
 | <sub>`start_google_auth`</sub> | <sub>Complete</sub> | <sub>Legacy OAuth 2.0 auth (disabled when OAuth 2.1 is enabled)</sub> |
--- a/core/tool_tiers.yaml
+++ b/core/tool_tiers.yaml
@@ -12,6 +12,8 @@ gmail:
     - list_gmail_labels
     - manage_gmail_label
     - draft_gmail_message
+    - update_gmail_draft
+    - delete_gmail_draft
     - list_gmail_filters
     - manage_gmail_filter
 
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -814,9 +814,13 @@ def _build_attachment_error_entry(
 ) -> Dict[str, Any]:
     """Preserve failed attachment context so message creation can continue."""
     failed_attachment = dict(attachment)
+    error_detail = str(exc)
     if "url" in failed_attachment:
-        failed_attachment["display_url"] = _redact_url(str(failed_attachment["url"]))
-    failed_attachment["error"] = str(exc)
+        raw_url = str(failed_attachment["url"])
+        redacted_url = _redact_url(raw_url)
+        failed_attachment["display_url"] = redacted_url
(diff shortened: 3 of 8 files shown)


future3OOO:
Update on scope after the latest force-push (`d80db44`): I intentionally stripped this PR back to a narrower draft lifecycle contract.

The earlier version tried to be a preservation/patch-style draft updater: it fetched the existing raw draft, parsed and rebuilt MIME, preserved omitted fields and attachments, handled inline/CID parts, and shared more MIME-building behavior with existing draft creation. That made the PR larger and increased the review surface around existing send/create-draft behavior.

The current version is deliberately replacement-style instead:

- `update_gmail_draft` replaces the draft message by `draft_id` from the fields supplied by the caller.
- `subject` and `body` are required replacement inputs.
- Optional recipients, sender alias, threading headers, and attachments are replacement inputs when supplied.
- Omitted optional fields are omitted from the rebuilt message; the tool no longer fetches or preserves existing draft MIME.
- Attachments are replacement attachments, not preserved attachments.
- `delete_gmail_draft` deletes by `draft_id`.

What was removed from the intended contract:

- raw draft fetch/parse before update
- preserve-on-omit semantics
- empty-list attachment clearing as a special patch operation
- MIME preservation for inline/CID parts, nested MIME, `message/rfc822`, unnamed parts, and original attachment disposition
- shared MIME rewrite changes that affected existing `send_gmail_message` / `draft_gmail_message` behavior

The diff is now `8 files changed, 402 insertions(+), 7 deletions(-)`, mostly the two tools, tier wiring, docs, and focused tests.

I also ran live Gmail proof against the configured account:

- create temporary draft -> update it -> fetch raw MIME and verify replacement subject/body -> delete -> verify Gmail returns 404
- repeat through the local MCP `streamable-http` transport
- test update attachments with base64 content, allowed local path, and HTTPS URL attachments; all worked when within the existing path/SSRF safety boundaries
- HTML signature behavior was verified with `body_format="html"` and `include_signature=True`

So the contract I would like reviewed now is the narrower one: complete draft lifecycle support for create/update/delete, where update is a full replacement of the draft message rather than an arbitrary MIME-preserving patcher.

PR #700, opened 2026-04-17
feat(auth): add list_configured_accounts tool for multi-account discovery

asreynolds1000:
## Description

Adds a new read-only MCP tool `list_configured_accounts` that enumerates Google accounts with stored credentials in the configured credential store, with per-account metadata (scopes, expiry, last-refreshed timestamp, refresh-token presence).

This is a discovery primitive for multi-account MCP clients. Rather than hardcoding an account list in the client's prompt or config, the client invokes this tool once and fans out across the returned emails. Useful for cross-account email triage, calendar sweeps, and any workflow that synthesizes across multiple authenticated Google accounts.

The tool does **not** call any Google API and does **not** use `@require_google_service`. It reads local credential-store state only (`auth/credential_store.py`'s `list_users()` + `get_credential()`).

Per-account load errors (malformed JSON, file deleted mid-call, permission issues) are captured in an `errors` array rather than raised, so partial credential-store corruption doesn't fail the whole call.

## Type of Change

## Registration

The tool is registered in both entrypoints and participates in tier filtering:

- **`main.py`**: `import_module("auth.account_tools")` runs unconditionally *before* the tier-gated `tool_imports` dict, so the tool registers regardless of which Google services the user loads (`--tools gmail` still exposes account discovery).
- **`fastmcp_server.py`**: imported alongside peer tool modules.
- **`core/tool_tiers.yaml`**: listed under `gmail.core` so `--tool-tier core` users still get the tool through `filter_server_tools`. A dedicated `auth` top-level service would have required a matching `tool_imports` entry, so this placement is pragmatic and documented with a comment in the YAML.

## Testing

- Added `tests/auth/test_account_tools.py` — 6 pytest cases covering: empty store, two-account enumeration, expiry surfacing, error-not-exception behavior on corrupt credentials, non-credential-file filtering (e.g., `oauth_states.json` skipped), and store-level failure path.
- Teardown uses the public `set_credential_store` API rather than poking the private `_credential_store` global.
- Full repo test suite passes: `tests/gmail/ tests/auth/` → 103 passed.
- Manually verified against a populated credential directory: all accounts returned with correct scopes, expiry, and last-refreshed timestamps.

## Checklist

- [x] My code follows the style guidelines of this project (`_impl`-function-plus-`@server.tool`-wrapper pattern, matching `_rsvp_event_impl` and similar)
- [x] I have commented my code, particularly in hard-to-understand areas (side-effect-free path reconstruction, narrowed exception handling rationale)

### Design choices worth flagging

1. **`_credential_file_mtime_iso` deliberately does not call `_get_credential_path`.** That helper creates `base_dir` via `os.makedirs` as a side effect, and a read-only discovery tool should never mutate disk. Instead, the mtime helper reconstructs the path inline with the same sanitization regex. A small duplication trade-off, but safer.

2. **Exception handling is narrowed, not broad.** `OSError` and `ValueError` at the store level (plus `AttributeError` at the per-account loop for misconfigured custom store backends). Programmer errors like `TypeError` and `KeyError` intentionally propagate so they surface loudly during development rather than being silently absorbed.

3. **Tier-registration placement under `gmail.core` is pragmatic.** Semantically the tool is a meta-tool that doesn't depend on any Google API scope. The cleanest design would be a top-level `auth` service in `tool_tiers.yaml`, but that would require matching changes in `main.py`'s `tool_imports` dict and I didn't want to expand the blast radius of a purely additive PR. Happy to restructure if you'd prefer a different placement.

### Return shape

```python
{
  "accounts": [
    {
      "email": str,
      "scopes": list[str] | None,
      "expiry": str | None,          # ISO-8601, may be naive UTC per google-auth's convention
      "last_refreshed": str | None,  # ISO-8601 UTC (credential file mtime)
      "has_refresh_token": bool,
    },
    ...
  ],
  "errors": [{"email": str | None, "error": str}, ...],
  "count": int,
  "store_type": str,   # class name of the active CredentialStore
}
```

No auth-layer or credential-store changes; purely additive.

## Summary by CodeRabbit

* **New Features**
  * Added an account-discovery tool that enumerates locally configured Google accounts and returns email, scopes, expiry, last-refreshed timestamps, and refresh-token presence; this tool is now registered at startup and exposed in the Gmail "core" tier.

* **Tests**
  * Added tests covering empty and multiple credentials, expiry/last-refreshed formatting, malformed files being reported (without crashing), ignored non-credential files, and credential-store listing failures.

### DIFF
--- a/auth/account_tools.py
+++ b/auth/account_tools.py
@@ -1,0 +1,162 @@
+"""
+Account enumeration tools for the Google Workspace MCP server.
+
+Exposes `list_configured_accounts` — a read-only tool that reports which Google
+accounts currently have stored credentials in the configured credential store.
+Useful for multi-account fan-out patterns in MCP clients: discover the account
+list at runtime rather than hardcoding it.
+
+This tool does NOT authenticate against any Google API and does NOT use
+`@require_google_service`. It reads local credential metadata only.
+"""
+
+import asyncio
+import logging
+import os
+from datetime import datetime, timezone
+from typing import Any
+from urllib.parse import quote
+
+from auth.credential_store import get_credential_store, LocalDirectoryCredentialStore
+from core.server import server
+
+logger = logging.getLogger(__name__)
+
+
+def _credential_file_mtime_iso(
+    store: LocalDirectoryCredentialStore, email: str
+) -> str | None:
+    """Return the credential file's mtime as an ISO-8601 UTC string, or None
+    if the file doesn't exist or the store isn't a local-directory store.
+
+    Side-effect-free: uses ``_resolve_credential_path`` (no ``os.makedirs``)
+    with the same ``quote()`` + legacy-fallback lookup that
+    ``_get_credential_path`` performs.
+    """
+    if not isinstance(store, LocalDirectoryCredentialStore):
+        return None
+    try:
+        ext = store.FILE_EXTENSION
+        safe_email = quote(email, safe="@._-")
+        path = store._resolve_credential_path(f"{safe_email}{ext}")
+        if not os.path.exists(path):
+            legacy = store._legacy_safe_email(email)
+            if legacy != safe_email:
+                path = store._resolve_credential_path(f"{legacy}{ext}")
+            if not os.path.exists(path):
+                return None
+        mtime = os.path.getmtime(path)
+        return datetime.fromtimestamp(mtime, tz=timezone.utc).isoformat()
+    except (OSError, ValueError) as e:
+        logger.debug(f"Could not read mtime for {email}: {e}")
+        return None
+
+
+async def _list_configured_accounts_impl() -> dict[str, Any]:
+    """Implementation of list_configured_accounts. Kept separate from the
+    @server.tool-decorated wrapper so tests can call it directly without
+    going through the MCP dispatch layer.
+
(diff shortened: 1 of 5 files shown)


asreynolds1000:
Rebased onto current main and addressed feedback:

- **Updated credential path logic** for the `quote()`-based sanitization from #724 — `_credential_file_mtime_iso` now uses `quote()` + `_resolve_credential_path()` with legacy fallback instead of the old regex, staying side-effect-free
- **Added `NotImplementedError` handling** for `GCSCredentialStore` (which doesn't support `list_users`) — returns a clean error response instead of crashing
- **Modernized type hints** — `dict[str, Any]` / `list[...]` instead of `typing.Dict` / `typing.List`
- **Added test** for the `NotImplementedError` path (7 tests total now)

@coderabbitai resume

asreynolds1000:
@coderabbitai review

asreynolds1000:
Addressed the review:

- Use `store.FILE_EXTENSION` instead of hardcoding `".json"`
- Sanitized error payloads — only `type(e).__name__` goes to clients, full details stay in logs

@coderabbitai review

asreynolds1000:
Closing — this was already merged into main directly. Thanks!

PR #701, opened 2026-04-17
feat(gmail): add analyze_thread_ownership tool for thread-state analysis

asreynolds1000:
## Description

Adds a new read-only Gmail tool `analyze_thread_ownership` that takes a thread ID and returns structured metadata about the thread's state: last non-draft sender, ball-in-court verdict (`"user"` | `"them"` | `None`), per-sender message counts, participant list, and draft-exclusion count.

**Why this exists.** A common MCP-client pattern is synthesizing who-owes-whom from a thread. Clients often infer this from a date-range search on sent/received messages, which is fragile — same-day follow-ups get missed, thread context is lost, and the "last message" determination ends up wrong. Reading the actual thread chronology via `threads().get(format='full')` eliminates the class of error.

This tool is a thin wrapper over that Gmail API call plus structured header extraction. The structured output shape is designed so downstream agents don't need to parse prose from `get_gmail_thread_content`.

## Type of Change

## Testing

**24 pytest cases total**, all passing:

- `tests/gmail/test_analyze_thread_ownership.py` (21 behavioral cases):
  - **Email normalization** — plus-addressing, display names, empty/malformed input
  - **Date parsing** — RFC822, internalDate fallback, both missing, malformed; plus a regression test for mixed naive/aware datetimes
  - **Ball-in-court logic** — vendor-last, user-last, plus-addressed user
  - **Draft exclusion** — newer draft doesn't flip ownership; all-drafts returns None
  - **Participants and counts** — 3-party thread, quoted-comma display names (e.g. `"Doe, John" <john@example.com>, vendor@example.com`), forwarded-to-self returns None
  - **Empty and single-message threads**, **malformed Date with internalDate fallback**, **malformed From returns None**
- `tests/gmail/test_analyze_thread_ownership_schema.py` (3 cases): follows the `test_modify_gmail_message_labels_schema.py` pattern. Asserts the tool is registered, required params are typed correctly, and no unexpected params have been added to the public contract.

Pure analysis (`_analyze_thread_ownership_impl`) is separated from the `@server.tool` wrapper so tests exercise the logic directly without network or auth. Follows the existing `_impl`-plus-wrapper pattern used by `_rsvp_event_impl`, `_create_event_impl`, etc.

Full repo test suite (`tests/gmail/ tests/auth/`): 121 passed.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas (timezone coercion rationale, ball-in-court edge cases, multi-alias limitation)

### Edge cases explicitly handled

- **Plus-addressing.** `alex+foo@domain.com` normalizes to `alex@domain.com` for identity comparison, so a user doesn't appear as a "third party" because of a tagged address.
- **Drafts.** Messages with the `DRAFT` label are excluded from the last-message determination and counted separately in `excluded_drafts`. A newer draft does NOT flip ball-in-court, since drafts haven't been sent.
- **Malformed Date header.** Falls back to Gmail's `internalDate` (Unix ms) for chronological ordering.
- **Self-only threads** (user → user, e.g., forward-to-self or archive-to-self): `ball_in_court_of = None`. There's no external party to owe anything, so downstream agents looking for "ball on them" shouldn't match.
- **Unparseable From** (display name only, no real email): `ball_in_court_of = None`. Rather than silently defaulting to `"user"`, surface the ambiguity.

### Timezone handling (subtle)

`email.utils.parsedate_to_datetime` returns a **naive** datetime when the RFC822 Date header has no timezone offset (e.g., `"Mon, 14 Apr 2026 09:00:00"`). The `internalDate` fallback path returns an **aware** UTC datetime. Mixing them in `>` / `<` comparisons raises `TypeError` at runtime, which would happen on real Gmail threads with any tz-less Date header alongside a malformed one. The helper coerces naive → UTC at parse time so downstream comparisons are safe. A dedicated regression test (`test_naive_and_aware_datetimes_do_not_raise`) covers this case.

### Multi-address parsing

Uses `email.utils.getaddresses` across From/To/Cc in a single call, rather than splitting by comma. This handles quoted display names with embedded commas correctly — e.g., `"Doe, John" <john@example.com>, vendor@example.com` resolves to two recipients, not three.

### Tier registration

Added to `core/tool_tiers.yaml` under `gmail.extended` so users running with `--tool-tier` see the tool.

### Known limitation

`user_google_email` is a single address. A user with multiple aliases (e.g., personal + work) will not be recognized as the same identity across a multi-account thread. Documented in the docstring.

### Return shape

```python
{
    "thread_id": str,
    "thread_subject": str | None,
    "last_sender": str | None,             # raw "Name <addr>" or None
    "last_timestamp": str | None,          # ISO-8601 (UTC-aware)
    "ball_in_court_of": "user" | "them" | None,
    "message_count_by_sender": {           # keyed by normalized email
        "sender@domain.com": int,
    },
    "participants": [                       # sorted, union of From/To/Cc
        "sender@domain.com",
    ],
    "excluded_drafts": int,
    "message_count": int,
}
```

No auth-layer or API-layer changes; purely additive.

## Summary by CodeRabbit

* **New Features**
  * Added Gmail thread ownership analysis tool that returns thread subject, last active non-draft sender, participants, per-sender message counts, draft exclusions, and a "ball in court" verdict.
* **Tests**
  * Added unit and schema tests validating ownership logic, email/date parsing, draft handling, participant extraction, and tool input schema.
* **Chores**
  * Registered the new tool in the tool tier configuration.

### DIFF
--- a/core/tool_tiers.yaml
+++ b/core/tool_tiers.yaml
@@ -8,6 +8,7 @@ gmail:
   extended:
     - get_gmail_attachment_content
     - get_gmail_thread_content
+    - analyze_thread_ownership
     - modify_gmail_message_labels
     - list_gmail_labels
     - manage_gmail_label
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -2381,6 +2381,261 @@ async def get_gmail_thread_content(
     )
 
 
+def _normalize_email(address: str) -> str:
+    """Lowercase an email address and strip plus-addressing so that
+    e.g. 'Alex <alex+foo@scopestack.io>' normalizes to 'alex@scopestack.io'.
+
+    This is the key primitive for 'is this message from Alex?' checks — plus
+    addresses are Alex, not a third party.
+    """
+    from email.utils import parseaddr
+
+    _name, addr = parseaddr(address or "")
+    addr = addr.lower().strip()
+    if not addr or "@" not in addr:
+        return addr
+    local, _, domain = addr.partition("@")
+    local = local.split("+", 1)[0]
+    return f"{local}@{domain}"
+
+
+def _parse_date_header(date_str: str, internal_date_ms: str | int | None):
+    """Parse a Date header to a timezone-aware datetime. Fall back to
+    internalDate (Unix ms as string) when the RFC822 Date header is malformed
+    or missing.
+
+    CRITICAL: the returned datetime is ALWAYS timezone-aware (UTC). Mixing
+    naive and aware datetimes in comparisons raises TypeError at runtime,
+    which matters here because a single thread can contain both well-formed
+    Date headers (with tz) and malformed ones (falling back to internalDate,
+    which is aware UTC). Coerce naive → UTC at parse time so downstream
+    callers can `>` / `<` the results freely.
+
+    Returns (iso_string, datetime) or (None, None) if both sources fail.
+    """
+    from email.utils import parsedate_to_datetime
+    from datetime import datetime, timezone
+
+    if date_str:
+        try:
+            dt = parsedate_to_datetime(date_str)
+            # parsedate_to_datetime returns a naive datetime when the header
+            # has no timezone offset (e.g., "Mon, 14 Apr 2026 09:00:00").
+            # Coerce to UTC so comparisons with internalDate fallbacks work.
+            if dt.tzinfo is None:
+                dt = dt.replace(tzinfo=timezone.utc)
+            return dt.isoformat(), dt
+        except (TypeError, ValueError):
+            pass
+
+    if internal_date_ms is not None:
+        try:
+            ms = int(internal_date_ms)
+            dt = datetime.fromtimestamp(ms / 1000.0, tz=timezone.utc)
+            return dt.isoformat(), dt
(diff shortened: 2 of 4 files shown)


taylorwilsdon:
I'm not sure I'm clear on why this would be a standalone tool - I'd much rather include whatever functionality you're looking for in the existing Gmail thread tools rather than adding an additional one. We're maxed out on tools (if you run everything you are at the OpenAI max tools payload length for native tool calling) and separating this requires the LLM to make an additional distinct tool call for something that will only be used when reading threads. If you can build this into an existing tool would be happy to merge!

asreynolds1000:
Makes sense — the tool-count ceiling and the "only useful when reading a thread anyway" point are both right. I'll close this PR and open a fresh one that adds the ownership analysis to `get_gmail_thread_content` as an opt-in parameter (defaulting to off, so existing behavior is unchanged). Return type becomes a dict when the flag is on, string otherwise.

Should have it up shortly. Thanks for the quick look.

asreynolds1000:
Closing per discussion — will reopen against get_gmail_thread_content with an opt-in include_analysis parameter so the ownership analysis is bundled into the existing thread-read path.

PR #702, opened 2026-04-17
feat(gmail): add include_analysis flag to get_gmail_thread_content

asreynolds1000:
## Description

Supersedes #701 per your feedback. Instead of adding a new `@server.tool`, this PR bundles the thread-ownership analysis into the existing `get_gmail_thread_content` as an opt-in `include_analysis: bool = False` parameter.

**Existing behavior preserved.** When `include_analysis=False` (the default), `get_gmail_thread_content` returns the same formatted string it always has. When `include_analysis=True`, it returns a dict `{"content": <existing string>, "analysis": <ownership metadata>}`. No existing caller is affected.

**Why the analysis exists.** A common MCP-client pattern is synthesizing who-owes-whom from a thread. Clients often infer this from a date-range search on sent/received messages, which is fragile — same-day follow-ups get missed, thread context is lost, and the "last message" determination ends up wrong. The helper reads the actual thread chronology returned by `threads().get(format='full')` and extracts structured metadata.

## Analysis shape

```python
{
    "thread_id": str,
    "thread_subject": str | None,
    "last_sender": str | None,         # raw "Name <addr>"
    "last_timestamp": str | None,      # ISO-8601, UTC-aware
    "ball_in_court_of": "user" | "them" | None,
    "message_count_by_sender": {normalized_email: int},
    "participants": [normalized_email, ...],   # sorted, union of From/To/Cc
    "excluded_drafts": int,
    "message_count": int,
}
```

`ball_in_court_of` names whose court the ball is in (who owes the next reply):
- `"user"` — authenticated user owes the reply (external party sent last)
- `"them"` — other party owes the reply (user sent last)
- `None` — unresolvable (all drafts, self-only thread, or malformed headers)

## Edge cases handled

- **Plus-addressing.** `alex+foo@domain.com` normalizes to `alex@domain.com` for identity comparison.
- **Drafts.** Messages with the `DRAFT` label are excluded from the last-message determination and counted separately in `excluded_drafts`. A newer draft does NOT flip ball-in-court.
- **Malformed Date headers.** Fall back to Gmail's `internalDate` (Unix ms) for chronological ordering.
- **Naive/aware datetime mixing.** `email.utils.parsedate_to_datetime` returns a naive datetime when the Date header has no timezone. The `internalDate` fallback is aware UTC. Mixing them in `>` / `<` comparisons raises `TypeError` at runtime — the helper coerces naive → UTC at parse time so downstream comparisons are safe. Regression test included.
- **Multi-address parsing.** Uses `email.utils.getaddresses` across From/To/Cc in one pass, not naive `split(",")`. Handles quoted display names with embedded commas like `"Doe, John" <john@example.com>, vendor@example.com` correctly.
- **Self-only threads** (user → user) and **unparseable From headers** return `ball_in_court_of=None` rather than silently defaulting.

## Type of Change

## Testing

Added two test files:

- `tests/gmail/test_thread_ownership_helpers.py` — **21 pytest cases** exercising the pure helpers (`_analyze_thread_ownership_impl`, `_normalize_email`, `_parse_date_header`) against fabricated `threads.get(format='full')` responses. No network, no credentials.
- `tests/gmail/test_get_gmail_thread_content_analysis.py` — **5 integration cases** covering the wire-up: default returns `str`, flag=True returns dict with both keys, analysis shape matches the documented contract, ball-in-court verdict is correct, content is byte-identical across both modes.

Full `tests/gmail/` suite passes: **95 passed**.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas (timezone coercion rationale, ball-in-court edge cases, multi-alias limitation, getaddresses vs split)

## Known limitation

`user_google_email` is a single address. A user with multiple aliases (e.g., personal + work) won't be recognized as the same identity across a multi-account thread. Documented in the docstring.

## Related

Closes #701 (replaced by this PR per maintainer guidance to bundle rather than add a new tool).

## Summary by CodeRabbit

* **New Features**
  * Thread retrieval can optionally include structured ownership analysis alongside the thread text (participants, per-sender counts, last sender/timestamp, and "ball in court"); text output remains identical with or without analysis.
  * Better handling of varied date formats and timezone normalization to UTC; consistent normalization of plus-address variants.

* **Tests**
  * Added tests validating flag behavior, content parity, ownership-analysis semantics, draft exclusion, complex recipient parsing, date/time fallbacks, and malformed-header robustness.

### DIFF
--- a/gmail/gmail_helpers.py
+++ b/gmail/gmail_helpers.py
@@ -1,0 +1,208 @@
+"""Email helper utilities for Gmail tools."""
+
+from __future__ import annotations
+
+import logging
+from collections import Counter
+from datetime import datetime, timezone
+from email.utils import getaddresses, parseaddr, parsedate_to_datetime
+from typing import Any, Optional
+
+logger = logging.getLogger(__name__)
+
+
+def _normalize_email(address: str) -> str:
+    """Lowercase an email address and strip plus-addressing so that
+    e.g. 'Alex <alex+foo@scopestack.io>' normalizes to 'alex@scopestack.io'.
+
+    This is the key primitive for 'is this message from Alex?' checks - plus
+    addresses are Alex, not a third party.
+    """
+    _name, addr = parseaddr(address or "")
+    addr = addr.lower().strip()
+    if not addr or "@" not in addr:
+        return addr
+    local, _, domain = addr.partition("@")
+    local = local.split("+", 1)[0]
+    return f"{local}@{domain}"
+
+
+def _parse_date_header(
+    date_str: str, internal_date_ms: str | int | None
+) -> tuple[Optional[str], Optional[datetime]]:
+    """Parse Gmail internalDate or a Date header to a UTC-aware datetime.
+
+    Prefer Gmail's internalDate because it reflects Gmail's message ordering;
+    fall back to the Date header when internalDate is unavailable or malformed.
+    Always returns UTC-aware datetimes so naive/aware comparisons don't raise
+    TypeError.
+
+    Returns (iso_string, datetime) or (None, None) if both sources fail.
+    """
+    if internal_date_ms is not None:
+        try:
+            ms = int(internal_date_ms)
+            dt = datetime.fromtimestamp(ms / 1000.0, tz=timezone.utc)
+            return dt.isoformat(), dt
+        except (TypeError, ValueError) as e:
+            logger.debug(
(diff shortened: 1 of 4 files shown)


taylorwilsdon:
Thanks!

taylorwilsdon:
@coderabbitai review

taylorwilsdon:
@coderabbitai re-review please

PR #707, opened 2026-04-20
feat(calendar): add send_updates to create/update/delete event actions

azelcs:
### Summary

`manage_event` exposes a `send_updates` parameter, but it's currently wired only for the `rsvp` action. The `create`, `update`, and `delete` actions ignore it, so there's no way to control attendee notifications for those flows.

### Changes

- Threaded `send_updates` through `_create_event_impl`, `_modify_event_impl`, and `_delete_event_impl`, forwarding it to the underlying `events().insert / update / delete` calls.
- Updated the `manage_event` docstring to reflect that `send_updates` now applies to create, update, delete, and rsvp.
- Centralized validation of the `send_updates` value in `manage_event` itself, rather than repeating it in each impl.
- Default remains `"all"` across all four actions, matching the existing `rsvp` behavior.

### Behavior note

Google Calendar's implicit default when `sendUpdates` is omitted is effectively "none". Defaulting to `"all"` here is intentional and consistent with the RSVP path's existing default — the tool's opinion is "notify attendees by default; callers can opt out with `externalOnly` or `none`".

## Summary by CodeRabbit

* **New Features**
  * Event management now consistently accepts a send_updates option for create, update, delete and RSVP operations.

* **Bug Fixes**
  * Centralized validation for send_updates so invalid values are rejected uniformly across all event actions.

* **Tests**
  * Added coverage ensuring send_updates validation applies to all actions; removed the old RSVP-only validation test.

### DIFF
--- a/gcalendar/calendar_tools.py
+++ b/gcalendar/calendar_tools.py
@@ -669,6 +669,7 @@ async def _create_event_impl(
     guests_can_modify: Optional[bool] = None,
     guests_can_invite_others: Optional[bool] = None,
     guests_can_see_other_guests: Optional[bool] = None,
+    send_updates: str = "all",
 ) -> str:
     """Internal implementation for creating a calendar event."""
     logger.info(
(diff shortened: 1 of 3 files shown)


taylorwilsdon:
Thanks!

PR #710, opened 2026-04-21
fix(auth): retry legacy OAuth callback without PKCE when Google rejects verifier

darrenapfel:
Summary:
This fixes a legacy OAuth callback edge case in the single-user / legacy auth path.

In some re-auth flows, Google redirects back without a `state` parameter. The existing fallback recovers the stored PKCE verifier and retries token exchange with that verifier, but Google can reject the exchange with:

`(invalid_grant) code_verifier or verifier is not needed.`

This patch keeps the existing behavior, but if that exact verifier-rejected error occurs, it retries the token exchange once without PKCE instead of failing the auth flow.

Changes:
- add `_is_pkce_verifier_not_needed_error()` helper
- in `handle_auth_callback()`, retry `fetch_token()` once without `code_verifier` when Google returns the specific `invalid_grant` verifier-not-needed error
- add a regression test covering the missing-state fallback + verifier-rejected path

Why:
This was reproduced locally against a streamable-http + `--single-user` deployment using Desktop OAuth credentials. The first callback failed with the verifier-not-needed error; after applying this retry logic, auth completed successfully.

Validation:
- `.venv/bin/python -m pytest tests/auth/test_google_auth_pkce.py tests/auth/test_google_auth_callback_refresh_token.py -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection and logging for a specific Google OAuth token-exchange error to make failures clearer during sign-in callbacks.

* **Tests**
  * Added a test to ensure the sign-in callback surfaces the specific OAuth error case and updated test helpers to support the new scenario.

### DIFF
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -677,20 +677,13 @@ def handle_auth_callback(
             flow.fetch_token(authorization_response=authorization_response)
             credentials = flow.credentials
         except Exception as exc:
-            if not _is_pkce_verifier_not_needed_error(exc):
-                raise
-
-            logger.warning(
-                "OAuth token exchange rejected PKCE verifier; retrying callback without PKCE."
-            )
-            fallback_flow = create_oauth_flow(
-                scopes=scopes,
-                redirect_uri=redirect_uri,
-                state=state,
-                autogenerate_code_verifier=False,
-            )
-            fallback_flow.fetch_token(authorization_response=authorization_response)
-            credentials = fallback_flow.credentials
+            if _is_pkce_verifier_not_needed_error(exc):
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
Coderabbit has it right here, but we can still patch for the behavior:
```
Recommended fix: If the first fetch_token fails due to PKCE constraints, the user must restart the authorization flow to obtain a fresh authorization code. Alternatively, if the error indicates a client-side configuration mismatch (code verifier not provided when expected, or vice versa), detect and fix the configuration before attempting the initial exchange rather than retrying with the consumed code.
```

Will push a refac

Issue #544, opened 2026-03-04
modify_gmail_message_labels broken after FastMCP v3 upgrade — list params received as strings


fmc08:
Here it is — just copy and paste the whole thing:
**Version:** 1.14.2

**Describe the bug:**
After the FastMCP v3 upgrade in v1.13.0, the `modify_gmail_message_labels` tool fails with a Pydantic validation error when called with `add_label_ids` or `remove_label_ids`. The parameters arrive as a JSON string (`'["Label_3"]'`) instead of a parsed list, causing the `Optional[List[str]]` type check to fail.

**Error:**
1 validation error for call[modify_gmail_message_labels]
add_label_ids
  Input should be a valid list [type=list_type, input_value='["Label_3"]', input_type=str]

**Expected behavior:** Label IDs passed as a JSON array are correctly deserialized into a Python list.

**Possible fix:** Add a Pydantic field validator on `add_label_ids` and `remove_label_ids` to coerce a string input into a list (e.g. `json.loads(v)` if `isinstance(v, str)`).

taylorwilsdon:
Hm, I can't reproduce. Just updated a label right now without issues on the current main branch. What client are you using? 

Claude Code doesn't have an issue - 
<img width="679" height="231" alt="Image" src="https://github.com/user-attachments/assets/e5df6a90-7eae-4d0e-a20d-f7bc5d206d49" />

The error in the issue is a Pydantic validation error — meaning it fails before the tool function body even executes. Any change on the tool side would never execute, it's a client-side serialization issue. FastMCP v3 didn't change how Pydantic validates tool input parameters, Optional[List[str]] has always expected an actual list rather than a JSON string so if the client sends a properly structured call, the list should arrive as a list.

fmc08:
Thanks for the quick response! I'm using Cowork, Anthropic's desktop app feature (powered by the Claude Agent SDK). The MCP call is being made by a Claude agent running inside Cowork, and it appears that client is serializing list parameters as strings rather than proper JSON arrays. Claude Code works fine — this seems to be specific to Cowork's MCP client implementation. Likely an Anthropic-side bug to report rather than anything on your end. Appreciate you looking into it!

florianlemenach:
For anyone hitting this from Cowork — I've added add_gmail_message_label and remove_gmail_message_label tools on this fork: https://github.com/florianlemenach/google_workspace_mcp/tree/feature/single-gmail-message-label
[...]

PR #613, opened 2026-03-25
fix: Issues/611

taylorwilsdon:
Closes #611 

## Summary by CodeRabbit

* **Tests**
  * Added schema validation tests for Gmail label modification operations to ensure correct parameter definitions.

* **Bug Fixes**
  * Improved JSON schema generation for Gmail label ID parameters to explicitly enforce array-of-strings constraints, enhancing validation accuracy.

Issue #712, opened 2026-04-21
run_script_function broken after FastMCP v3 upgrade — parameters list received as JSON string


qrade:
## Preamble

Hi, opening this issue on behalf of Claude - I ran into this bug while using the MCP from Claude, Claude diagnosed it and asked me to report it upstream in the hope it's a useful heads-up. For full disclosure: the issue text below was also drafted by Claude based on the stack trace I hit and the codebase. Happy to provide any extra environment detail if useful.

## Summary
Same regression pattern as #544, but affecting `run_script_function` in the Apps Script module. When a client passes `parameters` as a JSON-serialized string (which several MCP clients do for array args), the tool fails pydantic validation before ever reaching the Apps Script API.

## Reproduction
Call the tool with an array of one string:

```json
{
  "script_id": "…",
  "function_name": "myFunction",
  "parameters": ["PrestaShop"]
}
```

## Error

1 validation error for call[run_script_function]
parameters
Input should be a valid list [type=list_type, input_value='["PrestaShop"]', input_type=str]

The client serialized the array to a string; pydantic rejects it because the signature declares `list[object]`.

## Location
`gappsscript/apps_script_tools.py` — `run_script_function`:

```python
async def run_script_function(
    service: Any,
    user_google_email: str,
    script_id: str,
    function_name: str,
    parameters: Optional[list[object]] = None,
    dev_mode: bool = False,
) -> str:
```

## Proposed fix
Coerce JSON-string input via a `BeforeValidator` (same approach discussed in #544):

```python
from pydantic import BeforeValidator
from typing import Annotated
import json

def _coerce_json_list(v):
    if isinstance(v, str):
        try:
            parsed = json.loads(v)
        except json.JSONDecodeError:
            return v
        if isinstance(parsed, list):
            return parsed
    return v

ParametersList = Annotated[Optional[list[object]], BeforeValidator(_coerce_json_list)]

async def run_script_function(
    ...,
    parameters: ParametersList = None,
    dev_mode: bool = False,
) -> str:
```

The same coercion likely belongs on every `list[...]` tool argument post-FastMCP v3 upgrade (related: #544, #180).

## Environment
- Package version: latest `main` (post-FastMCP v3 upgrade, v1.13.0+)
- Client: Claude via Cowork mode, MCP transport

Issue #689, opened 2026-04-14
Extend BeforeValidator coercion to List[dict] params (attachments) for Cowork compatibility


km-mf:
## Summary

`draft_gmail_message` and `send_gmail_message` fail when called from **Claude Cowork** because the `attachments` parameter arrives as a JSON string instead of `List[dict]`.

We understand this is a client-side serialization issue in Cowork (tracked at anthropics/claude-code#18260 and anthropics/claude-code#27492), but since PR #603 already added a great server-side workaround for `List[str]` params via `BeforeValidator`, it would be awesome to extend the same pattern to `List[dict]` params like `attachments` — giving Cowork users a working experience until the upstream client fix lands.

## Error

```
1 validation error for call[draft_gmail_message]
attachments
  Input should be a valid list [type=list_type, input_value='[{"path":"/Users/...screenshot_1.png"}]', input_type=str]
```

## Reproduction

1. Configure workspace-mcp locally via `claude_desktop_config.json`
2. Open Claude Desktop → Cowork tab
3. Ask Claude to draft an email with a file attachment
4. Cowork serializes `attachments` as `"[{\"path\": \"...\"}]"` (string) instead of `[{"path": "..."}]` (array)
5. Pydantic rejects it before the tool function executes

Works correctly from Claude Code (CLI) — only Cowork is affected.

## Proposed Fix

The existing `_coerce_json_str_to_list` in `core/utils.py` is already generic enough. Just add a `DictList` type alongside `StringList`:

```python
# Existing (from PR #603):
StringList = Annotated[List[str], BeforeValidator(_coerce_json_str_to_list)]

# New — same pattern for dict lists:
DictList = Annotated[List[dict], BeforeValidator(_coerce_json_str_to_list)]
```

Then apply `DictList` to the `attachments` parameter in `draft_gmail_message` and `send_gmail_message`.

## Related

- #544 — original Cowork stringification report
- #603 — `StringList` workaround (merged)
- #611 / #613 — schema-level fixes for missing `type: array`
- anthropics/claude-code#18260 — Anthropic master tracking issue
- modelcontextprotocol/typescript-sdk#1562 — root cause in MCP SDK

Happy to open a PR for this if you'd prefer. Thank you for the great project and for the PR #603 workaround pattern!

taylorwilsdon:
PR welcome or I can push in next update, thanks!

taylorwilsdon:
[...]

PR #715, opened 2026-04-22
fix: make label_ids schemas compatible with Moonshot/Kimi strict JSON Schema

### LINKED ISSUE #611: modify_gmail_message_labels — missing type in JSON Schema
# Bug Report: modify_gmail_message_labels — missing type in JSON Schema

## Summary
The `modify_gmail_message_labels` tool in the Google Workspace Gmail MCP connector has a schema definition error: the `add_label_ids` and `remove_label_ids` parameters are missing the `"type": "array"` specification, causing all values to be deserialized as strings instead of lists.

## Current (broken) schema
```json
"add_label_ids": {"default": null},
"remove_label_ids": {"default": null}
```

## Expected (correct) schema
```json
"add_label_ids": {"default": null, "type": "array", "items": {"type": "string"}},
"remove_label_ids": {"default": null, "type": "array", "items": {"type": "string"}}
```

## Comparison with working tool
The `get_gmail_messages_content_batch` tool in the same connector correctly defines its array parameter:
```json
"message_ids": {"items": {"type": "string"}, "type": "array"}
```

## Reproduction steps
1. Call `modify_gmail_message_labels` with any valid message_id
2. Pass `add_label_ids` as `["Label_57"]` (JSON array format)
3. Observe error:
```
Input should be a valid list [type=list_type, input_value='["Label_57"]', input_type=str]
```
4. The value arrives as the string `'["Label_57"]'` instead of the Python list `["Label_57"]`

## Impact
- **All label modification operations on messages are broken** — adding and removing labels from messages is impossible through this MCP tool
- Label creation, deletion, listing, and message search/read operations work correctly
- This blocks email organization workflows that depend on programmatic label management

## Environment
- MCP connector: google-workspace-gmail (built-in Anthropic/Claude integration)
- Date: 2026-03-25

## Suggested fix
Add `"type": "array"` and `"items": {"type": "string"}` to both `add_label_ids` and `remove_label_ids` parameter definitions in the tool's JSON Schema, consistent with how `message_ids` is defined in `get_gmail_messages_content_batch`.

## Related
- GitHub issue #608 on taylorwilsdon/google_workspace_mcp: "apply_gmail_label / remove_gmail_label missing from tool_tiers.yaml"

michabbb:
## Problem

When connecting this MCP server to an agent whose LLM is Moonshot/Kimi (e.g. `kimi-k2.6`), the entire tool list is rejected with:

```
Invalid request: tools.function.parameters is not a valid moonshot flavored json schema,
details: <At path 'properties.add_label_ids': when using anyOf, type should be
defined in anyOf items instead of the parent schema>
```

## Root cause

Commit [`PR #613`](https://github.com/taylorwilsdon/google_workspace_mcp/pull/613) (closes #611) added `json_schema_extra={"type": "array", "items": {"type": "string"}}` to the `add_label_ids` / `remove_label_ids` parameters. Because those parameters are declared as `Optional[StringList]`, Pydantic already emits an `anyOf: [{"type":"array","items":{"type":"string"}},{"type":"null"}]` schema. Combining both produces a hybrid schema:

```json
{
  "anyOf": [
    {"type": "array", "items": {"type": "string"}},
    {"type": "null"}
  ],
  "default": null,
  "type": "array",
  "items": {"type": "string"}
}
```

OpenAI and Anthropic tolerate this redundancy; Moonshot/Kimi rejects it because the strict moonshot-flavored schema requires `type` to live **inside** each `anyOf` branch, not on the parent.

## Fix

Remove `json_schema_extra` from the two affected tool signatures. Pydantic then generates the standard Moonshot-compatible form:

```json
{
  "anyOf": [
    {"type": "array", "items": {"type": "string"}},
    {"type": "null"}
  ],
  "default": null
}
```

Affected functions (both in `gmail/gmail_tools.py`):
- `modify_gmail_message_labels`
- `batch_modify_gmail_message_labels`

## Why this is safe

- The `StringList = Annotated[List[str], BeforeValidator(_coerce_json_str_to_list)]` alias (from `core/utils.py`) already provides runtime coercion for MCP clients that serialise arrays as JSON strings — i.e. the original problem that motivated #611 (`'[\"Label_57\"]'` → `[\"Label_57\"]`) is still handled by the `BeforeValidator`, not by `json_schema_extra`.
- The emitted schema now contains `{\"type\": \"array\", \"items\": {\"type\": \"string\"}}` inside the first `anyOf` branch, so clients that rely on `type: array` to pass native lists still see it — just in the correct JSON-Schema location.

## Verification

- Inspected the live MCP `tools/list` output before/after — `anyOf` branches now include `type`, and no duplicate parent-level `type` / `items`.
- Called `list_gmail_labels` and verified Claude Code (Anthropic) still parses the schema and the tool still returns successfully.
- The same server running behind a Moonshot/Kimi-backed agent no longer raises the `anyOf`-location error.

## Compatibility notes

No behavioural change for existing clients; only the JSON-Schema shape changes. `modify_gmail_message_labels` / `batch_modify_gmail_message_labels` still accept:
- `["Label_57"]` (native array)
- `'[\"Label_57\"]'` (JSON-encoded string — via the unchanged `BeforeValidator`)
- `None` / omission (default)

Closes (partially) the integration gap for Moonshot/Kimi-hosted agents.

## Summary by CodeRabbit

* **Refactor**
  * Simplified how optional label ID lists are represented for Gmail label-modification tools; behavior and defaults unchanged.
* **Tests**
  * Updated and refactored tests to validate the new schema representation for optional label lists (array of strings or null) and reduced duplicated assertions via a shared helper.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -2806,14 +2806,8 @@ async def modify_gmail_message_labels(
     service,
     user_google_email: str,
     message_id: str,
-    add_label_ids: Annotated[
-        Optional[StringList],
-        Field(json_schema_extra={"type": "array", "items": {"type": "string"}}),
-    ] = None,
-    remove_label_ids: Annotated[
-        Optional[StringList],
-        Field(json_schema_extra={"type": "array", "items": {"type": "string"}}),
-    ] = None,
+    add_label_ids: Optional[StringList] = None,
+    remove_label_ids: Optional[StringList] = None,
 ) -> str:
     """
     Adds or removes labels from a Gmail message.
@@ -2864,14 +2858,8 @@ async def batch_modify_gmail_message_labels(
     service,
     user_google_email: str,
     message_ids: StringList,
-    add_label_ids: Annotated[
-        Optional[StringList],
-        Field(json_schema_extra={"type": "array", "items": {"type": "string"}}),
-    ] = None,
-    remove_label_ids: Annotated[
-        Optional[StringList],
-        Field(json_schema_extra={"type": "array", "items": {"type": "string"}}),
-    ] = None,
+    add_label_ids: Optional[StringList] = None,
+    remove_label_ids: Optional[StringList] = None,
 ) -> str:
     """
     Adds or removes labels from multiple Gmail messages in a single batch request.
--- a/tests/gmail/test_modify_gmail_message_labels_schema.py
+++ b/tests/gmail/test_modify_gmail_message_labels_schema.py
@@ -3,23 +3,35 @@ from core.tool_registry import get_tool_components
 import gmail.gmail_tools  # noqa: F401
 
 
-def test_modify_gmail_message_labels_optional_arrays_publish_array_type():
+def _assert_optional_string_array_anyof(field_schema):
+    """Assert that field_schema describes `Optional[List[str]]` in Moonshot-
+    compatible form: an ``anyOf`` with an array-of-strings branch and a null
+    branch, and no redundant parent-level ``type``/``items``.
+    """
+    assert "anyOf" in field_schema
+    assert "type" not in field_schema, (
+        "Parent-level `type` alongside `anyOf` breaks Moonshot/Kimi strict "
+        "JSON Schema validation"
+    )
+    assert "items" not in field_schema
+    assert field_schema["default"] is None
+
+    branches = field_schema["anyOf"]
+    assert {"type": "array", "items": {"type": "string"}} in branches
+    assert {"type": "null"} in branches
+
+
+def test_modify_gmail_message_labels_optional_arrays_publish_anyof_array_null():
     components = get_tool_components(server)
     schema = components["modify_gmail_message_labels"].parameters["properties"]
 
     for field_name in ("add_label_ids", "remove_label_ids"):
-        field_schema = schema[field_name]
-        assert field_schema["type"] == "array"
-        assert field_schema["items"] == {"type": "string"}
-        assert field_schema["default"] is None
+        _assert_optional_string_array_anyof(schema[field_name])
 
 
-def test_batch_modify_gmail_message_labels_optional_arrays_publish_array_type():
+def test_batch_modify_gmail_message_labels_optional_arrays_publish_anyof_array_null():
     components = get_tool_components(server)
     schema = components["batch_modify_gmail_message_labels"].parameters["properties"]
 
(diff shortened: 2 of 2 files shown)


taylorwilsdon:
So I'd love to have this, used to be how it was but these shims ie:
```
    add_label_ids: Annotated[
        Optional[StringList],
        Field(json_schema_extra={"type": "array", "items": {"type": "string"}}),
    ] = None,
```

Are only there because of a bizarre Claude Cowork bug with FastMCP v3 schemas (ref #712, #689, #611 etc) so this fix would break those.

michabbb:
claude code speaking here 👋 — jumping in on behalf of the PR author to share a hypothesis, happy to be wrong on any of this.

Could it be that the shim is actually redundant, and the real Cowork compatibility already lives inside `StringList` itself?

In `core/utils.py`:
```python
StringList = Annotated[List[str], BeforeValidator(_coerce_json_str_to_list)]
```

That `BeforeValidator` turns `'["Label_57"]'` into `["Label_57"]` *before* Pydantic validates, so runtime coercion happens regardless of the schema shape. If that reading is right, it would be PR #603 that actually closed #544 / unblocked Cowork, and the schema hint in #613 was belt-and-suspenders on top.

A couple of observations that seem to point in that direction — but you'd know the Cowork quirks way better than me:

- There are ~14 other `Optional[StringList]` params across `gcalendar` (`attendees`, `attachments`, `recurrence`, `calendar_ids`), `gcontacts`, `gsheets` (`sheet_names`), `gsearch` (`sites`) that have never had the `json_schema_extra` shim, and I haven't found Cowork complaints against those.
- In #611, the two Gmail label tools still had plain `Optional[List[str]]` (not `StringList`) at the time, so Cowork's string-serialized array hit pydantic raw. PR #613 upgraded them to `Optional[StringList]` *and* added the shim — could be that the type upgrade alone was doing the heavy lifting?

If that's the case, dropping the shim would:
- keep Cowork working (same `BeforeValidator` mechanism the other 14 params rely on),
- unblock strict JSON-Schema validators (Moonshot/Kimi reject parent-level `type` alongside `anyOf`),
- make the Gmail label tools consistent with the rest of the repo.

Totally possible I'm missing a Cowork corner case that specifically needed the schema hint — if so, would love to hear what it was so we can document it near `StringList`. Otherwise, happy to add a short comment on `StringList` itself so future contributors don't accidentally re-add the shim.

PR #717, opened 2026-04-22
fix(auth): release googleapiclient Resource cycles to prevent OOM

azelcs:
Runs a garbage collection pass in require_google_service's finally block, right after service.close(). googleapiclient's Resource objects reference each other in a loop that close() doesn't break, so without an explicit collection the memory from each call piles up until the memory runs out.

## Verification
- Local reproduction at 384Mi memory limit: previously OOM'd on `modify_sheet_values` around call 7.
  With this change, RSS plateaus at ~216 MB across 24+ tool calls.

## Summary by CodeRabbit

* **Chores**
  * Improved memory management efficiency in the authentication service.

### DIFF
--- a/auth/service_decorator.py
+++ b/auth/service_decorator.py
@@ -1,3 +1,4 @@
+import gc
 import inspect
 import json
 import logging
@@ -792,6 +793,10 @@ def require_google_service(
             finally:
                 if service:
                     service.close()
+                # googleapiclient's Resource tree holds circular refs that
+                # service.close() doesn't break. Without this sweep, memory
+                # grows on every call.
+                gc.collect()
 
         # Set the wrapper's signature to the one without 'service'
         wrapper.__signature__ = wrapper_sig


taylorwilsdon:
Hm - I cannot reproduce this behavior. I have instances with thousands of users that have been up for weeks/months and memory usage is static. I've got 5k users on a container with 1gb of memory that never gets close to the limit. What platform are you on? I recall there was a windows-specific memory leak at one point tied to fastmcp.

azelcs:
Reproduced on latest main:
```dockerfile
FROM python:3.14.3-slim

COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /usr/local/bin/

RUN apt-get update && apt-get install -y --no-install-recommends git && rm -rf /var/lib/apt/lists/*
RUN uv venv /opt/venv && uv pip install --python /opt/venv \
    "workspace-mcp @ git+https://github.com/taylorwilsdon/google_workspace_mcp@main"

ENV PATH="/opt/venv/bin:$PATH"
ENTRYPOINT ["workspace-mcp"]
CMD ["--transport", "streamable-http", "--tools", "drive", "docs", "sheets"]
```

And running:
```bash
docker build --no-cache -t workspace-mcp .
source .env && export GOOGLE_OAUTH_CLIENT_ID GOOGLE_OAUTH_CLIENT_SECRET
docker run --rm --name workspace-mcp -p 3012:8080 --memory 384m --memory-swap 384m \
  -e GOOGLE_OAUTH_CLIENT_ID \
  -e GOOGLE_OAUTH_CLIENT_SECRET \
  -e MCP_ENABLE_OAUTH21=true \
  -e EXTERNAL_OAUTH21_PROVIDER=true \
  -e WORKSPACE_MCP_STATELESS_MODE=false \
  -e WORKSPACE_EXTERNAL_URL=http://localhost:3012 \
  -e WORKSPACE_MCP_PORT=8080 \
  -e WORKSPACE_MCP_HOST=0.0.0.0 \
  workspace-mcp
```

Calling `modify_sheet_values` repeatedly seems to allocate extra 30-40 MB per call. Will investigate more and perhaps looking for a better solution rather than `gc.collect()`.

Edit: Most likely related to https://github.com/googleapis/google-api-python-client/issues/535

taylorwilsdon:
Maintainer edits are not enabled.

PR #719, opened 2026-04-23
Fix OAuth callback server silently accepting foreign listener on callback port

rpeck:
## Problem

In stdio mode, `MinimalOAuthServer.is_actually_running()` uses a raw TCP connect probe to decide if the OAuth callback server is already running. Any process listening on the callback port (default 8000) — including completely unrelated local servers — causes `connect_ex()` to return 0, which makes `ensure_oauth_callback_available()` skip starting the OAuth server and return success. The OAuth flow then fails silently: Google redirects to `localhost:8000/oauth2callback` and hits the foreign server instead of ours.

Reproduced with another local server running on port 8000. The startup log showed:
```
[INFO] Minimal OAuth server is already running
✅ Ready for MCP connections
```
…with no OAuth ever happening.

## Fix

Return `False` immediately from `is_actually_running()` when `is_running` is `False` and `server_thread` is `None` — meaning `start()` has never been called on this instance. We could not have started the listener, so any TCP responder on that port is foreign. The existing TCP probe is preserved for the case where we *have* previously started the server and want to check if it's still alive.

After this fix, port conflicts are correctly detected and reported:
```
[ERROR] Port 8000 is already in use on localhost. Cannot start minimal OAuth server.
⚠️  Warning: Failed to start OAuth callback server: Port 8000 is already in use...
```

Users can resolve conflicts by setting `WORKSPACE_MCP_PORT` to a free port.

## Documentation

Updated the `WORKSPACE_MCP_PORT` entry in the Environment Variable Reference to note that it also controls the stdio OAuth callback port and that `PORT` takes precedence over it. Added a note in the Security section explaining how to resolve port conflicts, including guidance that web/confidential OAuth clients must also update the Google Cloud Console redirect URI to match the new port.

## Type of Change
- [x] Documentation update

## Testing
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

Manual testing:
- Verified port conflict is detected and reported (not silently swallowed) with a foreign server on port 8000
- Verified full OAuth flow completes successfully with `WORKSPACE_MCP_PORT=8181` on a clean port, using Claude Code as the MCP client

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Clarified OAuth callback setup, how to move the callback to a free port to avoid dev conflicts, and when redirect URI updates are required for web/confidential vs. desktop/PKCE clients.
  * Explained environment variable precedence for selecting the server/callback port.

* **Bug Fixes**
  * Improved the local OAuth callback server so it only reports “running” after it has actually started, avoiding false positives.


taylorwilsdon:
@coderabbitai re-review the updated branch please

taylorwilsdon:
Thanks!

rpeck:
@taylorwilsdon Yer welcome!

PR #720, opened 2026-04-23
feat: add GCSCredentialStore backend with optional CMEK enforcement

andyroyle:
## Summary

- Adds a new `GCSCredentialStore` backend, selectable via `WORKSPACE_MCP_CREDENTIAL_STORE_BACKEND=gcs`, which stores each user's credentials as an object in a Google Cloud Storage bucket via the JSON API
- Supports atomic read-modify-write via generation preconditions (`ifGenerationMatch`), preventing lost updates when two workers rotate the same user's refresh token concurrently
- Adds optional `WORKSPACE_MCP_GCS_REQUIRE_CMEK=true` flag that verifies the bucket has a default KMS key configured at startup and fails fast otherwise
- Preserves `LocalDirectoryCredentialStore` as the default; new backend is strictly opt-in

## Motivation

For remote MCP deployments on GCP (Cloud Run and similar), storing credentials in a bucket gives:

- Durable storage that survives container restarts without a mounted filesystem
- Native integration with Cloud IAM, Cloud Audit Logs, and VPC-SC
- Transparent bucket-level CMEK encryption at rest — no application-level key to manage or rotate
- Atomic RMW semantics that gcsfuse cannot provide

## Design notes

- `list_users()` raises `NotImplementedError` — this backend is designed for multi-user OAuth 2.1 mode where users are looked up individually by email; enumerating a bucket is semantically wrong for that flow (and expensive on large buckets). Single-user mode users should stay on `LocalDirectoryCredentialStore`.
- `google-cloud-storage` is added as a default dependency rather than an optional extra. The primary deployment target for a GCS-backed store is the project's Docker image; requiring users to rebuild with an extra would be friction for the main intended use case. Happy to move to an extra if you'd prefer.
- Uses Application Default Credentials — on Cloud Run this means the runtime service account needs `roles/storage.objectUser` (or equivalent) on the bucket.

## Test plan

- [x] 25 new unit tests with mocked \`google.cloud.storage.Client\` cover: round-trip get/store/delete, atomic RMW using current generation, retry on \`PreconditionFailed\`, give-up after persistent contention, CMEK verification pass/fail (including empty-string KMS key edge case), bucket + prefix config via arg or env var, path sanitisation of user emails, and backend selection via env var
- [x] All 10 existing credential-store security tests unchanged and passing
- [x] \`ruff check\` clean

## Summary by CodeRabbit

* **New Features**
  * Added a Google Cloud Storage credential backend selectable via environment configuration, with optional CMEK enforcement; GCS backend does not support user listing.

* **Documentation**
  * Expanded credential store docs: backend selection, local vs GCS behaviors, configuration examples, and CMEK guidance; example clarifies listing is local-only.

* **Tests**
  * Added comprehensive tests covering the GCS backend and environment parsing.

* **Chores**
  * Added Google Cloud Storage client dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1471,11 +1471,11 @@ export WORKSPACE_MCP_GCS_REQUIRE_CMEK="true"                # optional; see belo
 
 By default GCS encrypts objects with Google-managed keys. For customer-managed encryption, set a default KMS key on the bucket (e.g. via Terraform's `google_storage_bucket.encryption.default_kms_key_name`). All credentials written to the bucket will inherit the key transparently — no application-level key to manage.
 
-To guard against accidentally deploying against a bucket without CMEK, set `WORKSPACE_MCP_GCS_REQUIRE_CMEK=true`. The store will verify the bucket has a default KMS key at startup and refuse to initialise otherwise.
+To guard against accidentally deploying against a bucket without CMEK, set `WORKSPACE_MCP_GCS_REQUIRE_CMEK=true`. The store will verify the bucket has a default KMS key at startup and refuse to initialise otherwise. Note that this check reads bucket metadata, so the runtime service account additionally needs `storage.buckets.get` (provided by `roles/storage.legacyBucketReader` or `roles/storage.admin`) — `roles/storage.objectUser` alone only covers object operations.
 
 **Usage Example:**
 ```python
-from auth.credential_store import get_credential_store
+from auth.credential_store import get_credential_store, LocalDirectoryCredentialStore
 
 # Get the global credential store instance
 store = get_credential_store()
@@ -1486,8 +1486,11 @@ store.store_credential("user@example.com", credentials)
 # Retrieve credentials
 creds = store.get_credential("user@example.com")
 
-# List all users with stored credentials
-users = store.list_users()
+# List all users with stored credentials (local_directory backend only;
+# GCSCredentialStore intentionally does not support enumeration — use the
+# upstream identity provider to enumerate users instead).
+if isinstance(store, LocalDirectoryCredentialStore):
+    users = store.list_users()
 ```
 
 The credential store automatically handles credential serialization, expiry parsing, and provides error handling for storage operations.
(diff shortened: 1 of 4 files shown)


andyroyle:
Thanks for the second pass! Pushed 31efb47 addressing the major points. Two items I deliberately didn't adopt, with reasoning:

**1. "Reject GCS outside OAuth 2.1 multi-user mode" in get_credential_store**

The suggested diff has \`get_credential_store\` read \`MCP_ENABLE_OAUTH21\` and refuse to instantiate \`GCSCredentialStore\` when it's off. I think this creates inappropriate module coupling: the credential-store layer shouldn't know about OAuth-mode config, which lives elsewhere entirely. The backend is designed for multi-user flows, documented as such (class docstring + README), and \`list_users()\` raises \`NotImplementedError\` with a clear message pointing at single-user fallbacks. If an operator mis-configures, they get a loud error at the one callsite where it matters, rather than a startup-time check that reaches across concerns. Happy to reconsider if you see it differently.

**2. \`_blob_name\` lossy sanitisation (e.g. \`user+admin\` vs \`user_admin\` collision)**

Valid concern — but \`LocalDirectoryCredentialStore._get_credential_path\` uses the exact same regex (\`[^a-zA-Z0-9@._-]\` → \`_\`). Fixing it only in \`GCSCredentialStore\` would introduce divergence between the two backends; fixing it in both would change long-standing filename semantics for existing local installations. Seems like it belongs in a separate PR that tackles both stores consistently, if at all.

taylorwilsdon:
You don't have maintainer edits on, not sure why the precommit missed that - 
```
➜  google_workspace_mcp git:(feat/gcs-credential-store) ✗ git push
remote: Permission to veedstudio/google_workspace_mcp.git denied to taylorwilsdon.
fatal: unable to access 'https://github.com/veedstudio/google_workspace_mcp.git/': The requested URL returned error: 403
```

PR #723, opened 2026-04-23
feat(gmail): add return_base64 option to get_gmail_attachment_content

DannyOosterveer:
## Summary

Adds an opt-in `return_base64: bool = False` parameter to `get_gmail_attachment_content` so that sandboxed MCP clients can retrieve attachment bytes directly in the tool response, without needing to reach a localhost download URL or read from the MCP server's filesystem.

Default stays `False` — no change in response shape or size for existing callers.

## The problem

`get_gmail_attachment_content` currently returns either:

- a file path on disk (stdio mode), or
- a localhost download URL, e.g. `http://localhost:8000/attachments/<uuid>` (HTTP mode).

Both assume the MCP client can reach the server's host — either the same filesystem, or the same loopback network namespace.

That breaks for containerized / sandboxed clients. In my case, the client (another Claude-based agent) runs in an isolated Linux container with:
- no access to the host's filesystem, and
- a network allowlist that blocks `localhost` and `host.docker.internal` as deliberate egress isolation (not a bug, a policy).

For that class of client the attachment bytes are unreachable even though the MCP server successfully downloaded them — they exist in `base64_data` in memory and are then discarded after being written to a path / URL the client cannot use.

## The change

When `return_base64=True`, the response additionally contains a labeled block:

```
📦 Base64 content (N chars, standard base64):
<base64 string>
```

A small private helper `_format_base64_content_block` converts Gmail's URL-safe alphabet to standard base64 so the output can be passed straight to tools that use `base64.b64decode` (e.g. `draft_gmail_message` in this same codebase). That closes the loop: fetch attachment → inline base64 → forward/reply with the same attachment, all inside the MCP channel.

The path / download-URL output is unchanged. The base64 block is strictly additive.

## Why not just a config flag / transport-mode switch

- A global flag would force every attachment fetch to carry the base64, bloating responses for callers who don't need it.
- The sandboxed-vs-unsandboxed distinction is a property of the **client**, not the server, so it's the caller who should decide per-call.
- This follows the existing convention of per-call options (like `format="raw"` on message-fetch tools).

## Alternatives considered

- **Client-side workaround**: some sandboxed clients can shell out via a privileged side-channel (e.g. `curl -o /tmp/x` from a non-sandboxed tool). Works for some setups, but brittle and not portable across MCP hosts.
- **Shared filesystem mount**: requires cooperation from the MCP host runtime and doesn't help network-isolated clients.
- **Separate `get_gmail_attachment_base64` tool**: more surface area for the same capability; an opt-in flag keeps the tool count stable.

## Tests

New file `tests/gmail/test_get_gmail_attachment_content.py` covers:

- Helper converts URL-safe base64 → standard base64 (round-trips the full `bytes(range(256))` alphabet).
- Helper handles invalid base64 input without raising.
- Default call (`return_base64=False`) produces no base64 block — backwards compat.
- `return_base64=True` adds the block and the bytes round-trip via `base64.b64decode`.
- `return_base64=True` is additive: file is still saved and the normal path / URL is still returned.
- Returned base64 is directly usable by standard `base64.b64decode` (the alphabet the rest of the codebase expects).

All 793 existing tests still pass locally (`uv run pytest`).

## Test plan

- [x] Unit tests pass (`uv run pytest tests/gmail/`)
- [x] Full suite pass (`uv run pytest`)
- [ ] Manual: sandboxed MCP client fetches an attachment with `return_base64=True` and successfully drafts a forward containing it

## Summary by CodeRabbit

* **New Features**
  * Attachments can optionally include a standard Base64-encoded content block in responses via a new opt-in flag (off by default), while preserving existing output.

* **Bug Fixes**
  * Graceful handling of invalid/undecodable Base64 input by returning a clear warning line instead of failing.
  * Maintains existing download/save output alongside the optional Base64 block.

* **Tests**
  * Added tests for conversion, error handling, backward compatibility, and round‑trip decodability.


taylorwilsdon:
Great idea, appreciate this - will get it merged today!

taylorwilsdon:
@coderabbitai review

PR #733, opened 2026-04-27
Relax MCP protocol auth scope requirements

sf-jin-ku:
## Summary
- add a separate minimal protocol auth scope list for MCP bearer-token validation
- require only user identity scopes needed to accept the token (`userinfo.email` and `openid`)
- keep advertised/default valid scopes unchanged, so clients can still request `userinfo.profile` and all tool scopes

## Why
A token minted for a limited Workspace grant, for example `openid userinfo.email calendar.readonly`, can be fresh and otherwise valid but still fail `/api/mcp` validation if `GoogleProvider.required_scopes` also requires `userinfo.profile`. That turns a missing optional identity profile grant into a generic `401 invalid_token`.

## Tests
- `uv run ruff check .`
- `uv run pytest tests/core/test_http_oauth_scope_config.py tests/test_scopes.py -q` (`35 passed, 1 warning`)
- Full-suite note: `uv run pytest -q` currently stops on missing optional GCS dependency; `uv run --extra gcs pytest -q` runs the suite but reports two unrelated failures in `tests/auth/test_oauth_callback_server.py` and `tests/gdocs/test_insert_doc_tab_response.py` outside this patch area.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced OAuth authentication scope configuration to improve protocol-layer bearer token acceptance. Updated scope definitions provide better control over authentication request handling and security while maintaining full compatibility with existing authentication workflows.

* **Tests**
  * Updated authentication configuration tests to validate scope requirements and ensure correct configuration behavior.

### DIFF
--- a/auth/scopes.py
+++ b/auth/scopes.py
@@ -131,6 +131,9 @@ def has_required_scopes(available_scopes, required_scopes):
 # Base OAuth scopes required for user identification
 BASE_SCOPES = [USERINFO_EMAIL_SCOPE, USERINFO_PROFILE_SCOPE, OPENID_SCOPE]
 
+# Minimal scopes required to accept an MCP bearer token at the protocol layer.
+PROTOCOL_AUTH_SCOPES = [USERINFO_EMAIL_SCOPE, OPENID_SCOPE]
+
 # Service-specific scope groups
 DOCS_SCOPES = [
     DOCS_READONLY_SCOPE,
--- a/core/server.py
+++ b/core/server.py
@@ -25,7 +25,7 @@ from auth.oauth_responses import (
     create_server_error_response,
 )
 from auth.auth_info_middleware import AuthInfoMiddleware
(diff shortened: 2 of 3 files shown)


sf-jin-ku:
@taylorwilsdon tagging you for a quick sanity check because git history points to the two related scope changes:

- `2836f6c` added `userinfo.profile` to `BASE_SCOPES`
- `392ceee` changed `/api/mcp`'s `GoogleProvider.required_scopes` to use `BASE_SCOPES`

This PR keeps `userinfo.profile` in the advertised/default valid scopes, but stops requiring it for protocol-layer MCP bearer-token acceptance. Could you confirm whether any current OAuth/profile assumption depends on `userinfo.profile` being mandatory at the `/api/mcp` bearer-token validation layer?

taylorwilsdon:
lgtm, thanks!

PR #737, opened 2026-04-28
Issues/673

### LINKED ISSUE #673: feat(chat): Resolve DM participant names in list_spaces via People API
### Problem
`list_spaces` returns "Unnamed Space" for all DIRECT_MESSAGE spaces because the Google Chat API doesn't include `displayName` for DMs. This makes it impossible to identify who a DM conversation is with without reading messages from each space.

### Proposed Solution
Enrich `list_spaces` output for DIRECT_MESSAGE spaces by:
1. Calling `spaces.members.list` for each DM space
2. Filtering out the authenticated user
3. Resolving the other member's display name via People API (pattern already exists in PR #482 for message sender resolution)

The result would show e.g. `Joe Blue (DM)` instead of `Unnamed Space`.

### Additional Context
- PR #482 already added People API sender resolution for `get_messages` and `search_messages`, so the pattern and auth scopes are already in the codebase
- May need `chat.memberships.readonly` scope added
- Could be opt-in via a parameter (e.g. `resolve_dm_names=True`) to avoid extra API calls when not needed
- Performance consideration: batch People API lookups rather than one per DM space

### Environment
- Version: v1.18.0 (deployed via Docker on Railway)
- Transport: streamable-http with OAuth 2.1
- Tool tier: extended

taylorwilsdon:
Closes: #673 

## Summary by CodeRabbit

* **New Features**
  * Google Chat space listing now supports resolving direct message participant names
  * OAuth authentication flow enhanced to support dynamic scope expansion

* **Documentation**
  * Updated guidance on handling OAuth scope changes and cached credentials
  * Chat integration documentation updated to document direct message resolution capability


OAburub:
when is this going to be added? it's currently impossible to automatically find and text people based on their name, you have to provide the google chat URL.

Mardl:
Yes, please—it would make things much easier to just write a message directly without having to create long "spaceid = name" Markdown tags first.

PR #746, opened 2026-05-02
Security: prevent fork-PR code execution with write token in ruff workflow (CWE-77)

sebastiondev:
## Summary

The `ruff.yml` workflow contains a GitHub Actions "pwn request" pattern that allows any user who can open a pull request from a fork to obtain code execution inside a runner that holds a write-scoped `GITHUB_TOKEN` for this repository.

- **CWE-77**: Improper Neutralization of Special Elements used in a Command
- **File**: `.github/workflows/ruff.yml`
- **Trigger**: `pull_request` from a forked repository
- **Severity**: High — leads to repository write access (push to `main`, tag releases, modify other workflows) and exfiltration of `GITHUB_TOKEN`.

## Vulnerability details

The workflow currently:

1. Triggers on `pull_request` (which fires for fork PRs against `main`).
2. Grants `permissions: contents: write` at the workflow level.
3. Checks out `repository: ${{ github.event.pull_request.head.repo.full_name }}` with `token: ${{ secrets.GITHUB_TOKEN }}` — i.e., it pulls the attacker's fork into a job that holds the write token.
4. Runs `uv sync` against the fork's `pyproject.toml`.

`uv sync` resolves the project's build backend from `[build-system]` and `[tool.uv.sources]`. A malicious fork can declare an arbitrary build backend (or a `tool.uv.sources` entry pointing at a hostile package) whose install hooks execute arbitrary Python during dependency resolution. That code runs in the same job step that holds a `contents: write` `GITHUB_TOKEN`, so the attacker can:

- `git push` directly to `main` (or any branch),
- create/move tags and trigger the release/publish workflows,
- read the token from the environment and exfiltrate it,
- modify other workflow files via the API.

The only precondition is the ability to open a PR from a fork, which is available to any GitHub user.

## Fix

Split the workflow into two jobs with clearly separated trust boundaries:

- **`ruff` (validation)** — runs for every PR, including forks. Uses the default checkout (base ref, no token), `permissions: contents: read`, and only runs `ruff check` / `ruff format --check`. Even though `uv sync` still resolves the project's build system, the job has no write capability and no privileged secrets, so the impact of any code that runs is limited to the ephemeral runner.
- **`autofix` (write)** — gated by `if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository`. It only runs for PRs originating from branches in this repository (i.e., from contributors who already have push access). Only this job is granted `contents: write` and uses `GITHUB_TOKEN` to commit auto-fixes.

The net effect: untrusted fork code is never executed in a context that holds a write token.

## Testing

- Added `tests/test_cwe77_ruff_workflow.py` which parses `ruff.yml` and asserts the security invariants:
  - Workflow-level permissions are `contents: read`.
  - The `autofix` job is gated on `head.repo.full_name == github.repository`.
  - No checkout step uses `head.repo.full_name` outside the gated job.
  - The validation job does not request `contents: write`.
- YAML parses cleanly.
- Manually traced both code paths: fork PR → only `ruff` job runs (read-only); same-repo PR → both jobs run (auto-fix as before).

## Security analysis

The pre-fix configuration is exploitable end-to-end with no special access — opening a PR from a fork is enough. The fix removes the only path that combined "fork-controlled code execution" with "write-scoped token", which is the necessary condition for the pwn-request class of attack.

A residual note for the maintainer: the new `ruff` validation job still runs `uv sync` against fork-supplied `pyproject.toml`, so a malicious PR can still execute Python inside that runner. With `contents: read` and no other secrets in scope, the blast radius is limited to the ephemeral runner and the public read-only token. If you'd like to harden further, options are (a) skip `uv sync` for fork PRs and run `ruff` from a pinned standalone install, or (b) move to `pull_request_target` with `ref` pinned to the base commit. The current fix is intentionally minimal and addresses the immediate high-severity issue.

## Adversarial review

Before submitting, we tried to disprove this. We checked whether GitHub's default fork-PR protections (the "require approval to run workflows from first-time contributors" setting) would block exploitation — they don't, because that setting only adds a one-click approval for the *first* PR from a new contributor and doesn't downgrade the token or sandbox the checkout. We also reviewed `check-maintainer-edits.yml` and the other workflows (`docker-publish.yml`, `publish-mcp-registry.yml`) to see if any prior mitigation existed or if the pattern was repeated; the former is unrelated (comment-only) and the latter trigger on `push`/`release` rather than fork PRs, so they're not affected. The vulnerability is real and the fix is targeted.

cc @lewiswigmore

## Summary by CodeRabbit

* **Tests**
  * Expanded CI security tests with a strict same-repo guard matcher, added checks that install steps run only for same-repo PRs, and added a test ensuring the linter validation runs on push to main; test runner now fails fast on errors.

* **Chores**
  * Split CI into separate validation and autofix jobs with reduced top-level permissions; autofix runs only on same-repo PRs and only commits when changes exist.

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -10,23 +10,6 @@ permissions:
   contents: read
 
 jobs:
-  ruff:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v6
-    - uses: actions/setup-python@v6
-      with:
-        python-version: '3.11'
-    - name: Install uv
-      uses: astral-sh/setup-uv@v7
-    - name: Validate
-      # Use `uvx` so we run ruff in an isolated environment without invoking
-      # the project's build backend or executing pyproject.toml hooks. This
-      # keeps the validation step safe to run on untrusted fork PR code.
-      run: |
-        uvx ruff check
-        uvx ruff format --check
-
   autofix:
     if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
@@ -56,3 +39,23 @@ jobs:
         git add -A
         git commit -m "style: auto-fix ruff lint and format"
         git push
+
+  ruff:
+    if: always() && github.event_name == 'pull_request'
+    needs: [autofix]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-python@v6
+      with:
+        python-version: '3.11'
+    - name: Install uv
+      uses: astral-sh/setup-uv@v7
+    - name: Validate
+      # Use `uvx` so we run ruff in an isolated environment without invoking
+      # the project's build backend or executing pyproject.toml hooks. This
+      # keeps the validation step safe to run on untrusted fork PR code.
+      # For same-repo PRs, this runs after autofix commits any formatting fixes.
+      run: |
+        uvx ruff check
+        uvx ruff format --check
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -11,7 +11,7 @@ permissions:
 
 jobs:
   autofix:
-    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     permissions:
       contents: write
@@ -41,7 +41,7 @@ jobs:
         git push
 
   ruff:
-    if: always() && github.event_name == 'pull_request'
+    if: always() && (github.event_name == 'pull_request' || github.event_name == 'push')
     needs: [autofix]
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -19,12 +19,13 @@ jobs:
         python-version: '3.11'
     - name: Install uv
       uses: astral-sh/setup-uv@v7
-    - name: Install dependencies
-      run: uv sync
     - name: Validate
+      # Use `uvx` so we run ruff in an isolated environment without invoking
+      # the project's build backend or executing pyproject.toml hooks. This
+      # keeps the validation step safe to run on untrusted fork PR code.
       run: |
-        uv run ruff check
-        uv run ruff format --check
+        uvx ruff check
+        uvx ruff format --check
 
   autofix:
     if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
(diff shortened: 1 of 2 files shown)


sebastiondev:
Thanks @coderabbitai — all four points addressed in dd603fd.

**1. `uv sync` in fork-facing job (ruff.yml:22-27)** — agreed, this was the remaining attack surface. Replaced `uv sync` with `uvx ruff check` / `uvx ruff format --check` in the `ruff` job so we never resolve the project's build backend or run pyproject.toml hooks against fork code. The `autofix` job (which is already gated by a same-repo `if`) was switched to `uvx` as well for consistency.

**2. Missing mypy-strict type hints (test_cwe77_ruff_workflow.py:25-138)** — added `from __future__ import annotations`, annotated `load_workflow(path: str) -> Tuple[Dict[str, Any], str]`, gave each test `-> None`, and typed the local helpers/variables (`wf`, `raw`, `jobs`, `perms`, etc.).

**3. `test_uv_sync_not_on_fork_prs` no longer dependent on `checkout_uses_fork` (line 51-80)** — rewritten to iterate every step's `run` block, look for any project-aware installer (`uv sync`, `pip install .`, `pip install -e`, `poetry install`, …) and require a fork guard on either the step's `if` **or** the enclosing job's `if`. The assertion now fires regardless of the checkout shape.

**4. Job-level write permissions (line 83-111)** — `test_no_write_permissions_or_fork_guarded` now folds in `job.get('permissions')` per job, so the `autofix` job (`contents: write`) is exercised. It also asserts the privileged job carries a same-repo `if` guard, not just a safe checkout.

All three tests pass locally:
```
tests/test_cwe77_ruff_workflow.py::test_no_fork_repo_checkout PASSED
tests/test_cwe77_ruff_workflow.py::test_uv_sync_not_on_fork_prs PASSED
tests/test_cwe77_ruff_workflow.py::test_no_write_permissions_or_fork_guarded PASSED
```

taylorwilsdon:
Ah good call, thanks! Also, some good back and forth with coderabbit there :D

taylorwilsdon:
Looks like it doesn't actually work, it's blocking - let me see what's what.

sebastiondev:
@taylorwilsdon Yep, the failure was just `ruff format --check` flagging unescaped quotes in f-strings inside the new test file (a side-effect of running formatter against fork PRs now that we dropped `uv sync`). Reformatted in 614b0cf — `ruff check` and `ruff format --check` both pass locally on the test file now.

Note: `gmail/gmail_helpers.py` was also flagged by the CI run but that file isn't touched by this PR (it doesn't exist on this branch — looks like the CI was checking against a stale tree that included main's pre-existing formatting drift). The ruff job should now go green for this PR's diff.

sebastiondev:
Appreciate the review. @lewiswigmore

PR #748, opened 2026-05-02
fix: limited tools + cimd

### LINKED ISSUE #668: CIMD clients (Claude Code) get invalid_scope error with limited TOOLS config

taylorwilsdon:
Closes #668 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced OAuth 2.1 default scope configuration to properly align with valid provider scopes, ensuring consistent scope configuration across authentication components.

### DIFF
--- a/core/server.py
+++ b/core/server.py
@@ -524,6 +524,13 @@ def configure_server_for_http():
                     provider.client_registration_options.default_scopes = (
                         provider_valid_scopes
                     )
+                # CIMD clients can bypass DCR defaults and fall back to FastMCP's
+                # internal scope string, so keep it aligned with valid scopes too.
+                cimd_default_scope = " ".join(provider_valid_scopes)
+                provider._default_scope_str = cimd_default_scope
+                cimd_manager = getattr(provider, "_cimd_manager", None)
+                if cimd_manager is not None:
+                    cimd_manager.default_scope = cimd_default_scope
                 # Enable protocol-level auth
                 server.auth = provider
                 logger.info(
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
Upstream issue opened at https://github.com/PrefectHQ/fastmcp/issues/4090 and PR at https://github.com/PrefectHQ/fastmcp/pull/4091 so we can revert this when fixed upstream

PR #751, opened 2026-05-03
feat(gdrive): add list_shared_drives tool with optional organizers

lacotousek:
## Motivation

`workspace-mcp` currently exposes `list_drive_items` and `search_drive_files`
for browsing Drive contents, but there is no wrapper around `Drives.list()`
– the user-level Shared Drive enumeration. Without it, a calling agent
cannot discover which Shared Drives the authenticated user has access to
unless the IDs are known upfront.

## Feature

Adds a `list_shared_drives` tool that wraps `drive.drives().list()` and
returns each drive's `id`, `name`, `createdTime`, `hidden` flag, plus
selected `capabilities` and `restrictions`. Supports paging (`page_size`
/ `page_token`) and an optional Drive query filter (`query`).

Includes an opt-in `include_organizers=True` flag (default `False`). When
enabled, performs an additional `permissions.list` call per drive scoped
by `corpora='drive'` and returns the principals with the `organizer` role
alongside the drive metadata. Default off keeps the cheap listing path
zero-cost; opt-in is documented in the docstring as the closest equivalent
to a per-drive "owner" (since shared drives are owned by the workspace
organization rather than an individual user).

## Test

Verified on a Workspace organization with one Shared Drive: returns the
drive correctly with all metadata fields populated. Organizer fetch
correctly returns the four organizers with their email addresses, display
names, and `type` (`user` / `group`). Whitespace, paging and the Drive
query filter were exercised against the same drive.

## Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Notes

Tool is registered in the `complete` tier (read-heavy enumeration,
low-frequency use). Open to feedback on tier placement, parameter names,
or output schema. Conceptually pairs with #750 (shared-drive lookup fix)
but the changes are independent and can be merged in any order.

## Summary by CodeRabbit

* **New Features**
  * Drive tools can list and query shared drives
  * File public-access checks can be scoped to a specific shared drive
  * Option to include organizer information when listing shared drives

* **Bug Fixes**
  * Unsupported resource types are now validated and rejected early

* **Documentation**
  * Updated tool docs to describe shared-drive capabilities

* **Tests**
  * Added tests for shared-drive listing, organizer inclusion, and invalid-input handling

### DIFF
--- a/README.md
+++ b/README.md
@@ -764,14 +764,13 @@ cp .env.oauth21 .env
 | <sub>`create_drive_folder`</sub> | <sub>Core</sub> | <sub>Create empty folders in Drive or shared drives</sub> |
 | <sub>`import_to_google_doc`</sub> | <sub>Core</sub> | <sub>Import files (MD, DOCX, HTML, etc.) as Google Docs</sub> |
 | <sub>`get_drive_shareable_link`</sub> | <sub>Core</sub> | <sub>Get shareable links for a file</sub> |
-| <sub>`list_drive_items`</sub> | <sub>Extended</sub> | <sub>List folder contents</sub> |
+| <sub>`list_drive_items`</sub> | <sub>Extended</sub> | <sub>List folder contents or shared drives</sub> |
 | <sub>`copy_drive_file`</sub> | <sub>Extended</sub> | <sub>Copy existing files (templates) with optional renaming</sub> |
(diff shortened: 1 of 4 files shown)


taylorwilsdon:
Going to roll this into list_drive_items rather than exposing a net new tool, we're already maxed out on what openai models can work with on full coverage toolset. Appreciate it, great idea!

PR #756, opened 2026-05-04
feat(auth): add header token passthrough mode

guifran001:
## Description
Fix the external OAuth 2.1 provider not accepting a JWT signing key.

This allows running the MCP server without sharing the Google OAuth
client secret, by providing a JWT signing key instead.

**Configuration:**
- `MCP_ENABLE_OAUTH21=true`
- `EXTERNAL_OAUTH21_PROVIDER=true`
- `GOOGLE_OAUTH_CLIENT_ID=xyz.apps.googleusercontent.com`
- `FASTMCP_SERVER_AUTH_GOOGLE_JWT_SIGNING_KEY=averysecretvalue`

With this setup, the access token passed in the `Authorization: Bearer`
header is validated against Google's userinfo API to resolve the user's
email, and then forwarded to the Google APIs directly — no client secret
required on the server.

**Root cause:** `configure_server_for_http()` derived the JWT signing key
from `FASTMCP_SERVER_AUTH_GOOGLE_JWT_SIGNING_KEY` but never forwarded it
to `ExternalOAuthProvider`, causing a startup failure when `client_secret`
was absent.

**Change:** One line — pass `jwt_signing_key=jwt_signing_key` to the
`ExternalOAuthProvider` constructor.

## Type of Change
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Testing
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I have tested this change manually

## Checklist
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] **I have enabled "Allow edits from maintainers" for this pull request**

## Additional Notes
N/A

---

**⚠️ IMPORTANT:** This repository requires that you enable "Allow edits from maintainers" when creating your pull request. This allows maintainers to make small fixes and improvements directly to your branch, speeding up the review process.

To enable this setting:
1. When creating the PR, check the "Allow edits from maintainers" checkbox
2. If you've already created the PR, you can enable this in the PR sidebar under "Allow edits from maintainers"

## Summary by CodeRabbit

* **Improvements**
  * External OAuth 2.1 provider initialization now receives the derived JWT signing key to ensure correct JWT handling during authentication.

* **Tests**
  * Added a regression test that verifies the JWT signing key is passed to external OAuth 2.1 providers to prevent future regressions.


taylorwilsdon:
I'm open to this, I know folks have asked for something similar in the past. I think the biggest "risk" so to speak is that unlike external OAuth mode, which handles bearer tokens from incoming requests too, this doesn't validate those tokens so users attempting to hit it without the necessary scopes for whatever they're doing will get an opaque 403 without the nice automatic handling with the rest of the modes. Since it's an advanced option, presumably the people leveraging it are capable of understanding that kind of limitation but still worth highlighting imo.

Biggest callout - it appears that the PR has it backwards. It raises an error if service is started in oauth 2.1 mode, but this can only work in oauth 2.1 mode. Legacy mode (which is what will kick on if not in oauth 2.1 mode) isn't capable of doing any of what you're asking for and will present user_google_email as a param + cache credentials. 

Like I said, I'm open to merging this functionality but it would need to be totally rewritten and will need (extensive) documentation. To merge, I'd say we need to update so that header token mode requires MCP_ENABLE_OAUTH21=true, which will let it inherit OAuth 2.1's email resolution and signature behavior and drop the user_google_email parameter path entirely.

guifran001:
Hi,

Thank you for your quick answer.

You're right on all counts.

I overcomplicated this. While debugging the rewrite you suggested, I traced the actual failure: ExternalOAuthProvider was never receiving the jwt_signing_key that had already been derived earlier in `configure_server_for_http()`, so startup failed whenever `client_secret` was absent.

The fix turned out to be a single line — forwarding jwt_signing_key to ExternalOAuthProvider. With that in place, the existing EXTERNAL_OAUTH21_PROVIDER=true mode already does everything I originally wanted:

GOOGLE_OAUTH_CLIENT_ID + FASTMCP_SERVER_AUTH_GOOGLE_JWT_SIGNING_KEY (no client_secret needed)
Token validation via the userinfo API
Email resolved from the token, user_google_email hidden from signatures
Opaque 403s on scope failures:  agreed that's the expected tradeoff for an advanced mode, and it's already documented behavior
I've replaced the entire PR with that one-line fix. No new env vars, no new documentation burden.

taylorwilsdon:
Amazing, appreciate the followup @guifran001. Will get a good eye on this and get it in.

taylorwilsdon:
quick change otherwise lgtm

PR #761, opened 2026-05-05
feat(permissions): add drive:file level for per-file Drive write scope

Jonboyweb:
## Description

Adds a third Drive permission level, `file`, that grants only `drive.file` (cumulative with `drive.readonly`) without the broader `drive` scope.

Today, `--permissions drive:full` bundles `DRIVE_FILE_SCOPE` with `DRIVE_SCOPE` (full account write). Callers who want to write into a specific folder using `drive.file` semantics — i.e. only see/touch files this app creates — have no way to request that scope alone. The choice is `readonly` or full account write.

This PR fills the gap:

```python
"drive": [
    ("readonly", [DRIVE_READONLY_SCOPE]),
    ("file",     [DRIVE_FILE_SCOPE]),    # NEW — between readonly and full
    ("full",     [DRIVE_SCOPE]),
],
```

Cumulative semantics mean `drive:full` continues to grant the same effective scope set as before (`drive.readonly` + `drive.file` + `drive`), so existing `drive:full` callers see no behaviour change. Only the layout of which scope is added at which level changes.

## Use case

The motivating use case is an agent that creates artefacts (generated reports, output files) into a designated folder on the user's Drive. With `drive:full` it can do this — but it also receives full account write, which is a much larger blast radius than required. With the new `drive:file` level, the agent gets exactly the scope it needs: it can create files anywhere the authenticated user has write access, and read/modify only the files it itself created. Pre-existing files in the same folder remain invisible to it at the API layer.

The `drive.file` scope description on the Google consent screen ("See, edit, create, and delete only the specific Google Drive files you use with this app.") is exactly what end users expect for this kind of integration, and `drive:full` has been over-permissive for them.

## Type of Change

## Testing

- [x] New tests added in `tests/test_permissions.py`:
  - `test_drive_file_is_valid_level` — `drive:file` is parseable.
  - `test_drive_file_includes_readonly_and_excludes_full` — file level grants `drive.file` + `drive.readonly` but not `drive`.
  - `test_drive_full_cumulatively_includes_file` — regression guard that `drive:full` still includes `drive.file` after the layout change.
- [x] Existing drive tests continue to pass unchanged. `test_drive_full_includes_readonly` still asserts both `DRIVE_READONLY_SCOPE` and `DRIVE_SCOPE` are present in `drive:full`'s effective scope set, which still holds with the new layout.
- [x] Full `tests/test_permissions.py` passes locally (33/33).

## Checklist

## Additional Notes

The module docstring and `--permissions` CLI help text are updated to mention the new level.

No changes to the older `--read-only` / `--tools` flag paths — those still use `TOOL_SCOPES_MAP` / `TOOL_READONLY_SCOPES_MAP` from `auth/scopes.py` and are unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added a finer-grained Drive "file" permission and a new "metadata" permission level to allow per-file access and metadata-only access without full Drive write.

* **Behavior**
  * Docs and Sheets write actions can now be limited to app-created files when using the file-level grant.

* **Documentation**
  * CLI help and in-app permission docs updated to describe the new levels and their cumulative scope behavior.

* **Tests**
  * Unit tests added to validate the new permission levels and their scope composition.

### DIFF
--- a/auth/permissions.py
+++ b/auth/permissions.py
@@ -9,7 +9,7 @@ Usage:
     --permissions gmail:organize drive:readonly
 
 Gmail levels: readonly, organize, drafts, send, full
-Drive levels: readonly, file, full
+Drive levels: readonly, file, metadata, full
 Tasks levels: readonly, manage, full
 Other services: readonly, full (extensible by adding entries to SERVICE_PERMISSION_LEVELS)
 
@@ -31,6 +31,7 @@ from auth.scopes import (
     GMAIL_SETTINGS_BASIC_SCOPE,
     DRIVE_READONLY_SCOPE,
     DRIVE_FILE_SCOPE,
+    DRIVE_METADATA_SCOPE,
     DRIVE_SCOPE,
     CALENDAR_READONLY_SCOPE,
     CALENDAR_EVENTS_SCOPE,
@@ -77,6 +78,11 @@ SERVICE_PERMISSION_LEVELS: Dict[str, List[Tuple[str, List[str]]]] = {
     "drive": [
         ("readonly", [DRIVE_READONLY_SCOPE]),
         ("file", [DRIVE_FILE_SCOPE]),
+        # "metadata" adds drive.metadata on top of "file": move / rename / reorganise
+        # ANY file the user can reach (including files this app did NOT create) without
+        # the broad drive scope — so still no content edit/delete of others' files.
+        # Cumulative, so this level grants drive.readonly + drive.file + drive.metadata.
+        ("metadata", [DRIVE_METADATA_SCOPE]),
         ("full", [DRIVE_SCOPE]),
     ],
     "calendar": [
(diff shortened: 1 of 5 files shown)


oliver-mee:
Production evidence for this, in case it helps move it along.

I have been running the equivalent scope narrowing on a self-hosted deployment since June, on a two-server setup serving about ten users. I arrived at the same layout independently before finding this PR:

```python
"drive": [
    ("readonly", [DRIVE_READONLY_SCOPE]),
    ("file",     [DRIVE_FILE_SCOPE]),
    ("full",     [DRIVE_SCOPE]),
],
```

Two things worth adding from having actually run it:

**Nothing broke in normal use.** Every Drive write tool works on `drive.file` alone. The one real gap is `manage_doc_comment` and `manage_sheet_comment`, which need the broad scope and fail without it. Comment READING still works, because that goes through `drive.readonly`. That is a fair trade at this level rather than a blocker, but it is worth documenting so people picking `drive:file` are not surprised.

**Cumulative semantics mean `drive:full` really is unchanged.** I was worried about that when I made the same change locally. It grants `drive.readonly` plus `drive.file` plus `drive`, exactly as before, so existing callers see no difference. Only the level at which each scope enters changes.

The motivation on my side is the same as yours: an agent writing generated artefacts into a Drive folder does not need see-and-delete-everything on the whole account, and until this lands the only options are read-only or full account write. Granting a scope an OAuth token then carries forever is not something you can walk back per-tool afterwards.

Anything I can do to help this along? I am happy to test a rebased version against a live deployment, or to take it over if the original author has moved on.

PR #765, opened 2026-05-06
fix: use sole credential user in single-user mode

### LINKED ISSUE #642: Single-user mode should override caller-provided user_google_email with credential file email
## Bug Description

In single-user mode (`--single-user`), `get_credentials()` in `google_auth.py` only overrides `user_google_email` with the email from the credential file when the caller did not provide one. If the caller provides a **different** email than what's in the credential file, the mismatch causes a session binding conflict and auth failure.

## Environment

- workspace-mcp v3.2.0
- Claude Desktop (Cowork mode)
- macOS, stdio transport
- 3 MCP server instances configured for 3 separate Google accounts

## Steps to Reproduce

1. Configure 3 Gmail MCP servers in `claude_desktop_config.json` with `--single-user` flag, each pointing to a different credentials directory with a single token file
2. Open **two** Cowork projects. Each project launches its own set of 3 MCP processes (6 total)
3. From Project B, ask Claude to search Gmail on a specific account. Claude passes a `user_google_email` that may differ from the credential file email (e.g. Claude uses the user's profile email instead of the account-specific email)
4. Auth fails with "re-authentication required"

## Root Cause

In `google_auth.py` around line 880:

```python
# Use the email from the credential file if not provided
if not user_google_email and found_user_email:
    user_google_email = found_user_email
```

The condition `if not user_google_email` means if the caller provides any email (even the wrong one), the credential file's email is ignored. Later, when the refreshed token is saved:

1. `save_credentials_to_session` uses the wrong email
2. `store_credential` writes a stale file with the wrong email name
3. `OAuth21SessionStore.store_session()` throws `ValueError: Session is already bound to a different user`

This also causes a case-sensitivity variant: if the credential file is `User@gmail.com.json` but the caller passes `user@gmail.com`, the same conflict occurs.

## Suggested Fix

In single-user mode, unconditionally override `user_google_email` with the email from the credential file:

```python
# In single-user mode, always use the credential file's email
if found_user_email:
    user_google_email = found_user_email
```

This is correct because single-user mode by definition has exactly one account per credentials directory. The server should trust its own files over what the caller passes.

## Debug Log Evidence

```
[get_credentials] Single-user mode: bypassing session mapping
[single-user] Found credentials for sameer@account1.com via credential store
[get_credentials] Credentials not valid. Attempting refresh. User: 'wrong@email.com'
[get_credentials] Credentials refreshed successfully.
Stored credentials for wrong@email.com <-- WRONG EMAIL
[get_credentials] Error refreshing credentials: Session is already bound to a different user
```

## Related Issues

- #372 (multi-account configuration)
- #310 (email address confusion in single-user mode)

zxyasfas:
## Description
Fixes single-user auth when the MCP client sends a stale or profile-derived `user_google_email` that does not match the local credential file. In single-user mode, the credential store is now allowed to use its sole stored account as the canonical user. If multiple credential users exist, the previous strict behavior is preserved to avoid selecting the wrong account.

Fixes #642

## Type of Change
- [ ] Documentation update

## Testing

Commands run:
- `uv run pytest tests\\auth\\test_google_auth_refresh_persistence.py`
- `uv run ruff check auth\\google_auth.py tests\\auth\\test_google_auth_refresh_persistence.py`

## Checklist

## Additional Notes
The fallback is intentionally limited to exactly one stored credential user, with a case-insensitive email match handled before the single-user fallback path.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable Google sign-in in single-user mode with case-insensitive email matching, improved fallback when the requested account is missing, and consistent canonical email propagation for sessions and persistence.

* **Tests**
  * Expanded coverage for single-user authentication flows, credential resolution and storage behaviors.

* **Documentation**
  * Updated authentication docs to reflect refined single-user decision rules.

[Review Change Stack](https://app.coderabbit.ai/change-stack/taylorwilsdon/google_workspace_mcp/pull/765)

### DIFF
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -133,6 +133,68 @@ def _find_any_credentials(
     return None, None
 
 
+def _find_single_user_credentials_for_request(
+    user_google_email: str,
+) -> tuple[Optional[Credentials], Optional[str]]:
+    """
+    Find credentials in single-user mode when the caller supplied an email.
+
+    A single-user deployment may receive a stale or client-profile email from
+    the MCP client. If the credential store contains exactly one account, trust
+    the local credential file over the caller-provided value. When multiple
+    users exist, keep the strict behavior to avoid selecting the wrong account.
+    """
+    credential_store = get_credential_store()
+    credentials = credential_store.get_credential(user_google_email)
+    if credentials:
+        logger.info(
+            f"[get_credentials] Single-user mode: found credentials for requested user {user_google_email}"
+        )
+        return credentials, user_google_email
+
+    try:
+        users = credential_store.list_users()
+    except Exception as e:
+        logger.info(
+            "[get_credentials] Single-user mode: no credentials for requested "
+            f"user {user_google_email}; cannot inspect local credential users: {e}"
+        )
+        return None, None
+
+    requested_lower = user_google_email.lower()
+    case_matches = [user for user in users if user.lower() == requested_lower]
+
+    if len(case_matches) == 1:
+        selected_user = case_matches[0]
+        logger.info(
+            "[get_credentials] Single-user mode: using credential store email "
+            f"{selected_user} for requested user {user_google_email}"
+        )
+    elif len(users) == 1:
+        selected_user = users[0]
+        logger.info(
+            "[get_credentials] Single-user mode: overriding requested user "
+            f"{user_google_email} with sole credential user {selected_user}"
+        )
+    else:
+        logger.info(
+            "[get_credentials] Single-user mode: no credentials for requested "
+            f"user {user_google_email}; not falling back because {len(users)} "
+            "credential users are available"
+        )
+        return None, None
+
+    credentials = credential_store.get_credential(selected_user)
+    if not credentials:
+        logger.info(
+            "[get_credentials] Single-user mode: selected credential user "
+            f"{selected_user} could not be loaded"
+        )
+        return None, None
+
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
Will this solve for ^ ? https://github.com/taylorwilsdon/google_workspace_mcp/pull/762

zxyasfas:
Added a follow-up commit for the credential-email propagation case.

The single-user fallback now keeps `get_credentials()` backward-compatible while exposing the resolved credential email to callers through an optional callback. `get_authenticated_google_service()` uses that resolved email before returning `(service, user_email)`, so downstream tools receive the canonical stored account instead of the stale/requested email when the sole local credential is selected.

This is complementary to #762: #762 recovers missing OAuth state in single-user stdio flows; this PR handles the next step where the MCP client may still pass a stale `user_google_email` after auth. Together they cover both the missing-state recovery path and the credential-email mismatch path.

Validation:
- `python -m py_compile auth/google_auth.py tests/auth/test_google_auth_refresh_persistence.py`

zxyasfas:
Addressed the latest review nits by adding explicit type annotations to the updated test helpers and test functions in `tests/auth/test_google_auth_refresh_persistence.py`.

Validation:
- `python -m py_compile auth/google_auth.py tests/auth/test_google_auth_refresh_persistence.py`
- `python -m ruff check tests/auth/test_google_auth_refresh_persistence.py`

Issue #754, opened 2026-05-04
OAuth callback "Port 8000 already in use" in stdio mode — boot-time MinimalOAuthServer collides with on-demand auth flow (v1.20.3, DXT install)


georgehomecloud:
I am not an engineer and this is just what Claude is telling me an how i was able to get it to work

## Environment
- Workspace MCP version: 1.20.3
- Install method: DXT extension installed in Claude Desktop (macOS)
- Transport: stdio (default)
- OAuth client: Web application (custom Google Cloud OAuth 2.0 client)
- macOS 14.x, Python 3.x (bundled .venv)

## Summary
In stdio mode, the MCP eagerly starts a `MinimalOAuthServer` at boot (per `main.py` lines ~723-728, `ensure_oauth_callback_available("stdio", port, base_uri)`). When an OAuth flow is later triggered by a tool call, the auth path calls `ensure_oauth_callback_available` again. The singleton check exists, but `_minimal_oauth_server.is_actually_running()` apparently returns False, so `start()` is invoked again. The new `start()` call attempts to bind the same port the boot-time singleton already holds, fails with `OSError`, and returns the error:

Cannot initiate OAuth flow - callback server unavailable (Port 8000 is already in use on localhost. Cannot start minimal OAuth server.)

`lsof -i :8000` confirms the running listener is the same Python process (the MCP itself).

## Steps to reproduce
1. Install Workspace MCP 1.20.3 as a DXT in Claude Desktop.
2. Configure with a Google Cloud Web OAuth client (Client ID + Secret), `WORKSPACE_MCP_PORT=8000`, `GOOGLE_OAUTH_REDIRECT_URI=http://localhost:8000/oauth2callback`, matching authorized redirect URI in Google Cloud.
3. Restart Claude Desktop.
4. Confirm with `lsof -i :8000` that the MCP is bound to port 8000.
5. Trigger any tool call that requires auth (e.g., `search_drive_files`).
6. Observe the error above instead of an auth URL.

## Expected behavior
The auth flow either reuses the running boot-time MinimalOAuthServer to handle the callback, or `start()` is idempotent in the case where the port is already bound by our own singleton.

## Workaround applied locally
In `auth/oauth_callback_server.py`, in `start()`, when the bind raises `OSError`, log a warning and return `True` instead of `False`, with `self.is_running = True`. This unblocks the flow because the boot-time listener already serves `/oauth2callback` correctly.

```python
except OSError:
    logger.warning(
[...]

Issue #667, opened 2026-04-08
Orphaned python instance holds port 8000, causing OAuth "invalid state" on every subsequent auth attempt


w3b5urf3r:
````markdown
**Describe the bug**
When Claude Desktop relaunches the workspace-mcp extension (app restart, extension reload, or crash recovery), the previous `python main.py` process is sometimes left alive while a new one is spawned. The orphan still holds the hard-coded OAuth callback port **8000**, while the new instance — which Claude Desktop is now talking to over stdio — fails to bind 8000 and silently continues without an HTTP server.

From this point on, every OAuth flow is broken:
- `start_google_auth` generates an OAuth `state` token and stores it in the **new** instance's in-memory state store.
- Google redirects the user to `http://localhost:8000/oauth2callback`.
- The callback is answered by the **orphan** instance, which has no record of that state.
- The orphan returns: `Authentication Processing Error: An unexpected error occurred while processing your authentication: Invalid or expired OAuth state parameter. Please try again. You can close this tab.`

Every retry hits the same wall because the orphan never dies and the new instance never gets to serve the callback. Previously authenticated accounts continue to work (their refresh tokens are on disk), but no new account can be added. The only workaround is manually `kill`-ing the orphan PID — and its `uv` + `disclaimer` parents — and re-running `start_google_auth` **without** quitting Claude Desktop (quitting wipes the surviving instance's in-memory state and restarts the cycle).

Suspected root cause: the `disclaimer → uv run → python main.py` parent chain doesn't reliably forward SIGTERM from Claude Desktop down to the python process, so when Claude Desktop reaps the parent the python child is reparented to launchd and continues running, holding port 8000.

**Startup Logs**
Not captured during this incident — Claude Desktop does not surface the extension's stdout/stderr to the user, and by the time the bug was diagnosed both processes had been running for several minutes. Happy to capture and attach a fresh Active Configuration → "Uvicorn running" block on a follow-up reproduction; please advise where Claude Desktop writes those logs on macOS so they can be grabbed next time.

**To Reproduce**
Steps to reproduce the behavior:
1. Install `local.dxt.taylor-wilsdon.workspace-mcp` as a Claude Desktop extension on macOS and complete OAuth for one Google account (account A) successfully.
[...]

PR #768, opened 2026-05-07
feat(auth): late-bind OAuth callback port with preferred + fallback range

theparlor:
# Late-bind OAuth callback port with preferred + fallback range

## Summary

In stdio mode the OAuth callback server port is hardcoded to `8000`. When a client (Claude Desktop, Claude Code, Cursor, etc.) spawns multiple `workspace-mcp` processes — a normal occurrence on app reload, extension restart, or multi-session usage — only one wins the port. The others either error out at boot (`Port 8000 already in use`) or, worse, silently continue without a callback listener and produce 403 invalid-state on every subsequent OAuth attempt because the state-token PID and the listener PID diverge.

This PR makes the callback port late-bound at process start: a preferred port plus a small fallback range is probed, the first available is bound, and `redirect_uri` is composed from the actual bound port.

## Fixes

- #546 — Multiple stdio processes cause OAuth callback server port conflict and state mismatch
- #667 — Orphaned python instance holds port 8000, causing OAuth "invalid state"
- #754 — OAuth callback "Port 8000 already in use" in stdio mode

## Approach (vs. existing PRs)

- **#590** (closed): Shared OAuth state across processes via filesystem. Works around the symptom; preserves the single-port assumption.
- **#606** (open): Better self-detection — recover when our own thread holds the port. Helps re-auth, doesn't help multi-process spawning.
- **This PR**: Eliminate the single-port assumption. Each process binds its own port from a small range; redirect URIs are composed from the actual bound port. No cross-process state sharing required.

The trade-off depends on the OAuth client type:

- **Desktop application clients** (the most common workspace-mcp setup, since stdio mode is a desktop-bound MCP transport) — no registration required. Google's loopback rule accepts any port on `localhost`/`127.0.0.1` for Desktop clients automatically. Verified directly in GCP Console: Desktop client config pages don't even surface a redirect URI field. Zero operational cost; the fix is fully automatic.

- **Web application clients** (less common for workspace-mcp, used when the same OAuth client is shared with a hosted streamable-http deployment) — each fallback port must be registered as an authorized redirect URI in the OAuth client config. One-time setup. The README documents the entries.

## Changes

| File                      | Change                                                                                                              |
| ------------------------- | ------------------------------------------------------------------------------------------------------------------- |
| `auth/port_resolver.py`   | New module. `resolve_port()` probes preferred + fallback range with connect-then-bind, mutates env to bound port.   |
| `main.py`                 | Calls `resolve_port()` after argparse, then `reload_oauth_config()` so `redirect_uri` reflects the bound port.      |
| `core/config.py`          | `WORKSPACE_MCP_PORT` is lazy-evaluated via PEP 562 `__getattr__` so the late-bound env value isn't shadowed.        |

`auth/oauth_callback_server.py`, `auth/oauth_config.py`, and `auth/google_auth.py` are unchanged — they consume `port`/`base_url`/`redirect_uri` through the existing API and pick up the late-bound port automatically.

## Probe semantics

`_is_port_free(host, port)` does a two-stage check:

1. **Connect probe** to `127.0.0.1:port`. Detects an existing listener regardless of which interface they bound to. On macOS, `SO_REUSEADDR=1` lets a `0.0.0.0` bind succeed even when `127.0.0.1` is already listening — so a bind probe alone gives a false-negative. Since the OAuth callback URI is always `http://localhost:port`, anything listening on loopback is a collision.
2. **Bind probe** to `(host, port)` with default `SO_REUSEADDR`. Catches `TIME_WAIT` and BSD-style local conflicts.

## Configuration

| Env var                              | Default | Meaning                              |
| ------------------------------------ | ------- | ------------------------------------ |
| `WORKSPACE_MCP_PORT`                 | `8000`  | Preferred port (existing semantics)  |
| `WORKSPACE_MCP_PORT_FALLBACK_COUNT`  | `4`     | Fallback slots (5 ports total)       |
| `WORKSPACE_MCP_HOST`                 | `0.0.0.0` | Bind host (existing)               |

Set `WORKSPACE_MCP_PORT_FALLBACK_COUNT=0` to restore the original single-port behavior.

## Test plan

- New stdlib `unittest` suite covers: preferred-when-free, fallback-when-held, raise-when-all-held, OAuthConfig reload picks up the bound port via `redirect_uri`, `core.config.WORKSPACE_MCP_PORT` lazy-evaluation.
- Smoke test: with port 8000 already held by another `workspace-mcp` process, a fresh instance correctly logs `falling back to 8001` and binds 8001 with redirect_uri `http://localhost:8001/oauth2callback`. Two simultaneous spawns share `WORKSPACE_MCP_PORT=8500` and one bound 8500, the other 8501 — no errors, no orphan listeners.

## Notes

- Idempotent for existing single-instance users — when port 8000 is free, behavior is unchanged.
- Desktop OAuth client users (the typical workspace-mcp config): no GCP changes needed; Google's loopback rule covers all fallback ports automatically.
- Web OAuth client users (rare for workspace-mcp): if fallback ports aren't registered, fallback-bound instances start cleanly but Google rejects auth flows with `redirect_uri_mismatch`. Set `WORKSPACE_MCP_PORT_FALLBACK_COUNT=0` to restore single-port behavior, or register the fallback URIs.
- The connect-probe assumes loopback is reachable. If you've configured the callback URI to bind to a non-loopback interface (rare), set `WORKSPACE_MCP_PORT_FALLBACK_COUNT=0` to keep the existing behavior.

## Summary by CodeRabbit

* **New Features**
  * Automatic OAuth callback port resolution: scans a preferred+fallback range, probes/reuses existing callback endpoints when safe, and updates runtime redirect URIs and resolved-port marker.

* **Bug Fixes**
  * Port value is now lazily read so it reflects current environment.
  * Startup reports clear, actionable errors when port config is invalid or no ports are available.
  * Transport-specific handling ensures only relevant transports trigger resolution.

* **Tests**
  * Added tests for port selection, reuse probing, lazy env resolution, and error cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/taylorwilsdon/google_workspace_mcp/pull/768?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/auth/oauth_callback_server.py
+++ b/auth/oauth_callback_server.py
@@ -18,6 +18,8 @@ from fastapi import FastAPI, Request
 from fastapi.responses import FileResponse, JSONResponse
 from typing import Optional
 from urllib.parse import urlparse
+from urllib.error import HTTPError, URLError
+from urllib.request import urlopen
 
 from auth.scopes import SCOPES, get_current_scopes  # noqa
 from auth.oauth_responses import (
@@ -198,6 +200,42 @@ class MinimalOAuthServer:
         other_path = other_parsed.path.rstrip("/") or ""
         return self_path == other_path
 
+    def _callback_endpoint_looks_like_workspace(self, hostname: str) -> bool:
+        """
+        Probe an occupied callback port for this app's OAuth callback handler.
+
+        This is intentionally only used after bind returns EADDRINUSE in stdio
+        mode. A no-code request to our handler returns the shared
+        "Authentication Error" page with HTTP 400, which is enough to distinguish
+        the boot-time workspace callback listener from a random local process.
+        """
+        try:
+            parsed_uri = urlparse(self.base_uri)
+            scheme = parsed_uri.scheme or "http"
+            if scheme not in {"http", "https"}:
+                return False
+            probe_host = hostname
+            if probe_host in {"0.0.0.0", "::"}:
+                probe_host = "127.0.0.1"
+            if ":" in probe_host and not probe_host.startswith("["):
+                probe_host = f"[{probe_host}]"
+            base_path = parsed_uri.path.rstrip("/")
+            probe_url = f"{scheme}://{probe_host}:{self.port}{base_path}/oauth2callback"
+
+            try:
+                with urlopen(probe_url, timeout=1.0) as response:
+                    status_code = response.status
+                    body = response.read(4096).decode("utf-8", errors="replace")
+            except HTTPError as exc:
+                status_code = exc.code
+                body = exc.read(4096).decode("utf-8", errors="replace")
+
+            return status_code in {200, 400} and (
+                "Authentication Error" in body or "Authentication Successful" in body
+            )
+        except (OSError, URLError, ValueError):
+            return False
+
     def start(self) -> tuple[bool, str]:
         """
         Start the minimal OAuth server.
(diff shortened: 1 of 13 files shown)


taylorwilsdon:
@coderabbitai re-review

taylorwilsdon:
@coderabbitai you never posted the review

taylorwilsdon:
Extending this to also close #790

PR #776, opened 2026-05-08
fix(auth): resolve OAuth 2.1 credentials via fastmcp 3.x stores

### LINKED ISSUE #775: bug: _build_credentials_from_provider reads non-existent fastmcp attributes — Credentials always have refresh_token=None in OAuth 2.1 mode
## Bug

`auth/oauth21_session_store.py:_build_credentials_from_provider()` reads three attributes off the FastMCP auth provider that **do not exist on any version of fastmcp** (verified against 2.10.0, 3.0.0, 3.1.0, 3.2.0, 3.2.4):

- `_auth_provider._access_to_refresh`
- `_auth_provider._refresh_tokens`
- `_auth_provider._access_tokens`

Because the code uses `getattr(..., default={})`, the function never raises — it just silently returns `Credentials(refresh_token=None, ...)`. When `google-auth` later attempts to auto-refresh an expired access token, it raises:

```
RefreshError: The credentials do not contain the necessary fields need to refresh
the access token. You must specify refresh_token, token_uri, client_id, and client_secret.
```

## Reproduction

1. Deploy `google_workspace_mcp` in OAuth 2.1 mode (`--transport streamable-http`, `MCP_ENABLE_OAUTH21=true`).
2. Authenticate a user via the OAuth 2.1 flow.
3. From an MCP client (e.g. Claude Code), make sequential tool calls to `read_sheet_values` / `modify_sheet_values` / Gmail helpers across a session that lasts longer than the upstream Google access token's TTL (~1 hour).
4. After the access token expires, the next tool call triggers `google-auth.Credentials.refresh()`, which raises `RefreshError` immediately on the upfront None-check.

The error appears in logs as:

```
[TOOL] Unexpected refresh error for user <email>: The credentials do not contain
the necessary fields need to refresh the access token...
```

## Root Cause

In `auth/oauth21_session_store.py` (HEAD as of 2026-05-08, commit `9d69115`):

```python
def _build_credentials_from_provider(access_token: AccessToken) -> Optional[Credentials]:
    if not _auth_provider:
        return None

    access_entry = getattr(_auth_provider, "_access_tokens", {}).get(access_token.token)  # always {} -> None
    if not access_entry:
        access_entry = access_token

    client_id, client_secret = _resolve_client_credentials()

    refresh_token_value = getattr(_auth_provider, "_access_to_refresh", {}).get(  # always {} -> None
        access_token.token
    )
    refresh_token_obj = None
    if refresh_token_value:
        refresh_token_obj = getattr(_auth_provider, "_refresh_tokens", {}).get(  # always {} -> None
            refresh_token_value
        )
    ...
    return Credentials(
        token=access_token.token,
        refresh_token=refresh_token_obj.token if refresh_token_obj else None,  # ALWAYS None
        ...
    )
```

`_auth_provider` is a `GoogleProvider` (subclass of `OAuthProxy`), set in `core/server.py:541`. Neither class exposes `_access_to_refresh`, `_refresh_tokens`, or `_access_tokens`. The corresponding attribute on `InMemoryOAuthProvider` is named `_access_to_refresh_map` (different name, different class — used only in tests).

The bug was introduced on 2025-10-05 in commit `923df7e` ("refactor oauth2.1 support to fastmcp native"). The function appears to have been ported from a hypothetical fastmcp API that doesn't exist.

## Why it's intermittent

Google access tokens are valid for ~1h. Most LLM tool calls happen within that window so `google-auth` never invokes `refresh()`. The bug only fires for sessions that hit the auto-refresh path (long-running, sequential calls past the 1h boundary, or cross-restart sessions). This makes it look like sporadic re-auth prompts rather than a constant failure, and is likely why it has gone unreported.

## Tests don't catch it

`tests/auth/test_auth_info_middleware.py:_FakeProvider` is a `SimpleNamespace` mock that doesn't include the missing attributes either, so the test path produces the same `Credentials(refresh_token=None)` as production. The bug is invisible to the suite.

## Impact

In production with OAuth 2.1 multi-user mode, every long-running session's tool calls past the 1h mark force a full re-consent flow. Users hit `prompt=consent` Google flows multiple times per day; on consumer Gmail accounts this also revokes prior refresh tokens (cf. #584), compounding the issue.

## Suggested Fix

The right data lives in fastmcp 3.x's encrypted KV stores on the OAuth proxy:
- `_auth_provider._jti_mapping_store: PydanticAdapter[JTIMapping]` — maps JTI → `JTIMapping(upstream_token_id, ...)`
- `_auth_provider._upstream_token_store: PydanticAdapter[UpstreamTokenSet]` — maps `upstream_token_id` → `UpstreamTokenSet(access_token, refresh_token, ...)`

Per `fastmcp/server/auth/oauth_proxy/proxy.py:1071`, both access-token JTIs and refresh-token JTIs are stored in `_jti_mapping_store` with the same `upstream_token_id`, so we can resolve directly from the inbound access-token JWT's `jti` claim.

Sketch:

```python
async def _build_credentials_from_provider(access_token):
    if not _auth_provider:
        return None
    jti_store = getattr(_auth_provider, "_jti_mapping_store", None)
    upstream_store = getattr(_auth_provider, "_upstream_token_store", None)
    if jti_store is None or upstream_store is None:
        return None  # not running under FastMCP OAuthProxy

    # Decode the FastMCP JWT (signature already verified by middleware) to get jti
    import jwt as _jwt
    try:
        claims = _jwt.decode(access_token.token, options={"verify_signature": False})
        jti = claims.get("jti")
    except Exception:
        return None
    if not jti:
        return None

    jti_mapping = await jti_store.get(key=jti)
    if jti_mapping is None:
        return None
    upstream_set = await upstream_store.get(key=jti_mapping.upstream_token_id)
    if upstream_set is None:
        return None

    client_id, client_secret = _resolve_client_credentials()
    return Credentials(
        token=upstream_set.access_token,
        refresh_token=upstream_set.refresh_token,
        token_uri="https://oauth2.googleapis.com/token",
        client_id=client_id,
        client_secret=client_secret,
        scopes=upstream_set.scopes,
        expiry=_normalize_expiry_to_naive_utc(
            datetime.fromtimestamp(upstream_set.expires_at, tz=timezone.utc)
            if upstream_set.expires_at else None
        ),
    )
```

Both call sites of `ensure_session_from_access_token` (`auth/service_decorator.py:368`, `auth/auth_info_middleware.py:159`) are already inside `async` functions, so the function can be made `async` cleanly without callback cascading.

I'm running this in production with a local sed-patch fork; happy to put up a PR if the maintainers agree on the approach.

## Environment

- `taylorwilsdon/google_workspace_mcp` v1.20.4 (commit `9d69115`)
- `fastmcp==3.2.4`
- Transport: `streamable-http`
- `MCP_ENABLE_OAUTH21=true`
- Storage backend: disk (`/app/oauth-data`)

mvalee:
Closes #775.

## Summary

`auth/oauth21_session_store.py:_build_credentials_from_provider()` reads three attributes off the FastMCP auth provider (`_access_to_refresh`, `_refresh_tokens`, `_access_tokens`) that **do not exist on any version of fastmcp** (verified against 2.10.0, 3.0.0, 3.1.0, 3.2.0, 3.2.4). Because the code uses `getattr(..., default={})`, the function silently returns `Credentials(refresh_token=None, ...)`. When `google-auth` later attempts to auto-refresh an expired access token, it raises:

```
RefreshError: The credentials do not contain the necessary fields need to refresh
the access token. You must specify refresh_token, token_uri, client_id, and client_secret.
```

The bug was introduced in `923df7e` ("refactor oauth2.1 support to fastmcp native", 2025-10-05). It only fires for sessions that hit the auto-refresh path (calls past Google's ~1h access-token TTL), which is why it has appeared as sporadic re-auth prompts rather than a constant failure.

## Fix

Replace the broken `getattr` chain with a proper resolution via the OAuth proxy's stores:

```
inbound FastMCP JWT → jti claim
                    → _jti_mapping_store.get(jti) → JTIMapping
                    → _upstream_token_store.get(upstream_token_id) → UpstreamTokenSet
                    → Credentials(access_token, refresh_token, ...)
```

Both stores are exposed on `OAuthProxy` (and therefore on `GoogleProvider`, the actual `_auth_provider` set in `core/server.py:541`). They handle Fernet decryption transparently. Per `proxy.py:1071`, both access-token and refresh-token JTIs are stored in the same `_jti_mapping_store`, so resolving from the inbound access-token JWT works directly.

When the active provider isn't a FastMCP `OAuthProxy` (external OAuth 2.1, single-user, raw token verifier modes), `_build_credentials_from_provider` returns `None` and the caller falls back to the existing manual-construction path — preserving current behavior in those modes.

## Async cascade

`_build_credentials_from_provider`, `ensure_session_from_access_token`, and `get_credentials_from_token` are now `async` to match the async stores (`PydanticAdapter[T].get`). Both production call sites are already inside `async` functions:

- `auth/service_decorator.py:368` (in `async def get_authenticated_google_service_oauth21`)
- `auth/auth_info_middleware.py:159` (in `async def on_call_tool`)

…so the cascade is local to these files. Existing mocks in `tests/auth/test_auth_info_middleware.py` are updated from `lambda` to `async def` helpers.

## Tests

Added `tests/auth/test_oauth21_session_store.py` coverage:

- `test_extract_jti_from_jwt_*` — header-only JWT decode for the `jti` claim, with malformed-input handling.
- `test_build_credentials_returns_none_when_no_provider` / `_when_provider_lacks_proxy_stores` — graceful fallback for non-OAuthProxy providers.
- `test_build_credentials_resolves_via_jti_to_upstream_chain` — **the regression test for the bug**. Mocks the two stores, calls the function with a JWT containing a known `jti`, and asserts the returned `Credentials` has `refresh_token` populated from the `UpstreamTokenSet` (not silently `None`).
- `test_build_credentials_returns_none_when_jti_mapping_missing` / `_when_upstream_token_missing` — cache-miss / revoked-token paths.
- `test_ensure_session_falls_back_to_manual_construction_when_provider_missing` — non-proxy mode regression.

Full suite: `1004 passed, 2 skipped` (vs `1004 passed, 2 skipped` on `main`). `ruff check` and `ruff format` both clean.

## Production verification

I'm running this patch in production via a sed-based fork (separate repo) and have validated the JWT → JTI → upstream-token → Google `refresh_token` chain works end-to-end against fastmcp 3.2.4. Happy to share more deployment detail or rebase if you'd like a different shape — also fine if you'd rather port the same idea on top of an in-flight refactor.

## Summary by CodeRabbit

* **Improvements**
  * More reliable Google OAuth token verification and session establishment for smoother sign-in flows
  * Better handling and reconstruction of provider-issued credentials to improve access token resolution
  * Improved fallbacks to ensure access remains consistent when upstream mappings are unavailable

* **Tests**
  * Added and updated tests covering token-to-session resolution and edge-case fallbacks

### DIFF
--- a/auth/auth_info_middleware.py
+++ b/auth/auth_info_middleware.py
@@ -156,7 +156,7 @@ class AuthInfoMiddleware(Middleware):
                                         mcp_session_id = getattr(
                                             context.fastmcp_context, "session_id", None
                                         )
-                                        ensure_session_from_access_token(
+                                        await ensure_session_from_access_token(
                                             access_token,
                                             user_email,
                                             mcp_session_id,
--- a/auth/oauth21_session_store.py
+++ b/auth/oauth21_session_store.py
@@ -1032,42 +1032,95 @@ def _resolve_client_credentials() -> Tuple[Optional[str], Optional[str]]:
     return client_id, client_secret
 
 
-def _build_credentials_from_provider(
+def _extract_jti_from_jwt(token: str) -> Optional[str]:
+    """Extract the ``jti`` claim from a FastMCP-issued JWT without verifying.
+
+    The JWT signature has already been verified by the FastMCP middleware
+    before this code path runs; here we only need the ``jti`` to look up the
+    upstream token mapping. Decoding manually avoids depending on a specific
+    JWT library and avoids a redundant verification step.
+    """
+    try:
+        parts = token.split(".")
+        if len(parts) != 3:
+            return None
+        payload_b64 = parts[1] + "=" * (-len(parts[1]) % 4)
+        import base64
+        import json as _json
+
+        payload = _json.loads(base64.urlsafe_b64decode(payload_b64))
+        jti = payload.get("jti")
+        return jti if isinstance(jti, str) else None
+    except Exception:  # pragma: no cover - defensive
+        return None
+
+
+async def _build_credentials_from_provider(
     access_token: AccessToken,
 ) -> Optional[Credentials]:
-    """Construct Google credentials from the provider cache."""
+    """Construct Google credentials by resolving the upstream token via the
+    FastMCP OAuth proxy stores (``_jti_mapping_store`` -> ``_upstream_token_store``).
+
+    Returns ``None`` when the active provider is not a FastMCP ``OAuthProxy``
+    (e.g. external OAuth 2.1 providers, or single-user mode), so callers can
+    fall back to the manual-construction path.
+    """
     if not _auth_provider:
         return None
 
-    access_entry = getattr(_auth_provider, "_access_tokens", {}).get(access_token.token)
-    if not access_entry:
-        access_entry = access_token
+    jti_store = getattr(_auth_provider, "_jti_mapping_store", None)
+    upstream_store = getattr(_auth_provider, "_upstream_token_store", None)
+    if jti_store is None or upstream_store is None:
+        # Provider isn't a FastMCP OAuthProxy (e.g. raw token verifier in
+        # external-OAuth or single-user modes). Caller will fall back.
+        return None
+
+    jti = _extract_jti_from_jwt(access_token.token)
+    if not jti:
+        return None
+
+    try:
+        jti_mapping = await jti_store.get(key=jti)
+    except Exception as exc:  # pragma: no cover - defensive
+        logger.debug("JTI mapping lookup failed for jti=%s: %s", jti[:8], exc)
+        return None
+    if jti_mapping is None:
+        # Token JTI not in mapping store (revoked, expired, or never issued
+        # by this proxy). Treat as a cache miss and let the caller fall back.
+        return None
+
+    upstream_token_id = getattr(jti_mapping, "upstream_token_id", None)
+    if not upstream_token_id:
+        return None
+
+    try:
+        upstream_set = await upstream_store.get(key=upstream_token_id)
+    except Exception as exc:  # pragma: no cover - defensive
+        logger.debug(
+            "Upstream token lookup failed for id=%s: %s", upstream_token_id[:8], exc
+        )
+        return None
+    if upstream_set is None:
+        return None
 
     client_id, client_secret = _resolve_client_credentials()
 
-    refresh_token_value = getattr(_auth_provider, "_access_to_refresh", {}).get(
-        access_token.token
-    )
-    refresh_token_obj = None
-    if refresh_token_value:
-        refresh_token_obj = getattr(_auth_provider, "_refresh_tokens", {}).get(
-            refresh_token_value
-        )
-
     expiry = None
-    expires_at = getattr(access_entry, "expires_at", None)
-    if expires_at:
+    upstream_expires_at = getattr(upstream_set, "expires_at", None)
+    if upstream_expires_at:
         try:
-            expiry_candidate = datetime.fromtimestamp(expires_at, tz=timezone.utc)
-            expiry = _normalize_expiry_to_naive_utc(expiry_candidate)
+            expiry = _normalize_expiry_to_naive_utc(
+                datetime.fromtimestamp(upstream_expires_at, tz=timezone.utc)
+            )
         except Exception:  # pragma: no cover - defensive
             expiry = None
 
-    scopes = getattr(access_entry, "scopes", None)
+    scope_str = getattr(upstream_set, "scope", "") or ""
+    scopes = scope_str.split() if scope_str else None
 
     return Credentials(
-        token=access_token.token,
-        refresh_token=refresh_token_obj.token if refresh_token_obj else None,
+        token=upstream_set.access_token,
+        refresh_token=upstream_set.refresh_token,
         token_uri="https://oauth2.googleapis.com/token",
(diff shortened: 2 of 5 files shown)


adaofeliz:
hey @taylorwilsdon why did you moved this PR to Draft? it's actually an important fix, as the sessions are being expired in clients like OWUI.

taylorwilsdon:
Thanks for the thorough diagnosis here @mvalee — you correctly identified that the legacy `_access_tokens`/`_access_to_refresh`/`_refresh_tokens` reads no longer exist on fastmcp 3.x and that the upstream refresh token now lives behind the `jti → _jti_mapping_store → _upstream_token_store` chain.

Closing this in favor of a fresh implementation targeted directly at #886/#775 on current fastmcp 3.4.2. One behavioral difference worth noting: on 3.4.x the validated `AccessToken` reaching our code carries the already-swapped Google access token (not the FastMCP reference JWT), so the `jti` has to be read from the raw inbound `Authorization` bearer and verified via the provider's own `jwt_issuer`, rather than decoded from `access_token.token`. Appreciate the groundwork — it made the correct fix much faster to land.

PR #777, opened 2026-05-08
fix(gslides): expose strict schema for Slides batch updates

### LINKED ISSUE #769: batch_update_presentation allows empty request objects, causing "No request set"
When an LLM uses batch_update_presentation, it can repeatedly call:

{
  "presentation_id": "...",
  "requests": [{}]
}

Google Slides API returns:
Invalid requests[0]: No request set.

Expected:
The MCP server should reject empty/unknown request objects before calling Google API and return actionable guidance, e.g.:
requests[0] must contain exactly one Slides request type such as createSlide, createShape, insertText, updateTextStyle, createImage, deleteObject, etc.

This is related to #753/#760 but different: #760 validates insertText targets, while this issue is about empty request objects and schema/tool guidance.

YuriyTW:
## Summary

This updates the Google Slides `batch_update_presentation` tool schema so strict-schema MCP clients can construct valid Slides `batchUpdate` request objects instead of collapsing each request item to `{}`.

The runtime input shape stays the same (`List[Dict[str, Any]]`), but the tool now publishes explicit `anyOf` schema variants for the Google Slides batchUpdate request union. It also strips `null` values before calling the Google API so strict nullable fields can be represented safely without leaking `null` into the request payload.

Fixes #769

## Why

Some strict-schema clients normalize generic object inputs into closed empty objects. For this tool, that can make models repeatedly call the tool with:

```json
{
  "requests": [{}]
}
```
  

## Notes

- This change intentionally keeps `DuplicateObjectRequest.objectIds` out of the strict schema. That field is a free-form string-to-string map, and exposing it directly would require reopening the object shape and would bring back the same empty-object failure mode for strict clients.
- The raw emitted schema for this tool grows by about 13k tokens in strict mode (`~96 -> ~14,023`, delta `~13,927`). That is an intentional tradeoff: the larger schema gives agents enough structure to understand the Slides request union and fill arguments correctly, which is preferable to repeated invalid `{}` batchUpdate calls.
- Downstream runtimes that care about token budget can still keep a JSON-string passthrough adapter.

## Release Notes

* **New Features**
  * Enhanced validation for Google Slides batch update requests with strict schema enforcement.
  * Automatic removal of null values and empty nested objects in batch requests for cleaner API calls.
  * Improved OAuth authentication flow with browser auto-opening for legacy sign-in.

* **Tests**
  * Added comprehensive test coverage for batch update schema validation, request handling, and OAuth authentication scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/taylorwilsdon/google_workspace_mcp/pull/777?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


YuriyTW:
What's holding me back is that it takes 13k tokens! What do you think about that?

taylorwilsdon:
Hm, was this not resolved for you with yesterday's https://github.com/taylorwilsdon/google_workspace_mcp/pull/770?

I did fairly extensive manual testing and even haiku was able to put together elaborate slides in a single shot following the validation logic input.

YuriyTW:
Yes, #770 fixes the validation/recovery path, and that part looks good.

What I’m trying to address here is a slightly different issue: in v1.20.4 the MCP schema for `batch_update_presentation.requests.items` is still just a generic object with `additionalProperties: true`, so a client consuming the MCP schema directly does not get field-level structure for the Slides request union.

That means the agent has to infer payload shapes from prose and validation errors rather than from the tool schema itself. It can recover, but in our testing this can still lead to a few invalid argument attempts before it lands on the correct payload.

This PR is meant to improve first-shot argument construction by exposing the request types and fields directly in the schema. The tradeoff is schema size: the raw emitted schema grows by about 13k tokens. I think that can be justified for agents that rely heavily on tool schemas, but if you prefer keeping the schema small and relying on #770’s validation guidance, I understand that too.

taylorwilsdon:
Great, will take a look. Thanks for the detailed reply!

oliver-mee:
Thanks for putting this together. I tested the Slides portion against a production remote MCP deployment and integrated it into a deployment branch.

The schema change solved a real client issue:

- `batch_update_presentation.requests` now exposes field-level shapes for all 44 variants.
- The full suite passed: 1,808 passed, 2 skipped.
- A built deployment reports all 44 variants, including `createShape` and `insertText`.
- The runtime remains raw request dictionaries, so this does not force a breaking Python API change.

I also applied the single-source-of-truth and nested-list cleanup changes from review. The feature is now deployed for that client, with the previous image retained for rollback.

Thanks for the work here. The field-level schema is valuable for clients that construct MCP arguments directly.

PR #787, opened 2026-05-12
feat(oauth): separate port for stdio-mode OAuth callback server

JSaun77:
## Problem

In stdio mode, the `MinimalOAuthServer` binds the same port as the main MCP server. When an OAuth refresh fires while the main server is already listening (e.g., the wrapper pins `WORKSPACE_MCP_PORT=18765` to avoid `PORT` collisions and that port is held by a prior process), the callback server cannot bind and dies with:

```
Port 18765 is already in use on localhost. Cannot start minimal OAuth server.
```

This is recurring for users whose OAuth consent screen is in Testing mode (refresh tokens expire every 7 days, forcing weekly re-auth) and is non-obvious to debug because the error points at the symptom rather than the cause.

## Fix

Add `WORKSPACE_MCP_OAUTH_CALLBACK_PORT` env var so the callback server can bind a separate port from the main MCP server. Defaults to the main port for backward compatibility. Updates redirect URI derivation, the three call sites that preflight the callback server, and the port-in-use error message.

## Tests

All 133 tests in `tests/auth` and `tests/core` pass (excluding unrelated `test_gcs_credential_store` collection error from missing optional dep). Mock SimpleNamespace in `test_start_google_auth.py` updated with the new attribute.

## Summary by CodeRabbit

* **New Features**
  * OAuth callback server now supports a separate dedicated port for greater flexibility in network setups.

* **Bug Fixes**
  * Improved error messaging when the OAuth callback port is unavailable, with clearer causes and guidance including the expected redirect URI format.

* **Tests**
  * Updated tests to account for the separate callback port behavior during OAuth preflight checks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/taylorwilsdon/google_workspace_mcp/pull/787)

### DIFF
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -1278,7 +1278,7 @@ async def get_authenticated_google_service(
             success, error_msg = await asyncio.to_thread(
                 ensure_oauth_callback_available,
                 transport_mode,
-                config.port,
+                config.oauth_callback_port,
                 config.base_uri,
             )
             if not success:
--- a/auth/oauth_callback_server.py
+++ b/auth/oauth_callback_server.py
@@ -229,7 +229,16 @@ class MinimalOAuthServer:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
                 s.bind((hostname, self.port))
         except OSError:
-            error_msg = f"Port {self.port} is already in use on {hostname}. Cannot start minimal OAuth server."
+            error_msg = (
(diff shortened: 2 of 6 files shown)


taylorwilsdon:
We already have auth/port_resolver.py that automatically finds a free port with a fallback range, this PR seems to manually reinvent that with a new env var. I'm concerned about a footgun where users now have to go  reconfigure their gcs redirect URI every time they change the callback port. From what I can tell, the problem is a zombie process on the user's machine.

PR #793, opened 2026-05-14
fix: try alternative ports when OAuth callback port is in use

zmainen:
## Summary

When multiple Claude Code sessions run on the same machine, each spawns its own `workspace-mcp` instance in stdio mode. The first instance binds port 8000 for the OAuth callback server; subsequent instances fail with "Port 8000 is already in use" and cannot complete any OAuth flow — even though cached credentials exist and the callback server is only needed for fresh auth.

This PR makes the OAuth callback server try alternative ports (8001, 8002, ...) when the preferred port is already in use, up to a configurable range (`WORKSPACE_MCP_PORT_RANGE` env var, default 10). Google allows any localhost port for loopback redirect URIs on installed/desktop OAuth clients ([RFC 8252 §7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3)), so no pre-registration is needed. The `OAuthConfig.redirect_uri` is updated dynamically to match the port that was actually bound.

## Changes

- **`auth/oauth_callback_server.py`**: In `ensure_oauth_callback_available()`, when `MinimalOAuthServer.start()` fails due to port conflict, iterate through alternative ports and update the OAuth config's redirect URI to match.
- **`tests/auth/test_oauth_callback_server.py`**: Two new tests — one for successful fallback to an alternative port, one for graceful failure when all ports are exhausted.

## Test plan

- [x] All 7 existing + new tests pass (`pytest tests/auth/test_oauth_callback_server.py`)
- [ ] Manual test: run two Claude Code sessions simultaneously, verify both can complete OAuth flows

## Summary by CodeRabbit

* **Bug Fixes**
  * OAuth callback server now automatically attempts alternative ports when the preferred port is unavailable, updating configuration to match the successfully bound port. If all ports fail, an error is logged.

* **Tests**
  * Added comprehensive test coverage for OAuth callback server port fallback behavior, including successful alternative port binding and port exhaustion scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/taylorwilsdon/google_workspace_mcp/pull/793)

### DIFF
--- a/auth/oauth_callback_server.py
+++ b/auth/oauth_callback_server.py
@@ -307,7 +307,9 @@ def ensure_oauth_callback_available(
     Ensure OAuth callback endpoint is available for the given transport mode.
 
     For streamable-http: Assumes the main server is already running
-    For stdio: Starts a minimal server if needed
+    For stdio: Starts a minimal server if needed, trying alternative ports
+               if the preferred port is already in use (common when multiple
+               Claude Code sessions each spawn their own MCP server instance).
 
     Args:
         transport_mode: "stdio" or "streamable-http"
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
This capability should be live in main now, give it a try and let me know if you have any issues!

PR #830, opened 2026-06-05
fix(auth): longer-lived refresh tokens (FastMCP 3.4.0) + fix proxied/embedded OAuth consent 403

Aiacos:
## Summary

Two related OAuth/login robustness improvements, motivated by self-hosting this server behind a reverse proxy and connecting it from remote MCP clients (Perplexity, Claude web).

### 1. `chore`: bump FastMCP `3.2.4 → 3.4.0` (longer-lived refresh tokens)

`OAuthProxy` in 3.4.0:
- Raises the default FastMCP-issued **refresh-token lifetime from 30 days to 1 year** (`DEFAULT_REFRESH_TOKEN_EXPIRY_SECONDS`).
- Fixes refresh tokens **collapsing to the ~1h access-token TTL** when the upstream IdP (Google) returns no `refresh_token` on a refresh exchange — observed in practice as MCP refresh tokens stored with ~30–70 min expiry, forcing frequent re-login.
- Adds configurable `fallback_refresh_token_expiry_seconds` on `GoogleProvider`.

No code changes required beyond the version bump; the existing `GoogleProvider(...)` call is signature-compatible.

### 2. `fix(auth)`: default OAuth proxy consent to `"external"` to fix callback 403

When `require_authorization_consent=True` (current default), the built-in consent screen sets a `__Host-*` `SameSite=Lax` binding cookie that must round-trip through the Google redirect back to `/oauth2callback`. Behind a reverse proxy, or inside an **embedded OAuth webview** (e.g. Perplexity's connector flow), that cookie is not returned, so the callback rejects with:

> 403 — Authorization session mismatch. This can happen if you followed a link from another person or your session expired.

This makes the server effectively unusable from such clients. This change defaults `require_authorization_consent` to `"external"` (consent is still fully gated by **Google's own consent screen**, which already runs with `prompt=consent`), and makes it overridable:

```
WORKSPACE_MCP_AUTH_CONSENT = true | remember | external | false   # default: external
```

**Note for maintainers:** this changes a security-relevant default. If you'd rather keep `True` as the default and only expose the env override, I'm happy to adjust — just say the word. The env knob alone already lets self-hosters opt in.

## Testing

- Container builds on `v1.21.1` + these two commits.
- `/health` → healthy; `/.well-known/oauth-authorization-server` and `/.well-known/oauth-protected-resource/mcp` → 200.
- Startup logs: `Built-in consent screen disabled; consent is handled externally.`
- OAuth 2.1 GoogleProvider initialises with no errors.

## Summary by CodeRabbit

* **New Features**
  * OAuth 2.1 authorization consent behavior can be controlled via an environment variable, giving admins configurable consent flows for Google sign-in.

* **Chores**
  * Updated runtime dependency requirement for fastmcp to a newer minimum version.

### DIFF
--- a/core/server.py
+++ b/core/server.py
@@ -590,6 +590,24 @@ def configure_server_for_http():
                         "OAuth 2.1: restricting DCR client redirect URIs to allowlist: %s",
                         allowed_client_redirect_uris,
                     )
+                # Consent mode for the OAuth proxy. Defaults to "external" because
+                # Google's own consent screen already gates authorization. The
+                # built-in consent screen sets a __Host-* SameSite=Lax binding
+                # cookie that must round-trip through the Google redirect; behind a
+                # reverse proxy or inside an embedded OAuth webview (e.g. Perplexity)
+                # that cookie is not returned, producing a 403 "Authorization
+                # session mismatch" on the IdP callback. "external" skips the
+                # built-in consent + binding check while keeping Google's consent.
+                # Override with WORKSPACE_MCP_AUTH_CONSENT=true|remember|external|false.
+                _consent_raw = os.getenv(
+                    "WORKSPACE_MCP_AUTH_CONSENT", "external"
+                ).strip().lower()
+                if _consent_raw in ("external", "remember"):
+                    require_authorization_consent = _consent_raw
+                elif _consent_raw in ("false", "0", "no", "off"):
+                    require_authorization_consent = False
+                else:
+                    require_authorization_consent = True
                 provider = GoogleProvider(
                     client_id=config.client_id,
                     client_secret=config.client_secret,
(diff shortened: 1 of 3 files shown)


Aiacos:
Addressed the CodeRabbit finding on the `WORKSPACE_MCP_AUTH_CONSENT` parsing in `d930559`.

Blank (`WORKSPACE_MCP_AUTH_CONSENT=`) or unrecognized values no longer fall through to `True` (which would re-enable the built-in consent + binding-cookie flow and bring back the proxied/embedded callback 403 this PR fixes). They now default to `"external"`, with a warning logged for invalid input. Explicit true-like strings (`true|1|yes|on`) are handled distinctly so the override still works as documented.

```
unset / blank        -> "external"
external | remember  -> as-is
false | 0 | no | off -> False
true | 1 | yes | on  -> True
anything else        -> "external" (+ warning)
```

PR #840, opened 2026-06-09
fix: add httplib2 socket timeout to prevent indefinite hangs (#835)

### LINKED ISSUE #835: Gmail API calls hang indefinitely — httplib2 missing socket timeout
## Problem

When making Gmail API calls, if a specific email triggers a slow or stalled server response (or a TCP connection enters a half-open state), the `socket.read()` inside `httplib2` blocks indefinitely — no socket-level timeout is configured.

Tool functions use `asyncio.to_thread()` to wrap synchronous Google API calls. Since Python cannot cancel a blocked thread, a single hung call eventually makes the entire workspace-mcp process unresponsive: all subsequent MCP calls queue behind the blocked thread and time out on the caller's side.

**Observed:** A specific Gmail message caused every `get_gmail_message_content` call to hang. The process stayed "alive" but frozen — calls that didn't touch the problematic message continued to work, while every email-body fetch timed out at 20s (our caller-side limit).

## Root Cause

`auth/google_auth.py` builds the Google service client without an explicit `http` parameter:

```python
service = build(service_name, version, credentials=credentials)
```

`googleapiclient` then creates an `httplib2.Http()` instance with **no socket timeout** (httplib2 default). Any `socket.recv()` that stalls will block the thread forever.

## Proposed Fix

Pass `httplib2.Http(timeout=N)` when building the service:

```python
import httplib2
from google.auth.transport import httplib2 as auth_httplib2

http = httplib2.Http(timeout=30)
authed_http = auth_httplib2.AuthorizedHttp(credentials, http=http)
service = build(service_name, version, http=authed_http)
```

This gives every Google API call a hard socket-level timeout, making the process resilient to network stalls without leaking background threads.

## Environment

- workspace-mcp: 1.21.1 (latest on PyPI at time of writing)
- Python: 3.11
- Transport: httplib2 (HTTP/1.1, not HTTP/2)

## Workaround

We added `asyncio.wait_for()` around each `asyncio.to_thread()` call in `gmail_tools.py` (18s timeout for reads, 15s for writes). This prevents process-wide hangs but the underlying thread still runs in the background after cancellation. Combined with a health check that kills the subprocess after N consecutive timeouts, this provides adequate resilience — but a proper socket-level timeout would be the clean fix.

simongonzalezdc:
Closes: #835 

## Problem

`googleapiclient.discovery.build()` with `credentials=credentials` internally creates an `httplib2.Http()` with no socket timeout (default `None`). On stalled or flaky network connections, this causes indefinite hangs rather than a clean timeout error.

Issue #835 reports this as a hard hang during service initialization.

### `auth/google_auth.py`

- Added `_build_authorized_http(credentials, timeout=30)` helper that explicitly constructs an `httplib2.Http(timeout=30)` and wraps it with `google_auth_httplib2.AuthorizedHttp`
- Updated `get_user_info()` to pass `http=_build_authorized_http(credentials)` instead of `credentials=credentials`
- Updated `_get_google_service()` (the main service builder) with the same fix

### Dependencies

`google-auth-httplib2>=0.2.0` and `httplib2` are already declared in `pyproject.toml` and locked in `uv.lock` — no new dependencies.

### Tests

- `tests/auth/test_httplib2_timeout.py`: two focused unit tests verifying:
  1. Explicit timeout value is forwarded to `httplib2.Http`
  2. Default timeout is 30 seconds

## Verification

```bash
uv run pytest tests/auth/test_httplib2_timeout.py -v
# 2 passed

uv run pytest tests/auth/ -v
# 127 passed
```

Fixes #835

## Summary by CodeRabbit

* **Bug Fixes**
  * Google API authentication now uses an explicit socket timeout to prevent indefinite request hangs. A 30-second default is applied and custom timeout values are supported.

* **Tests**
  * Added regression tests verifying explicit timeout propagation, the 30-second default behavior, and that authenticated Google requests are built using the timed HTTP transport.

### DIFF
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -89,13 +89,7 @@ DEFAULT_CREDENTIALS_DIR = get_default_credentials_dir()
 def _build_authorized_http(
     credentials: Credentials, timeout: int = 30
 ) -> google_auth_httplib2.AuthorizedHttp:
-    """Return an httplib2.Http wrapped with credentials and a socket timeout.
-
-    Issue #835: googleapiclient.discovery.build() with credentials=credentials
-    creates an httplib2.Http() with no timeout (default None), causing
-    indefinite hangs on stalled TCP. We explicitly construct an
-    AuthorizedHttp with a timeout and pass it as the *http* argument.
-    """
+    """Return credentialed HTTP with an explicit socket timeout."""
     http = httplib2.Http(timeout=timeout)
     return google_auth_httplib2.AuthorizedHttp(credentials, http=http)
 
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -17,6 +17,8 @@ from google.auth.transport.requests import Request
 from google.auth.exceptions import RefreshError
 from googleapiclient.discovery import build
 from googleapiclient.errors import HttpError
+import httplib2
+import google_auth_httplib2
 from auth.scopes import SCOPES, get_current_scopes, has_required_scopes  # noqa
 from auth.oauth21_session_store import get_oauth21_session_store
 from auth.credential_store import get_credential_store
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
Thanks!

PR #846, opened 2026-06-12
feat(gmail): surface inline (cid:) image mapping in get_gmail_message_content

DannyOosterveer:
## Problem

HTML emails reference inline images (signature logos, pasted screenshots) via `src="cid:<content-id>"`. To render them, a client must know **which attachment each Content-ID maps to**. Today that mapping isn't reachable: the `--- ATTACHMENTS ---` listing exposes `attachmentId` but not the part's `Content-ID`, and `body_format="raw"` is truncated — so clients are forced to guess by part order or size, which gets inline images wrong when e.g. a small signature logo and a large screenshot are both present.

## Change (updated per review)

This originally added a dedicated `get_gmail_message_attachment_map` tool. Per @taylorwilsdon's note that the server is at the tool limit, the mapping is now folded into the **existing** message-content tools instead — no new tool:

- `_extract_attachments` now reads each part's `Content-ID`, and also includes filename-less inline (`cid:`) parts that carry an `attachmentId`.
- The `--- ATTACHMENTS ---` listing (in `get_gmail_message_content`, the batch variant, and the thread-content variant) shows a `Content-ID: <cid>` line for inline parts, via a shared `_format_attachment_entry` helper.

A client can then map `src="cid:<id>"` to the correct attachment and fetch it via `get_gmail_attachment_content` — no heuristics needed.

Removed `get_gmail_message_attachment_map` and the now-unused `_extract_cid_map`.

## Tests

All gmail tests pass (`uv run pytest tests/gmail/` — 129 passed), lint clean (`ruff check`).

## Summary by CodeRabbit

* **New Features**
  * Enhanced Gmail attachment extraction to capture `Content-ID` for inline (`cid:`) parts, so embedded images can be identified with their `cid` mapping alongside filename, MIME type, size, and attachment identifiers in structured JSON output.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -8,6 +8,7 @@ import logging
 import asyncio
 import base64
 import binascii
+import json
 import re
 import ssl
 import mimetypes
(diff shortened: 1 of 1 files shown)


DannyOosterveer:
Refactored per your tool-limit note on #455 — dropped the dedicated `get_gmail_message_attachment_map` tool and folded the Content-ID → attachment mapping into the existing message-content tools instead. `_extract_attachments` now carries each part's `Content-ID` (and includes filename-less inline parts), and the `--- ATTACHMENTS ---` listing shows a `Content-ID: <cid>` line for inline parts via a shared helper. No new tools; 129 gmail tests pass, lint clean.

PR #848, opened 2026-06-12
fix: issues/845

### LINKED ISSUE #845: draft_gmail_message: thread_id reply drafts not visible in Gmail UI — regression/variant of #630 on v1.21.2
## Bug

Drafts created via `draft_gmail_message` with `thread_id` succeed at the API level — the tool returns a draft ID, the draft appears in `in:drafts` searches and in `get_gmail_thread_content` — but **never appear in the Gmail web UI Drafts folder**. The user cannot see, edit, or send them.

This is the same parameter family as #630 (closed by #631, shipped ~v1.17.0), reproducing on **v1.21.2**. Symptom differs slightly: #630 described the draft as *visible in the list but empty on open*; here the draft is *absent from the Drafts UI entirely*. Possibly the same root cause with a different rendering outcome.

## Steps to Reproduce

1. Call `draft_gmail_message` with:
   - `thread_id` (valid Gmail thread ID)
   - `in_reply_to` + `references` (valid RFC Message-ID from that thread)
   - `to`, `cc`, `subject` set; non-empty HTML `body`, `body_format: "html"`
   - `include_signature: true`
2. Tool returns `Draft created! Draft ID: r...`
3. `search_gmail_messages` with `in:drafts <term>` → draft IS returned; `get_gmail_thread_content` shows it appended to the thread with correct headers/recipients/body
4. Open Gmail web UI → Drafts folder

## Expected

Draft visible in Drafts, opens with body and recipients populated.

## Actual

Draft is not shown in the Drafts UI at all. Reproduced 2/2 with `thread_id` set (two separate calls, same thread, minutes apart).

## Parameter isolation

Four drafts created the same day against the same account:

| # | thread_id | in_reply_to/references | Visible in Gmail UI |
|---|-----------|------------------------|---------------------|
| 1 | ✅ | ✅ | ❌ ghost |
| 2 | ✅ | ✅ | ❌ ghost |
| 3 | ❌ | ✅ | ✅ renders |
| 4 | ❌ | ❌ | ✅ renders, sent fine |

`thread_id` alone appears to be the trigger; `in_reply_to`/`references` without `thread_id` is fine.

## Workaround

Standalone draft (no `thread_id`) with the external ticket number kept in the subject — loses Gmail-side thread grouping but renders, sends, and threads correctly on the recipient side via the reply headers / subject match.

## Environment

- workspace-mcp **1.21.2** (uvx, unpinned — includes the #631 fix from ~v1.17.0)
- `--tools gmail drive sheets calendar --tool-tier extended`
- Gmail API via OAuth, Google Workspace account
- macOS (darwin 24.3), Claude Code CLI
- `body_format: "html"`, `include_signature: true` (also the default-signature path)

taylorwilsdon:
Closes: #845 

## Release Notes

* **Bug Fixes**
  * Fixed an issue where draft messages would appear hidden in the Gmail UI when replying to threads. Thread information is now properly managed through message headers.


mvalee:
Heads-up that this change (dropping `message.threadId` from `drafts().create`) fixes the ghost-draft *visibility* but **removes draft nesting entirely** — reply drafts no longer thread.

Gmail only header-threads on **send**; for a *draft* to appear inside its thread, `message.threadId` must be set on the draft message (which `send_gmail_message` still does). So post-#848, `draft_gmail_message(thread_id=...)` lands the draft in a brand-new standalone thread instead of the target one.

Downstream we restored `threadId` on the draft body and the drafts now nest **and** render correctly in the Gmail web UI.

We also tested the subject-mismatch theory directly, since #845's repro pointed at the `threadId`+`in_reply_to` combo: a draft with `threadId` + a valid **visible** parent `In-Reply-To`, but a deliberately **wrong** subject — and it was **still visible** in the UI. So subject mismatch does not appear to be the ghost trigger.

The known ghost path is the one #728 already fixed (an `In-Reply-To` parent that isn't visible, e.g. trashed). Given that, removing `threadId` looks like the wrong remedy — it trades a narrow visibility edge case for breaking threading on every reply draft. Suggest restoring `message.threadId` (criterion 1 of Gmail's [threading contract](https://developers.google.com/workspace/gmail/api/guides/threads)) and keeping #728's visible-parent handling.

PR #852, opened 2026-06-13
feat(gdocs): section-addressable markdown editing — replace/append/replace-fully by heading

johntrandall:
## Description

Adds three MCP tools that let agents address Docs edits by **heading text** instead of `(startIndex, endIndex)` character offsets:

| Tool | Behavior |
|---|---|
| `replace_doc_section_by_heading` | Atomically deletes a heading-delimited section and inserts new markdown in a single batchUpdate. |
| `append_doc_after_heading` | Pure insertion at the section end. Nothing is deleted/orphaned. |
| `replace_doc_fully_from_markdown` | Wipes the body (defensively computing the protected trailing-newline index) and inserts new markdown at index 1. Preserves the file ID — distinct from `import_to_google_doc`. **Caution:** body-scoped named ranges are destroyed; header/footer/footnote named ranges survive. |

## Motivation

The Docs API addresses edits by `(startIndex, endIndex)` character offsets in a structural-element tree. Agents struggle:

- Falling back to `import_to_google_doc` against an existing doc creates a NEW file with a NEW ID, breaking every external link to the original.
- Issuing dozens of `batch_update_doc` calls with hand-computed indices drifts after the first insertion — each `InsertTextRequest` shifts every subsequent index.

The repo already has the primitives: `get_doc_as_markdown` (with comments) and `markdown_to_docs_requests(...,start_index=N)` (CommonMark → batchUpdate request list at any index). What's missing is **section-addressable editing** — tools that locate a target by heading text and atomically delete + insert markdown at that slot.

## Safety guarantees

All three tools:

- Resolve indices via a `documents.get` with `includeTabsContent=True`, wrapped in `asyncio.wait_for(timeout=30)` per `get_doc_as_markdown` precedent (large-doc safety).
- Use `doc.get("revisionId")` and raise `UserInputError` if absent (read-only tokens omit the field).
- Pass `WriteControl.requiredRevisionId` to `batchUpdate` so a concurrent edit between read and write returns 400 with no partial application.
- Require `tab_id` for multi-tab docs (UserInputError lists available tab IDs).
- **Pre-check the target range for footnote references** — the API does not document cross-`footnoteReference` delete semantics, so the tool raises `UserInputError` with a remediation message instead of discovering at runtime. Suggests `batch_update_doc` for surgical edits in footnoted sections.
- Restrict matching to body paragraphs (headings inside table cells, footnotes, headers, or footers are not addressable; use `batch_update_doc` for those).

Section end is the start of the next paragraph whose `namedStyleType` is a heading at equal-or-shallower level, or `TITLE`. `SUBTITLE` is **not** a terminator (functions as a continuation of `TITLE` in Docs UI).

## Non-goals

- **Pure delete** — use `replace_doc_section_by_heading` with empty `new_markdown` (deletes the section's body, keeps the heading).
- **Cross-tab move.**
- **Header / footer / footnote section editing** — use `batch_update_doc` with the appropriate `segmentId`.

## Helpers added

Three new **public** helpers in `gdocs/docs_structure.py` (exposed so other callers can reuse the same body-only heading-lookup logic):

- **`find_heading_range(doc, heading_text, heading_level, match, tab_id) -> (section_start, section_end, footnote_count)`** — body-only heading lookup with multi-tab handling.
- **`body_protected_end_index(doc, tab_id)`** — walks backward through `body.content` to find the last paragraph (the Docs API does NOT guarantee the last body element is a paragraph) and returns `endIndex - 1` to avoid "Deleting the last newline character of a Body" 400.
- **`count_footnote_refs_in_range(doc, tab_id, start, end)`** — pre-check.

## Preservation matrix (full table in README appendix)

Outside the touched range: comments anchored, suggestions live, named ranges intact. Inside a deleted range: comments orphan in the UI but remain `deleted=false` in `comments.list` (verified by absence of any orphan-filter parameter in the Drive API); suggestions are silently dropped (the API has no programmatic accept/reject path). Footnote references trigger the pre-check before any write. Cross-segment deletes (body → header/footer/footnote) are not expressible in the API — `Range.segmentId` is single-segment.

## Related PRs (recommended to land first)

**#850** — `fix(gdocs): emit NORMAL_TEXT reset for body paragraphs + spacers in markdown_to_docs_requests`. Strongly recommend landing #850 first: without that fix, body paragraphs inserted by these section tools inherit the surrounding `HEADING_N` style at boundaries — the resulting docs read as "all headings". Tests in this PR are mocked and pass standalone, but the live rendering is much cleaner with #850 merged. Also pairs well with #851 (`copy_doc_as_snapshot`) as a recommended pre-edit safety net.

## Tests

24 new mocked unit tests in `tests/gdocs/test_section_edit.py` covering:
- `find_heading_range`: single/multiple match, TITLE/SUBTITLE behavior, inline-object rejection, multi-tab requirement, `namedStyleType` absent default, footnote count piggyback
- `body_protected_end_index`: empty body, walks-backward-from-sectionBreak-tail
- `count_footnote_refs_in_range`
- All three section tools: single batchUpdate, writeControl, request order, footnote pre-check fires before any API call

Full suite on this branch (`upstream/main` + this feature): **1243 passed, 2 deselected, 0 failures**.

Reproducible: `git checkout upstream/main && uv run pytest tests/ -m "not integration"` → 1219 passed; this PR adds 24 tests → 1243.

## Type of Change

## Testing

- [x] Tests added that prove the feature works
- [x] Existing tests pass
- [x] **Local integration test** against a real Doc:
  - `replace_doc_section_by_heading("Section B", "# Section B (replaced)\n\nBody.\n\n**Bold** para.")` → Section B body replaced, Section A and C untouched, doc ID preserved
  - `append_doc_after_heading("Section A", "Body para.")` → inserted at section_end, A's content untouched
  - `replace_doc_fully_from_markdown("# Fresh\n\nBody.\n\n## Sub\n\nMore.")` → body wiped + replaced, doc ID survives, body named ranges destroyed (matrix-confirmed)
  - Error path: non-existent heading → clean `UserInputError`, no doc change
  - Multi-tab without `tab_id` → `UserInputError` listing tab IDs

## Checklist

- [x] My code follows the style guidelines of this project (`ruff check`, `ruff format --check` both clean)
- [x] Self-reviewed
- [x] No new warnings

## Additional Notes

Backwards compatible: no existing tools renamed, removed, or signature-changed. The `markdown_to_docs_requests` converter is consumed unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1643,7 +1643,7 @@ Validations:
 
 ## Section-Edit Preservation Matrix
 
-The three Docs section-edit tools (`replace_doc_section_by_heading`, `append_doc_after_heading`, `replace_doc_fully_from_markdown`) operate via atomic `documents.batchUpdate` calls guarded by `WriteControl.requiredRevisionId`. Combine with `copy_doc_as_snapshot` (from the parallel `copy_doc_as_snapshot` PR / tool) before destructive edits as the recommended safety net.
+The three Docs section-edit tools (`replace_doc_section_by_heading`, `append_doc_after_heading`, `replace_doc_fully_from_markdown`) operate via atomic `documents.batchUpdate` calls guarded by `WriteControl.requiredRevisionId`. Pair with `copy_doc_as_snapshot` before destructive edits as the recommended safety net.
 
 **Legend:** ✅ kept • ⚠ docstring warning applies • ❌ destroyed • n/a = not applicable
 **Confidence:** V = Verified against Google docs • I = Inferred (verify via local integration test)
--- a/README.md
+++ b/README.md
@@ -858,6 +858,9 @@ Saved files expire after 1 hour and are cleaned up automatically.
 | <sub>`insert_doc_elements`</sub> | <sub>Extended</sub> | <sub>Add tables, lists, page breaks</sub> |
 | <sub>`update_paragraph_style`</sub> | <sub>Extended</sub> | <sub>Apply advanced paragraph styling including headings, spacing, direction, pagination controls, shading, and bulleted/numbered/checkbox lists with nesting</sub> |
 | <sub>`get_doc_as_markdown`</sub> | <sub>Extended</sub> | <sub>Export document as formatted Markdown with optional comments</sub> |
+| <sub>`replace_doc_section_by_heading`</sub> | <sub>Extended</sub> | <sub>Replace a heading-delimited section's contents with new markdown, in a single batchUpdate guarded by `WriteControl.requiredRevisionId`. Pre-checks for footnote refs in the target range. See [Section-Edit Preservation Matrix](#section-edit-preservation-matrix).</sub> |
+| <sub>`append_doc_after_heading`</sub> | <sub>Extended</sub> | <sub>Insert markdown at the end of a heading-delimited section. Pure insertion — nothing is deleted or orphaned.</sub> |
+| <sub>`replace_doc_fully_from_markdown`</sub> | <sub>Extended</sub> | <sub>Replace the entire body of a Doc with new markdown while preserving the file ID (distinct from `import_to_google_doc` which creates a new file).</sub> |
 | <sub>`insert_doc_image`</sub> | <sub>Complete</sub> | <sub>Insert images from Drive/URLs</sub> |
 | <sub>`update_doc_headers_footers`</sub> | <sub>Complete</sub> | <sub>Create or update headers and footers with correct segment-aware writes</sub> |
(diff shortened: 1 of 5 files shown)


taylorwilsdon:
Ooh I love the idea, but can we bake these into an existing tool somewhere? We are at the maximum number of tools for an OpenAI compatible native tool calling payload so we effectively cannot add more tools without breaking usage for folks who don't disable certain tiers or tool types.

PR #862, opened 2026-06-18
fix(forms): send valid setPublishSettings request body

jfrantz-cw:
## Problem

`set_publish_settings` fails on every call with HTTP 400:

```
Invalid JSON payload received. Unknown name "publishAsTemplate": Cannot find field.
Invalid JSON payload received. Unknown name "requireAuthentication": Cannot find field.
```

The tool sends a request body of `{"publishAsTemplate": ..., "requireAuthentication": ...}`, but neither field exists on the Forms API `SetPublishSettingsRequest`. The API expects a `publishSettings.publishState` object (`isPublished`, `isAcceptingResponses`) plus an `updateMask`, so the tool can never succeed as written.

## Fix

Build the documented request body and expose the real settings as parameters:

```python
{
    "publishSettings": {
        "publishState": {
            "isPublished": is_published,
            "isAcceptingResponses": is_accepting_responses,
        }
    },
    "updateMask": "publishState.isPublished,publishState.isAcceptingResponses",
}
```

The previous `publish_as_template` / `require_authentication` parameters did not map to any real API field, so they are replaced by `is_published` and `is_accepting_responses` (both default `True`, i.e. publish the form and accept responses). This changes the tool's parameter contract, but the old parameters could never have worked.

## Tests

Added two unit tests covering the request body shape and the defaults. `ruff check` is clean and the forms test module passes (11/11).

Reference: https://developers.google.com/workspace/forms/api/reference/rest/v1/forms/setPublishSettings

## Summary by CodeRabbit

* **Feature Updates**
  * Updated the form publishing settings tool parameters to use `is_published` and `is_accepting_responses` (defaulting to true), improving control over whether a form is published and whether it accepts responses.

* **Documentation**
  * Refreshed the tool reference to document the new parameters and their defaults.

* **Tests**
  * Added unit tests covering both explicit and default publishing/accepting-response behaviors, including validation of the request payload and success messaging.


taylorwilsdon:
Thanks!

PR #871, opened 2026-06-21
feat(drive): remote-safe uploads via resumable URLs for create/update

xiaoyang24601:
## Summary

Makes `create_drive_file` and `update_drive_file` **transport-mode aware** so file uploads behave correctly when the server is reached remotely (`streamable-http`), and adds a way to upload binary/large files without routing the bytes through the server.

### The problem
Today the upload inputs assume the client and server share a filesystem:
- `create_drive_file(fileUrl=…)` — `file://` reads the **server's** filesystem, not the caller's; the http(s) branch makes the server fetch arbitrary URLs (SSRF surface).
- `update_drive_file(file_path=…, file_url=…)` — same thing for content replacement.

On a remote deployment that's misleading to an agent: it points a tool at "a file" that only exists on the agent's machine.

### The change
- **Remote-mode guard.** In `streamable-http` mode, `fileUrl` (create) and `file_path`/`file_url` (update) are rejected with an actionable error. **`stdio` mode is unchanged** — local file ingestion still works there.
- **Resumable upload URLs.** Both tools gain `return_upload_url=True`, which initiates a [Drive resumable upload session](https://developers.google.com/workspace/drive/api/guides/manage-uploads#resumable) and returns the pre-authorized session URL. The caller then `PUT`s the bytes **straight to Google with no `Authorization` header** — the payload never flows through the server (or the model context).
  - `create_drive_file` → `POST` (new file)
  - `update_drive_file` → `PATCH` (replace an existing file's content, after applying any metadata changes)
- Inline text `content` and all of `update_drive_file`'s metadata operations (rename / move / star / trash / sharing / properties) are unaffected in every mode.
- Supports shared drives (`supportsAllDrives=true`). Optional `content_length` is forwarded as `X-Upload-Content-Length`.

Backward compatible — the new parameters default off.

## Tests

Adds unit tests for the resumable-URL path (create `POST` + update `PATCH`, headers/metadata/returned URL) and the remote-mode rejection of `fileUrl`/`file_url`. Full `tests/gdrive/test_drive_tools.py` passes (82 tests).

## Summary by CodeRabbit

* **New Features**
  * Added support for resumable Google Drive uploads by returning pre-authorized resumable session instructions (with optional content length), for both creating new files and replacing existing file content.
  * Added safer behavior for remote/streamed scenarios by rejecting incompatible inputs (e.g., inline content/file URLs) when resumable session mode is used.

* **Tests**
  * Added async test coverage for resumable session initiation and returned upload/session URLs, including validation of argument combinations and remote-mode rejection cases.

### DIFF
--- a/gdrive/drive_tools.py
+++ b/gdrive/drive_tools.py
@@ -5,6 +5,7 @@ This module provides MCP tools for interacting with Google Drive API.
 """
 
 import asyncio
+import json
 import logging
 import io
 import base64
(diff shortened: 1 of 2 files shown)


123andy:
Cherry-picked this onto a downstream fork and I've been running it in a stateless, `streamable-http` Docker deployment — the `return_upload_url` flow works exactly as intended: the client PUTs bytes straight to Google, nothing base64'd through the model context or the server. Really nice to have, and it lines up with where the spec is heading (the pre-authorized-upload shape in MCP SEP-2631). Two small notes from adopting it:

1. **Mutual exclusion vs. the merged `base64_content`.** Since #869 landed `base64_content` on `create_drive_file`, `return_upload_url` should be rejected alongside it too (not just `content`/`fileUrl`) — otherwise a caller could pass both. One-liner in the guard:
   ```python
   if content is not None or fileUrl is not None or base64_content is not None:
       raise ValueError("return_upload_url cannot be combined with 'content', 'fileUrl', or 'base64_content'.")
   ```
2. **`service._http` for the session init.** `_initiate_resumable_upload_session` uses the private `service._http.request(...)` and reads httplib2-style `(response, content)` / `response.get("location")`. That's undocumented internals whose type (httplib2 vs. `google-auth` transport) varies by client-library version, so it can break on a bump. A more stable equivalent is `google.auth.transport.requests.AuthorizedSession(creds).post(endpoint, headers=…, data=metadata)` and reading `resp.headers["Location"]`, with `creds` from the auth layer rather than the Resource's transport.

Neither is a blocker — this is working well for me in production. Happy to open a small follow-up PR against this branch with those two if useful.

123andy:
Thanks for this work. #871 and #872 conflict with current `main` and with each other, so they have been carried forward as one PR, #1160, with `Co-authored-by` credit on the feature commit. It keeps the design here, adds the `base64_content` guard and a few update-path fixes, and was tested against live Drive. It is offered when local files are disabled via `WORKSPACE_MCP_DISABLE_LOCAL_FILES` (#1036), and review there would be welcome.

PR #872, opened 2026-06-21
feat(drive): remote-safe imports via resumable upload URLs

xiaoyang24601:
## Summary

Extends the remote-safe upload treatment from #871 to the `import_to_google_doc` / `import_to_google_sheets` / `import_to_google_slides` tools.

These tools ingest a source file **server-side** (`file_path` / `file_url`) and convert it to a native Google format. On a remote (`streamable-http`) deployment that's misleading: the path/URL resolves on the server, not the caller.

### Change
- In remote mode, `file_path`/`file_url` are rejected with an actionable error. **stdio mode is unchanged** — local ingestion still works.
- `return_upload_url=True` (with `source_format`) initiates a resumable `POST` session whose `body.mimeType` is the target Google Apps type and whose `X-Upload-Content-Type` is the source format. The caller PUTs the source bytes directly and **Drive performs the same conversion** — no bytes through the server or model context.
- Inline `content` is unchanged.

Backward compatible (new params default off); reuses the helpers introduced in #871.

> **Note:** stacked on #871 — it introduces the shared `_initiate_resumable_upload_session` / `_is_remote_transport` helpers and the transport-aware `create_drive_file`/`update_drive_file`. Please merge #871 first; the extra commit here will drop out once it lands.

## Tests
Adds tests for the resumable import URL (source → target conversion headers), the `source_format` requirement, and the remote-mode rejection. Full `tests/gdrive/test_drive_tools.py` passes (85 tests).

## Summary by CodeRabbit

* **New Features**
  * Added resumable upload URL support to Google Drive file creation, update, and import operations (`create_drive_file`, `import_to_google_doc`, `import_to_google_slides`, `import_to_google_sheets`, `update_drive_file`)
  * New `return_upload_url` and `content_length` parameters to enable pre-authorized upload workflows
  * Enhanced validation and transport mode protections for file operations

### DIFF
--- a/gdrive/drive_tools.py
+++ b/gdrive/drive_tools.py
@@ -5,6 +5,7 @@ This module provides MCP tools for interacting with Google Drive API.
 """
 
 import asyncio
+import json
 import logging
 import io
 import base64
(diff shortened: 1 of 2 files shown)


123andy:
Thanks for this work. #871 and #872 conflict with current `main` and with each other, so they have been carried forward as one PR, #1160, with `Co-authored-by` credit on the feature commit. It keeps the design here, adds the `base64_content` guard and a few update-path fixes, and was tested against live Drive. It is offered when local files are disabled via `WORKSPACE_MCP_DISABLE_LOCAL_FILES` (#1036), and review there would be welcome.

PR #873, opened 2026-06-21
feat(gmail): attach Google Drive files; guard local paths in remote mode

xiaoyang24601:
## Summary

Gives `send_gmail_message` / `draft_gmail_message` a way to attach files that works on a remote (`streamable-http`) deployment, where the existing `attachments[].path` (a server-local file) is misleading.

Gmail has no per-attachment upload resource — attachments are inline MIME parts of the whole RFC-822 message — so there's no resumable-URL equivalent here. Instead, let the caller reference a **Google Drive file** the server can already read on their behalf:

- **`attachments[].drive_file_id`** — the server downloads the file from Drive and attaches it as binary. Native Docs/Sheets/Slides are exported to PDF.
- **`attachments[].drive_file_id` + `as_link: true`** — the file is *not* attached; its share link is appended to the message body (handy for large files).
- **`attachments[].path`** is rejected in remote mode with an actionable error. **stdio mode is unchanged.** `content` (base64) and `url` attachments are unaffected.

Implemented via a shared `_resolve_drive_attachments` helper used by both tools (Drive service built from the caller's existing authorized transport). Backward compatible — existing attachment shapes keep working.

## Tests
New `tests/gmail/test_drive_attachments.py`: binary download, native export-to-PDF, `as_link` body insertion, and local-path handling (rejected remotely / passed through in stdio). The Gmail test suite passes (86 tests).

## Summary by CodeRabbit

* **New Features**
  * Added support for Google Drive-based attachments in Gmail, including file downloads and native document export to PDF.
  * New option to attach Drive files as shareable links directly in email body instead of as MIME attachments.
  * Path-based attachments now restricted in remote transport modes for added security.

* **Tests**
  * Added comprehensive test coverage for Drive attachment handling and transport mode safeguards.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -14,7 +14,7 @@ import mimetypes
 import html
 from html.parser import HTMLParser
 from pathlib import Path
-from typing import Annotated, Optional, List, Dict, Literal, Any
+from typing import Annotated, Optional, List, Dict, Literal, Any, Tuple
 from urllib.parse import unquote, urlparse, urlunsplit
 
 from email.message import EmailMessage
@@ -26,6 +26,7 @@ from mcp.types import ToolAnnotations
 
 from pydantic import Field
 from googleapiclient.errors import HttpError
+from googleapiclient.discovery import build
 
 from auth.service_decorator import require_google_service
 from core.attachment_storage import get_attachment_storage, STORAGE_DIR
@@ -33,6 +34,7 @@ from core.config import (
     WORKSPACE_EXTERNAL_URL,
     WORKSPACE_MCP_BASE_URI,
     WORKSPACE_MCP_PORT,
+    get_transport_mode,
 )
 from core.http_utils import ssrf_safe_stream
 from core.utils import (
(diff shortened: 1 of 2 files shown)


123andy:
Adopted this on a downstream fork, specifically for the **draft** path — composing a draft with a `drive_file_id` attachment on a Drive-capable profile works great and neatly avoids base64-through-context for Gmail attachments (where, as you note, there's no resumable-upload equivalent). Three findings from running it:

1. **HTML injection in `_append_drive_links_to_body`.** The `as_link` HTML branch interpolates the Drive filename and URL straight into markup:
   ```python
   f'<li><a href="{ln["url"]}">{ln["name"]}</a></li>'
   ```
   A Drive file can be named e.g. `"><script>…`, so a maliciously-named file (shared to the sender) could inject HTML into the recipient's mail client. Worth `html.escape(ln["name"])` and `html.escape(ln["url"], quote=True)`.
2. **Implicit Drive scope via `service._http`.** `build("drive", "v3", http=service._http)` reuses the Gmail service's transport, so the Drive read relies on whatever scopes the Gmail token *happens* to carry. That works where the profile has broad Drive access but 403s on a minimal/send-only profile — and the scope requirement isn't stated anywhere. Declaring it with `require_multiple_services([... {"service_type": "drive", "scopes": "drive_read", "param_name": "drive_service"}])` would inject a properly-scoped `drive_service`, make the Drive-read dependency explicit, and drop the private-attribute dependency (same `service._http` fragility as noted on #871).
3. Minor: `as_link` appends the existing `webViewLink` but doesn't grant access, so a recipient without permission gets "access denied" — might be worth a note in the docstring that sharing is the caller's responsibility.

Overall this is a genuinely useful addition — thanks for it. Glad to send a small PR with the escaping fix if that helps.

Issue #832, opened 2026-06-06
stdio: OAuth/attachment callback server bound eagerly at startup; un-reaped client subprocesses exhaust the 8000-8004 port range


arthur-bamboo:
## Environment
- workspace-mcp v1.21.1, Python 3.11.15, macOS (Apple Silicon)
- Transport: `--transport stdio`, legacy auth (not OAuth 2.1; `MCP_ENABLE_OAUTH21` unset)
- Spawned by Claude Code as a stdio MCP server: `claude mcp add … -- sh -c '… uv run main.py --tools gmail calendar drive --transport stdio'`

## Summary
In legacy stdio mode the server binds a local OAuth-callback port in the 8000–8004 range **at startup** and does not release it (the process does not exit) when the stdio connection closes. MCP clients spawn the server repeatedly — health checks (`claude mcp list`), session restarts — so orphaned `main.py` processes accumulate, each holding one port. After ~5, new spawns fail at startup:

```
Error: No available port in range [8000, 8001, 8002, 8003, 8004]; all in use. Set WORKSPACE_MCP_PORT to a different preferred port, or increase WORKSPACE_MCP_PORT_FALLBACK_COUNT.
```

From the client side this surfaces as `✗ Failed to connect`.

## Observed
`lsof -nP -iTCP:8000-8004 -sTCP:LISTEN` showed **5 LISTENing `main.py` processes**, ages 20–38 min, each bound to one of 8000–8004 — all from the same stdio command (accumulated across Claude Code health-checks + session restarts). Killing them frees the ports and the next spawn connects.

## Expected
A stdio-transport server should tie the OAuth-callback listener's lifecycle to the stdio session:
- (a) bind the callback port **lazily** — only when an auth flow is actually initiated — and release it on completion; and/or
- (b) exit cleanly on **stdin EOF** (client disconnect), releasing any bound ports.

As-is, every spawn (including health checks that never trigger auth) binds and leaks a port.

## Impact
Recurs in normal use: any client that health-checks or restarts (Claude Code does both) leaks one process + port per spawn, so the failure resurfaces roughly every ~5 spawns until the user manually kills stale processes.

## Workaround
```
lsof -nP -iTCP:8000-8004 -sTCP:LISTEN
kill <stale main.py PIDs>
```

(Root cause framed as a hypothesis from the symptom — I observed the orphaned listeners and the startup failure directly, but did not inspect the server's shutdown / callback-server code.)

[...]

PR #877, opened 2026-06-23
feat(auth): ephemeral OAuth callback port for stdio mode

seidnerj:
## Problem

In stdio mode the OAuth callback server binds a fixed port (`WORKSPACE_MCP_PORT`, default 8000) for the lifetime of the process. Since stdio spawns one subprocess per client session, multiple concurrent sessions of the same server (e.g. two MCP clients, or the same server configured twice) contend for that single port — only one wins, the others fail their callback and surface as broken. The port is also held even though, after the initial auth, stdio needs no inbound port at all (tokens refresh without a redirect; attachments return disk paths).

## Change

In stdio mode, when the callback port is **not** explicitly pinned, defer binding until an auth flow actually starts, then bind a fresh **OS-assigned ephemeral port** (`bind` to port 0) and release it as soon as the callback resolves — or after a 5-minute watchdog timeout if the flow is abandoned (browser/tab closed without authenticating). Desktop/PKCE OAuth clients accept any loopback redirect port, so this is transparent to them.

Pinning `WORKSPACE_MCP_PORT`/`PORT` (Web/confidential clients, whose redirect URI must match a registered port) preserves the existing fixed-port behavior unchanged.

### Concurrency semantics

- **Inter-session:** each stdio process binds its own ephemeral port, so concurrent sessions never collide; an abandoned flow's port is reclaimed by that session's own watchdog and never blocks another.
- **Intra-session:** a single listener at a time. Re-triggering auth before timeout tears down the old listener (freeing its port), binds a new ephemeral port, and arms a fresh watchdog — never reused-in-place, never parallel.
- **Release triggers:** callback success (immediate) and the watchdog timeout are the only two; a closed tab produces no signal (the OAuth page is Google's origin), so the timeout is the catch-all.

## Tests

Adds `tests/auth/test_oauth_ephemeral_port.py` (pinned-detection snapshot, ephemeral acquisition, ensure-stdio wiring for pinned/unpinned/noop, callback teardown, watchdog arming, and an end-to-end serve-then-release). Existing fixed-port tests updated to pin the snapshot. All green.

## Compatibility

No behavior change for pinned-port / Web-client / streamable-http setups. Streamable-http continues to use the main FastAPI server for callbacks.

Related: builds on the lazy-binding work in #842 (which resolved the eager-startup-bind port-exhaustion issue #832). That change made the callback bind lazy/on-demand but still bound the fixed `WORKSPACE_MCP_PORT` and held it for the process lifetime; this PR makes that bound port ephemeral and releases it once the callback resolves.

## Summary by CodeRabbit

- **New Features**
  - OAuth sign-in using standard input/output can now automatically use an available temporary callback port.
  - Callback ports are published and configured automatically during authentication.
  - Temporary callback servers shut down after completing authentication or becoming idle.

- **Bug Fixes**
  - Fixed behavior so explicitly configured callback ports continue to be honored.
  - Improved callback validation and cleanup when authentication finishes or encounters an error.


seidnerj:
Rebased onto `main` at v1.25.0. I merged upstream in rather than force-pushing, so the existing review history stays intact. CI is green: pytest, ruff, build-and-push.

The only conflict was the env-var table in the README, which the readme consolidation removed. Rather than re-add it, I dropped the `WORKSPACE_MCP_PORT` notes from the diff, so this is now 5 files, all in the OAuth callback path.

No rush on my end. Happy to rebase again whenever it is convenient to look at.

Issue #123, opened 2025-07-26
Pass credentials to the MCP


tuanchris:
Awesome project! Thank you for the great work @taylorwilsdon. 

I'm building an application that has a Google SSO login before interacting with the LLM. What do you think is the best way to pass in credentials to the MCP so the user doesn't have to authenticate twice?

taylorwilsdon:
https://github.com/taylorwilsdon/google_workspace_mcp/issues/112 < I suspect this is what you're looking for!

taylorwilsdon:
This is in the dev branch now @tuanchris https://github.com/taylorwilsdon/google_workspace_mcp/pull/137 hope to have it in the wild shortly.

tuanchris:
@taylorwilsdon thank you for getting back! I'll keep an eye out for the release & do some tests when it eventually does.

taylorwilsdon:
It's live!

samuellusandi:
Thanks for your efforts!

I would like to ask a few questions related to this, I'm trying to get this to work with my IDEs or any agentic clients:

1. VSCode utilizes the Dynamic Client Registration faces the issue:
```
/oauth2/authorize?client_id={uuid}&response_type=code&state=vscode%3A%2F%2Fdynamicauthprovider...
```

And this error in Google Login page:
```
The OAuth client was not found.
If you are a developer of this app, see [error details](https://accounts.google.com/).
Error 401: invalid_client
```

Expected since VSCode will always put a dynamic client ID unique to it, following its dynamic registration method.

2. Attempting to use Windsurf using `mcp-remote` seems to yield `failed to get tools: Invalid request parameters`. It could just be Windsurf's issue, as Cursor has no such issue, but I initially thought it could just work out of the box with any MCP Client!
<details>
<summary>Screenshot</summary>

<img width="436" height="36" alt="Image" src="https://github.com/user-attachments/assets/155e8787-b3a3-4cae-bbef-75970bdc195c" />

</details>

3. Is it possible to ignore the parameter `user_google_email` altogether when OAuth2.1 is used?
The following screenshot shows that it doesn't know what email to use, when it *should* always use our own email when authenticated. Otherwise, we would use someone else's email (impersonation) that should be reserved only on the legacy authentication, correct? Tested this on MCP Inspector, and it's still required, when I should reliably expect once I log in, I will always act as that user!
<details>
<summary>Screenshot</summary>

[...]

PR #878, opened 2026-06-23
feat(gmail): Gmail-web compose fidelity for new/reply/forward (+attachments, inline & RTL) and opt-in SMTP transport

seidnerj:
## Summary

Makes mail composed and sent through the Gmail tools indistinguishable from mail submitted via the Gmail **web** interface - structurally (MIME) and visually - across **new messages, replies, and forwards**, including attachments and inline images. Also adds an opt-in SMTP send transport that submits the same raw MIME over `smtp.gmail.com` instead of the Gmail API.

### Compose/send fidelity
- **MIME** mirrors Gmail web: `multipart/alternative` (plain + typed `<div dir="ltr">`/`<div dir="rtl">` HTML), Gmail-format boundary (`000000000000` + 16 hex), UTF-8, and **content-dependent per-part transfer encoding** (quoted-printable for mostly-ASCII, base64 for heavily-non-ASCII).
- **Base text direction** is auto-detected per body via the Unicode bidi first-strong-character rule: right-to-left content is wrapped `<div dir="rtl">` so Gmail right-aligns it; left-to-right output stays byte-identical. An explicit `direction` param (`auto`|`ltr`|`rtl`, default `auto`) on `send_gmail_message`/`draft_gmail_message` overrides detection. (+17 tests)
- **Replies** reproduce Gmail's `gmail_quote_container` / `gmail_attr` "On … wrote:" attribution + `blockquote` trail.
- **Forwards** reproduce the `---------- Forwarded message ---------` header block (with `gmail_sendername`), no blockquote, non-`>`-quoted plain text.
- **Attachments & inline images** use Gmail's nesting: `multipart/mixed > multipart/related > multipart/alternative` with inline `Content-ID` image parts.
- All locked against **golden structural skeletons captured from real Gmail-web mail** (PII-safe; `tests/gmail/fixtures/`).

### Display-name resolution
Recipient/sender names resolve as Gmail web does: saved Contacts → thread/sender-supplied → Other contacts → Workspace directory → bare. The sender's `From` name comes **only** from the matching Send-As entry.

### Opt-in SMTP transport (`GMAIL_SEND_TRANSPORT`)
- `api` (default) - the standard Gmail-API send path, unaffected by the SMTP feature: the dispatch layer is a pure pass-through, so the API path's bytes are exactly the fidelity builders' output above. SMTP is strictly opt-in.
- `smtp` submits the same message over `smtp.gmail.com:587` (STARTTLS + XOAUTH2); auto-requests the `https://mail.google.com/` scope (re-auth required) and **gracefully falls back to the API** + a result note if that scope is absent.
- Governs new/reply/forward sends only; **drafts and all reads stay API-only**. On the SMTP path, Bcc is set via envelope (header omitted) and a unique `Message-Id` is authored so the sent message can be located for the result string.
- **Send-transport fingerprint (verified live):** body/MIME is byte-identical on both transports, but the server-stamped envelope differs. **Neither** path is byte-identical to the web UI's `Received` headers - both add an origin hop the web composer lacks. `api` keeps the real `@mail.gmail.com` Message-ID and never exposes your IP (its extra hop names `gmailapi.google.com`); `smtp` carries an **authored `<…@sender-domain>` Message-ID** (not `@mail.gmail.com`) **and leaks your client IP/ISP** to recipients. Prefer `api`. (See the README fingerprint table.)

## Graceful scope fallback

All People API surfaces are **optional**. When `contacts.readonly`, `contacts.other.readonly`, or `directory.readonly` are absent - or a tier returns 401/403 - name resolution falls back to bare-address behavior and the tool result reports which scope to add. The SMTP path degrades to the API the same way when `mail.google.com` is missing.

## Verification

- **Unit/structural:** 1460 passing - web-fidelity, MIME, reply/forward, attachments/inline, per-part CTE, transport wiring, SMTP (XOAUTH2 + authored-Message-Id lookup), scope, and golden-fixture comparison - plus 17 new tests covering RTL base-direction detection, wrapping, and the send/draft/forward tool paths.
- **Live round-trip (own inbox, raw RFC822 vs golden):** new, reply, forward, attachment+inline, and SMTP all sent and verified - reply/forward/inline match the golden skeletons. Live testing surfaced and fixed bugs (forwarded `To:` inserted as raw HTML → escaped + mailto-linked; non-ASCII `Subject:` emitted raw → RFC 2047-encoded; SMTP message-id lookup was subject-only → now authors a unique `Message-Id` and matches on it via `rfc822msgid:`) and characterized the send-transport fingerprint: faithful body/MIME on both, but neither is byte-identical in `Received` - `api` keeps the Gmail Message-ID and leaks no IP, `smtp` uses an authored `<…@sender-domain>` Message-ID and exposes the client IP.

## Update: reply subject inheritance (thread-safety across clients)

`subject` is now **optional when replying** (`thread_id`/`in_reply_to`). When omitted, the reply inherits the parent message's **exact** subject, adding a single `Re:` only if none is present - and recognizing an existing `Re:`/`RE:` even behind leading list/ticket tags (e.g. `[list] Re: ...`), so it never doubles the prefix and never strips tags like `[list]`/`[#123]`.

**Why:** RFC threading (`In-Reply-To`/`References`) was already correct and Gmail groups by the References chain - but subject-sensitive clients (Spark, Apple Mail, Outlook) group by normalized subject, so a rewritten/hand-authored reply subject would split the thread there even though it looked fine in Gmail. Inheriting the parent subject keeps replies in-thread everywhere.

Implementation: a new idempotent, tag-aware `normalize_reply_subject()` helper; subject inheritance + prefixing centralized in the compose path (replacing three ad-hoc `startswith("re:")` checks that could double the `Re:` on tagged parents). New unit + integration tests (helper cases; send + draft reply-with-omitted-subject, including the tagged/no-double-`Re:` case); verified end-to-end against live Gmail.

### Note on RTL rendering in non-conforming clients

The `<div dir="rtl">` output is correct and renders right-to-left in Gmail web/mobile. Some clients (e.g. Spark iOS) ignore `dir` on a wrapperless HTML fragment and render RTL text left-aligned - a client-side limitation that equally affects real Gmail-web-composed RTL mail, since Gmail web emits the same bare-fragment structure. Matching Gmail-web output and fixing those clients are mutually exclusive, so this PR keeps fidelity; the `direction` docstring was corrected to describe this accurately (it previously mis-attributed the effect to the plain-text part).

## Summary by CodeRabbit

* **New Features**
  * Added Gmail “web faithful” compose/send/draft/forward with improved quote/attribution rendering (including RTL), deterministic attachment handling, and direction control.
  * Added Gmail web-faithful MIME/HTML helper utilities.
  * Introduced configurable Gmail send transport (`api` or `smtp`) with scope-aware routing and clearer fallback messaging.
* **Bug Fixes**
  * Optional multi-service authentication failures now only suppress authentication errors; non-auth errors correctly surface.
  * Tightened Forms publish settings updates to the intended fields.
* **Documentation**
  * Documented `GMAIL_SEND_TRANSPORT` behavior and transport differences.
* **Tests**
  * Expanded web-fidelity, MIME construction, transport routing, and golden fixture coverage.


seidnerj:
@coderabbitai full review

seidnerj:
@coderabbitai review

seidnerj:
@coderabbitai review

seidnerj:
@coderabbitai review

seidnerj:
@coderabbitai review

seidnerj:
@coderabbitai review

seidnerj:
@coderabbitai review

seidnerj:
@coderabbitai review

seidnerj:
@coderabbitai full review

seidnerj:
@coderabbitai review

seidnerj:
@coderabbitai resume

seidnerj:
@coderabbitai full review

seidnerj:
@coderabbitai full review

seidnerj:
@coderabbitai full review

seidnerj:
@coderabbitai review

seidnerj:
Merged current `main` (v1.25.0) in and reconciled with the Gmail work that landed since: #1004, #917, #902, #889. CI is green.

Where both sides had implemented the same thing, this branch now defers to yours: it drops its own `_parse_message_id_chain` and `_derive_reply_headers` in favor of the `gmail_helpers.py` versions, and `send_gmail_message` unions #1004's `reply_all` derivation with the send-as fetch it already had.

Still a large diff. Happy to split it into the MIME builder, attachments and inline, RTL, and the SMTP transport if that is easier to review.

PR #879, opened 2026-06-23
fix(core): make credentials-dir permission check concurrency-safe

seidnerj:
## Summary

Fixes a startup race when two MCP server processes initialize the same credentials directory at once (e.g. a personal and a work google-workspace server launched together by the same client).

## The race

The old `check_credentials_directory_permissions`:
- branched on `os.path.exists`,
- wrote a **shared** `.permission_test` filename, and
- on failure removed the directory with `os.rmdir`.

When two processes run this concurrently against the same `~/.google_workspace_mcp/credentials`, one process's cleanup/remove can yank the directory or the shared probe file out from under a sibling. The sibling's write then fails with `ENOENT` and startup crashes — the server shows as failed / "Connection closed".

## The fix

Replace it with `makedirs(exist_ok=True)` plus a **unique** `tempfile.NamedTemporaryFile` probe, and remove the `rmdir`. Concurrent initializers no longer race, and existing credentials are never deleted. The existing success log line is preserved so log formatting still matches.

## Verification

`tests/core/test_credentials_dir_permissions.py` — 3 passing, including a 24-thread concurrent-init test that reproduces the original crash, plus assertions that the check preserves an existing dir + contents and leaves no probe files behind.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credentials directory permission checks to be safe when multiple processes start at the same time.
  * The check now avoids deleting or modifying existing contents and no longer leaves temporary probe files behind.
  * Better error reporting when directory creation or access fails, including the full path in the message.

* **Tests**
  * Added regression coverage for concurrent initialization, existing directories, and nested missing paths.


taylorwilsdon:
Thanks!

PR #885, opened 2026-06-26
feat: add generic api_call tool for arbitrary Google API endpoints

### LINKED ISSUE #884: [feat request] api_call tool for arbitrary Google API endpoints
**Is your feature request related to a problem? Please describe.**
There's no way to hit a Google API endpoint that doesn't have a dedicated tool yet. For agent use cases where the goal is "do anything against my account," any missing endpoint/param/service is a dead end until someone adds a tool. `run_script_function` can technically do it, but standing up an Apps Script project + deployment for a one-shot REST call is overkill.

**Describe the solution you'd like**
An `api_call` tool: authenticated request to an arbitrary `*.googleapis.com` URL using the existing user credentials. Inputs: `method`, `url`, `query`, `body`. Output: status + JSON. It reuses the current OAuth token, so it can only do what the granted scopes already allow — no new auth surface.

**Describe alternatives you've considered**
- `run_script_function` — works, too heavy for one call.
- Separate single-tool MCP for raw calls — duplicates the OAuth/credential handling this server already does.
- #498 (code mode) would also enable arbitrary calls, but that's a bigger rewrite; a single `api_call` tool is a much smaller step.

**Additional context**
Bounded by granted scopes. Willing to implement.

domdomegg:
Closes #884.

Adds a single `api_call` tool — an escape hatch for any Google API operation that does not have a dedicated tool (long-tail endpoints, parameters, and services). The motivation is agent use cases where the goal is "do anything against my account": today any missing endpoint is a dead end until a tool is added.

### How it works
- Reuses the **authorized transport from the existing service-decorator auth path** — no new auth surface. The request is made with the user's existing credentials, so it is **bounded entirely by the scopes they have already granted**; it cannot do anything they have not consented to.
- Authenticates against the **base identity scope** (`userinfo.email`) rather than an unrelated service scope, so it is not coupled to any one service and always loads.
- **Restricted to https `*.googleapis.com` hosts** — the OAuth token must not be sent anywhere else. Responses are pretty-printed and length-capped.
- Registered **unconditionally**, independent of `--tool-tier` / `--permissions`.

### Signature
```
api_call(method, url, query?, body?) -> "HTTP <status>\n<body>"
```
e.g. `GET https://gmail.googleapis.com/gmail/v1/users/me/labels`.

### Tests
Unit tests cover method/URL validation, host allow-listing, query merging, body serialization, response formatting/truncation, and an end-to-end call against a mocked transport. `ruff` clean.

### Notes / open to direction
This is intentionally **ungated** (always on). I'm happy to add any gating you'd prefer — an opt-in env flag, restricting it to the `complete` tier, audit logging, or a read-only variant — just let me know what fits the project. Also glad to adjust the credential approach if you'd rather it hang off a dedicated accessor than the identity scope.

## Summary by CodeRabbit

* **New Features**
  * Added an authenticated Google API Call tool for supported HTTPS Google endpoints, with query and body inputs.
  * Added configurable tool disabling through startup options and environment settings.
  * Improved startup displays with clearer configuration, service status, notices, and readiness information.

* **Bug Fixes**
  * Strengthened request validation, response formatting, truncation, and sensitive URL redaction.
  * Ensured the API Call tool remains available unless explicitly disabled.

* **Tests**
  * Added comprehensive coverage for API requests, responses, redaction, read-only behavior, and tool filtering.

### DIFF
--- a/core/tool_registry.py
+++ b/core/tool_registry.py
@@ -18,6 +18,13 @@ logger = logging.getLogger(__name__)
 # Global registry of enabled tools
 _enabled_tools: Optional[Set[str]] = None
 
+# Tools registered regardless of tier or --tools selection. api_call is the
+# generic escape hatch: it belongs to no service and appears in no tier, and is
+# bounded by the scopes the user has already granted rather than by selection.
+# The explicit block list (--disabled-tools) still removes it: that is
+# subtractive and wins over every form of selection.
+ALWAYS_ENABLED_TOOLS: Set[str] = frozenset({"api_call"})
+
 # Tools explicitly blocked regardless of tier or permission selection
 _disabled_tools: Set[str] = set()
 
@@ -68,7 +75,7 @@ def is_tool_enabled(tool_name: str) -> bool:
     """Check if a specific tool is enabled."""
     if _enabled_tools is None:
         return True  # All tools enabled by default
-    return tool_name in _enabled_tools
+    return tool_name in _enabled_tools or tool_name in ALWAYS_ENABLED_TOOLS
 
 
 def conditional_tool(server, tool_name: str):
--- a/fastmcp_server.py
+++ b/fastmcp_server.py
@@ -157,6 +157,7 @@ import gforms.forms_tools
 import gslides.slides_tools
 import gtasks.tasks_tools
 import gsearch.search_tools
+import gapi.api_tools
 
 # Configure tool registration
 wrap_server_tool_method(server)
--- a/gapi/api_tools.py
+++ b/gapi/api_tools.py
@@ -131,8 +131,7 @@ async def _api_call_impl(
     # httplib2's request() is synchronous, so run it off the event loop.
     authorized_http = service._http
     response, content = await asyncio.to_thread(
-        authorized_http.request,
-        url, method=method, body=request_body, headers=headers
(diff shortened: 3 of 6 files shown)


domdomegg:
Thanks @coderabbitai — addressed in f11703d:

1. **Log leaks query params** — fixed. The log line now uses a `_redact_url()` helper that keeps scheme/host/path and drops the query string, so the method + endpoint stay visible but any token-bearing query params don't hit the logs.

2. **Blocking call on the event loop** — fixed. The synchronous `httplib2` request now runs via `asyncio.to_thread(...)`, so the network round-trip no longer stalls other MCP traffic.

3. **Unconditional registration bypasses read-only / permissions** — partially addressed, by design. `api_call` is intended to be an always-available escape hatch, so I've kept it registered unconditionally rather than gating it behind service selection. But the valid safety concern — that it shouldn't be able to mutate in read-only mode — is now enforced *in the tool*: when `is_read_only_mode()` is set, `api_call` rejects POST/PUT/PATCH/DELETE and only allows GET. Granular per-service `--permissions` are intentionally not applied, since by definition this tool isn't scoped to one service; access remains bounded by the OAuth scopes the user has actually granted.

Happy to adjust if you'd prefer it gated differently (e.g. an opt-in env flag or `complete`-tier only) — let me know what fits the project.

PR #889, opened 2026-06-26
fix: attach reply drafts to existing thread via message.threadId

kenjimax:
## Bug

`draft_gmail_message` creates drafts in a brand-new Gmail thread when `thread_id` is provided, instead of appearing inline in the original conversation.

**Root cause:** The draft create body omits `message.threadId`:

```python
# comment claimed this causes "UI-hidden drafts"
draft_body = {"message": {"raw": raw_message}}
```

Gmail threads drafts by `threadId`, not by `In-Reply-To`/`References` headers. Without `threadId`, Gmail ignores the headers for threading and opens the draft as a standalone conversation.

## Fix

Set `message.threadId` when `thread_id` is provided:

```python
draft_body = {"message": {"raw": raw_message}}
if thread_id:
    draft_body["message"]["threadId"] = thread_id
```

The prior concern about "UI-hidden drafts" is not reproducible. Verified: drafts created with matching `threadId` and `In-Reply-To` headers appear correctly inline in Gmail's conversation view and are not hidden.

## Verification

Tested against live Gmail API. Draft with `threadId` set appears inline in the original thread; draft without `threadId` opens in a new standalone thread.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Gmail draft replies to include the conversation `threadId` in the raw draft payload when all necessary threading details are available.
  * Strengthened draft-threading behavior to ensure replies stay grouped correctly, and that drafts omit `threadId` when thread metadata can’t be determined (e.g., fetch failure or empty threads).


taylorwilsdon:
Thanks!

PR #892, opened 2026-06-29
feat(sheets): add data validation tools (set/clear + read dropdowns)

emilio-S03:
## What
Adds two complementary Google Sheets data-validation (dropdown) tools:
- `manage_data_validation` — set or clear validation on a range
- `read_data_validation` — read back the validation rule on each cell (read-only)

## Why
The server can format cells (`format_sheet_range`) and manage conditional formatting (`manage_conditional_formatting`), but there was no way to add a data-validation dropdown, nor to read one back to verify it. These fill that gap and pair naturally: set a dropdown, then confirm it.

## Behavior
**`manage_data_validation`** (write)
- `action="set"` + `values=[...]` → fixed-list dropdown (`ONE_OF_LIST`)
- `action="set"` + `source_range="Lists!A1:A20"` → range-sourced dropdown (`ONE_OF_RANGE`; a leading `=` is added if omitted)
- `action="clear"` → removes validation on the range
- Options: `strict` (reject vs. warn), `show_dropdown` (the chip / `showCustomUi`), `input_message`

**`read_data_validation`** (read-only)
- Inspects a range and returns, for each cell that has a rule: A1 address, condition type, the list options or source range, `strict`, and whether the dropdown chip is shown.

## Implementation notes
- Mirrors existing patterns: split `_impl` + decorated tool, `_parse_a1_range` / `_format_a1_cell` helpers, `asyncio.to_thread(...)`.
- `read_data_validation` uses `spreadsheets.get` with `fields=…dataValidation…` (sheets **read** scope only); `manage_data_validation` uses the existing sheets **write** scope. No new OAuth scopes.
- Both registered under `sheets` → `complete` tier in `core/tool_tiers.yaml`.
- Adds `tests/gsheets/test_manage_data_validation.py` and `tests/gsheets/test_read_data_validation.py`.

## Testing
`uv run pytest tests/gsheets/test_manage_data_validation.py tests/gsheets/test_read_data_validation.py` → 7 passed.

## Summary by CodeRabbit

* **New Features**
  * Added support for managing Google Sheets dropdown/data-validation rules.
  * Added a way to inspect existing validation rules on cells, including type, allowed values, and settings.

* **Bug Fixes**
  * Improved handling of validation inputs so only valid combinations are accepted.
  * Clearer cell-level reporting when checking validation across a range.

* **Tests**
  * Added coverage for creating, clearing, and reading validation rules.

### DIFF
--- a/core/tool_tiers.yaml
+++ b/core/tool_tiers.yaml
@@ -96,6 +96,7 @@ sheets:
     - manage_spreadsheet_comment
     - manage_conditional_formatting
     - manage_data_validation
+    - read_data_validation
 
 chat:
   core:
--- a/core/tool_tiers.yaml
+++ b/core/tool_tiers.yaml
@@ -95,6 +95,7 @@ sheets:
     - list_spreadsheet_comments
     - manage_spreadsheet_comment
     - manage_conditional_formatting
+    - manage_data_validation
 
 chat:
   core:
--- a/gsheets/sheets_tools.py
+++ b/gsheets/sheets_tools.py
@@ -26,6 +26,7 @@ from gsheets.sheets_helpers import (
     _fetch_detailed_sheet_errors,
     _fetch_grid_metadata,
     _fetch_sheets_with_rules,
+    _format_a1_cell,
     _format_conditional_rules_section,
     _format_sheet_error_section,
     _parse_a1_range,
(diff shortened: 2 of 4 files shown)


alexnbg88:
Ran into the same gap and built something overlapping locally, so flagging it here rather than opening a competing PR.

Two things you might want to fold in:

**1. Condition coverage.** This handles `ONE_OF_LIST` and `ONE_OF_RANGE`. The API supports 25 condition types for data validation, so checkboxes (`BOOLEAN`), `CUSTOM_FORMULA`, `TEXT_IS_EMAIL`/`TEXT_IS_URL`, `NUMBER_BETWEEN` and the `DATE_*` family currently aren't reachable.

**2. A trap worth encoding.** The data-validation condition set is *not* the conditional-formatting set already in `CONDITION_TYPES`. `BLANK`, `NOT_BLANK`, `TEXT_STARTS_WITH` and `TEXT_ENDS_WITH` are valid for conditional formatting but rejected by `setDataValidation` with an opaque 400, so reusing `CONDITION_TYPES` here would be a subtle footgun. Value counts differ per condition too (`NUMBER_BETWEEN` needs exactly 2, `TEXT_IS_EMAIL` none), so validating up front turns a 400 into an actionable message.

I have both implemented with tests and can open a follow-up PR on top of this branch, or just hand over the code, whichever you prefer. Your `read_data_validation` tool and the `=` auto-prefix on `source_range` are both nicer than what I did.

Happy either way, this is your PR.

PR #897, opened 2026-07-01
perf: defer GoogleProvider/Flow/external-oauth-provider imports until actually used

### LINKED ISSUE #896: stdio cold-start: unconditional imports pull in the full OAuth-proxy + google-auth-oauthlib stacks even in legacy OAuth 2.0 mode
## Summary

Every `workspace-mcp` process spawn — even `--help`, even `--transport stdio` with legacy OAuth 2.0 and already-cached, valid credentials — unconditionally imports (and in one place instantiates) modules that are architecturally only reachable from FastMCP's OAuth 2.1 proxy mode or from starting a brand-new interactive OAuth flow. For MCP hosts that spawn `workspace-mcp` fresh per session rather than running it as a persistent daemon (Claude Desktop, Claude Code stdio, and any other `uvx`-per-session integration), this import cost is paid on **every single spawn**, not once.

## Context

I maintain a small MCP host ([pkb-mcp](https://github.com/agenthill/personal-knowledge-base)) that spawns `workspace-mcp` via `uvx ... workspace-mcp --transport stdio --tool-tier extended --permissions drive:full docs:full gmail:send slides:full` — legacy OAuth 2.0, no OAuth 2.1 proxy, no external OAuth provider. Investigating a user-visible "slow cold start" complaint, I profiled `python -X importtime -c "import main"` in a fresh venv (`workspace-mcp==1.21.1` from PyPI, cold bytecode cache) and found ~1.36s of cumulative import time attributable to modules that are unreachable in this configuration:

```
import time:       843 |      89086 |               google_auth_oauthlib.helpers
import time:      1204 |     711840 |             google_auth_oauthlib.flow
import time:       632 |     712472 |           google_auth_oauthlib.interactive
import time:       419 |     712891 |         google_auth_oauthlib
import time:        10 |     712900 |       google_auth_oauthlib.flow
import time:      5400 |    1363874 |     auth.google_auth
import time:      5648 |    2723458 |   core.utils
import time:      4743 |     647931 |             fastmcp.server.auth.oauth_proxy.proxy
import time:       385 |     648316 |           fastmcp.server.auth.oauth_proxy
import time:      1069 |     649654 |         fastmcp.server.auth.providers.google
import time:      4636 |     774929 |   core.server
import time:      4463 |    4430829 | main
```

(Absolute numbers are noisy across machines/cache-warmth — the two chains total ~1.36s here — but the *shape* is the actionable part: both chains are structurally unreachable in this config and are still paid for on every spawn.)

## Root cause

Three unconditional top-level imports, each pulling in a heavy chain that's only needed for a *different* configuration than legacy OAuth 2.0 / stdio:

1. **`core/server.py:38`** — `from fastmcp.server.auth.providers.google import GoogleProvider`. Only ever instantiated inside `configure_server_for_http()`'s "Standard OAuth 2.1 mode" branch (i.e. only when `is_oauth21_enabled()` and *not* external-provider) — but imported unconditionally at module load, which runs on every spawn regardless of transport or auth mode.
2. **`auth/google_auth.py:15`** — `from google_auth_oauthlib.flow import Flow`. Only used inside `create_oauth_flow()`, itself only called when starting a *new* interactive OAuth authorization — never when serving tool calls off already-cached, refreshed credentials (the steady-state case for any long-running integration).
3. **`auth/auth_info_middleware.py:12`** — `from auth.external_oauth_provider import get_session_time`. Importing `auth.external_oauth_provider` at all forces `fastmcp.server.auth.providers.google.GoogleProvider` to load too, because `ExternalOAuthProvider(GoogleProvider)` subclasses it at module scope — so this import alone re-triggers chain (1) via a completely separate path even if (1) is fixed in isolation. `get_session_time()` itself is only reached deep inside a conditional branch (`verified_auth` not already a `WorkspaceAccessToken`).

All three exist because `main.py`'s `_load_startup_dependencies()` imports `core.server` (which imports `auth.auth_info_middleware`, which imports `auth.external_oauth_provider`) unconditionally at module load — before `argparse` even runs, so it's paid even for `--help`.

## Fix (PR incoming)

Defer all three imports to inside the function/branch that actually uses them, module-global-cached so the cost is paid at most once per process (only if that branch is ever reached) — matching precedent already in this codebase (`core/config.py`'s `__getattr__`-based lazy `WORKSPACE_MCP_PORT`, and the existing local `from core.server import get_auth_provider` import already inside `auth_info_middleware.py`'s own `_process_request_for_auth`).

Details, verified locally:
- `GoogleProvider` / `Flow` are kept as real (if initially `None`) module attributes — not omitted — so existing `monkeypatch.setattr(module, "Name", Fake)` tests keep working unmodified. `Flow`'s existing `patch("auth.google_auth.Flow.from_client_config", ...)`-style tests needed their patch target updated to `google_auth_oauthlib.flow.Flow.*` directly, since patching a nested attribute (`Flow.from_client_config`) requires `Flow` to already be the real class at patch-setup time — a `None` sentinel can't satisfy that. `GoogleProvider` didn't need this since tests patch the whole name, not a sub-attribute.
- Verified via `sys.modules` after `import main` that none of `fastmcp.server.auth.providers.google`, `fastmcp.server.auth.oauth_proxy`, `google_auth_oauthlib`, or `auth.external_oauth_provider` load anymore in the default (legacy OAuth 2.0, stdio, non-external-provider) configuration. A full `python -X importtime main.py --help` module-graph diff (patched vs. unpatched) shows **278 modules imported only in the unpatched build, zero imported only in the patched one** — the win is bigger than just the two named chains above: the entire `authlib` + `joserfc` (JWT/JWE) + `key_value.aio`/`caio`/`aiofile` OAuth-proxy dependency tree, and the entire `google_auth_oauthlib` → `oauthlib`/`requests_oauthlib`/`wsgiref` tree, disappear entirely from the default config.
- Full existing test suite green: 1265 passed / 2 skipped, same single pre-existing failure in `tests/gforms/test_forms_tools.py::test_set_publish_settings_builds_publish_state_body` present on unmodified `main` too (unrelated to this change, not touched).
- `ruff check .` clean.

## Related

Adjacent to #832 / #842 — the eager OAuth/attachment-callback HTTP server bind in stdio mode, which #842 already fixed (lazy-started both the OAuth-callback and attachment-serving paths). That closed out the *other* half of stdio cold-start cost; this issue is the import-time half that's still open.

## Environment

- `workspace-mcp` @ `6e1d1457746777f8512f52d40fb195b2a40bad36` (`main`, matches the `v1.22.0` release); reproduces identically at `v1.21.1`
- Python 3.11 / 3.12, macOS (arm64)
- `uv` 0.11.11

szhygulin:
Closes #896

## Summary

Three unconditional top-level imports pull in the full FastMCP OAuth-proxy stack (`authlib`, `joserfc`, `key_value.aio`) and the full `google_auth_oauthlib` stack (`oauthlib`, `requests_oauthlib`) on **every** process spawn — even `--help`, even in the common legacy-OAuth-2.0/stdio configuration that never reaches any of the code paths that actually need them. This patch defers all three to lazy, module-global-cached imports, only paid if the owning branch is ever reached.

See #896 for the full profiling writeup and root-cause detail.

## What changed

- **`core/server.py`** — `GoogleProvider` is only ever instantiated inside `configure_server_for_http()`'s "Standard OAuth 2.1 mode" branch. Import deferred there.
- **`auth/google_auth.py`** — `Flow` is only used inside `create_oauth_flow()` (starting a *new* interactive auth flow — never when serving tool calls off cached, refreshed credentials). Import deferred there.
- **`auth/auth_info_middleware.py`** — importing `auth.external_oauth_provider` at all drags in `GoogleProvider` too (`ExternalOAuthProvider(GoogleProvider)` subclasses it at module scope) via a path independent of `core/server.py`'s own import — so fixing (1) alone doesn't help unless this one is fixed too. `get_session_time()` is only reached inside the non-`WorkspaceAccessToken` branch; import deferred right above its one call site (mirrors the existing local `from core.server import get_auth_provider` import already a few lines up in the same method).

Pattern used for all three: keep the name as a real module attribute initialized to `None` (`if TYPE_CHECKING: from ... import X else: X = None`), then at the point of use: `global X; if X is None: from ... import X`. This keeps every name a valid `monkeypatch.setattr(module, "Name", Fake)` target (matches the existing lazy-singleton-via-module-global convention already used elsewhere in this codebase — `auth/credential_store.py`, `auth/oauth_config.py`, `auth/scopes.py`, `core/tool_registry.py`, `auth/permissions.py`). `Flow`'s existing tests (`tests/auth/test_google_auth_pkce.py`) needed their `patch(...)` targets updated from `auth.google_auth.Flow.*` to `google_auth_oauthlib.flow.Flow.*` directly — patching a nested attribute (`Flow.from_client_config`) requires the class to already be resolvable at patch-setup time, which a `None` sentinel can't satisfy. `GoogleProvider`'s tests didn't need this since they patch the whole name, not a sub-attribute.

## Verification

- `sys.modules` check after `import main`: none of `fastmcp.server.auth.providers.google`, `fastmcp.server.auth.oauth_proxy`, `google_auth_oauthlib`, or `auth.external_oauth_provider` load in the default config.
- Full `python -X importtime` module-graph diff (patched vs. unpatched): **278 modules imported only in the unpatched build, zero imported only in the patched one.**
- `uv run --frozen pytest -q`: 1265 passed / 2 skipped, same single pre-existing failure (`tests/gforms/test_forms_tools.py::test_set_publish_settings_builds_publish_state_body`) present on unmodified `main` too — unrelated, not touched by this change.
- `ruff check .`: clean.

## Known trade-off

Because `GoogleProvider`/`Flow` are literally `None` until first real use, `typing.get_type_hints()` called on `create_oauth_flow`, `get_auth_provider`, or the containing modules *before* either lazy import has fired resolves the forward-ref annotation to `NoneType` silently. Currently inert — neither function is `@server.tool()`-decorated and the repo has no mypy/pyright/Sphinx step that would trip on it — but flagging it as an accepted trade-off of keeping these names monkeypatch-compatible, in case that changes later.

## Scope note

This is deliberately scoped to the import-time cost only. #832/#842 already fixed the *other* half of stdio cold-start (eager OAuth/attachment-callback HTTP server bind) — not revisited here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication startup behavior by deferring imports for Google OAuth components until an OAuth flow is actually created.
  * Preserved existing sign-in, token verification, and session timing behavior while making OAuth handling more reliable.

* **Tests**
  * Updated OAuth PKCE-related tests to patch the upstream OAuth flow constructor methods instead of the module-level flow symbol.

### DIFF
--- a/auth/auth_info_middleware.py
+++ b/auth/auth_info_middleware.py
@@ -9,7 +9,6 @@ from fastmcp.server.middleware import Middleware, MiddlewareContext
 from fastmcp.server.dependencies import get_access_token
 from fastmcp.server.dependencies import get_http_headers
 
-from auth.external_oauth_provider import get_session_time
 from auth.oauth21_session_store import ensure_session_from_access_token
 from auth.oauth_types import WorkspaceAccessToken
 
@@ -126,7 +125,17 @@ class AuthInfoMiddleware(Middleware):
                                             access_token = verified_auth
                                         else:
                                             # Standard GoogleProvider returns a base AccessToken;
-                                            # wrap it in WorkspaceAccessToken for identical downstream handling
+                                            # wrap it in WorkspaceAccessToken for identical downstream handling.
+                                            # Imported here (not at module top) so importing this
+                                            # middleware doesn't drag in `auth.external_oauth_provider`
+                                            # (and transitively GoogleProvider's whole FastMCP
+                                            # OAuth-proxy stack) unless a token actually needs
+                                            # wrapping — mirrors the existing lazy
+                                            # `from core.server import get_auth_provider` import above.
+                                            from auth.external_oauth_provider import (
+                                                get_session_time,
+                                            )
+
                                             verified_expires = getattr(
                                                 verified_auth, "expires_at", None
                                             )
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -8,11 +8,10 @@ import logging
 import os
 import webbrowser
 
-from typing import List, Optional, Tuple, Dict, Any
+from typing import List, Optional, Tuple, Dict, Any, TYPE_CHECKING
 from urllib.parse import parse_qs, urlparse
 
 from google.oauth2.credentials import Credentials
-from google_auth_oauthlib.flow import Flow
 from google.auth.transport.requests import Request
 from google.auth.exceptions import RefreshError
 from googleapiclient.discovery import build
@@ -34,6 +33,18 @@ try:
     from fastmcp.server.dependencies import get_context as get_fastmcp_context
 except ImportError:
     get_fastmcp_context = None
+
+if TYPE_CHECKING:
+    from google_auth_oauthlib.flow import Flow
+else:
+    # Deferred at runtime (see the lazy import in create_oauth_flow below): this
+    # single import pulls in the whole google_auth_oauthlib package (including
+    # the `.interactive` submodule), which costs ~700ms of import time and is
+    # only needed when actually *starting* a new OAuth flow — never when
+    # serving tool calls off already-cached, refreshed credentials. Kept as a
+    # real (if `None`) module attribute so it stays a valid
+    # `monkeypatch`/`patch` target for anything that patches the whole name.
+    Flow = None
 
 # Configure logging
 logging.basicConfig(level=logging.INFO)
@@ -326,8 +337,17 @@ def create_oauth_flow(
     state: Optional[str] = None,
     code_verifier: Optional[str] = None,
     autogenerate_code_verifier: bool = True,
-) -> Flow:
+) -> "Flow":
     """Creates an OAuth flow using environment variables or client secrets file."""
+    # Imported lazily (module-global-cached) so importing this module — which
+    # every code path does, including a plain `--help` — doesn't pay the
+    # ~700ms google_auth_oauthlib import cost unless a NEW auth flow is
+    # actually being started. `global` + the NameError guard keep `Flow`
+    # patchable at the module level (existing tests do
+    # `patch("auth.google_auth.Flow.from_client_config", ...)`).
+    global Flow
+    if Flow is None:
+        from google_auth_oauthlib.flow import Flow
     flow_kwargs = {
         "scopes": scopes,
         "redirect_uri": redirect_uri,
--- a/core/server.py
+++ b/core/server.py
@@ -4,7 +4,7 @@ import asyncio
 import hashlib
 import logging
 import os
-from typing import List, Optional
+from typing import List, Optional, TYPE_CHECKING
 from importlib import metadata
 from urllib.parse import urlparse, ParseResult
 
@@ -35,7 +35,6 @@ from core.config import (
 )
 from fastapi.responses import HTMLResponse, JSONResponse, FileResponse
 from fastmcp import FastMCP
-from fastmcp.server.auth.providers.google import GoogleProvider
 from mcp.types import ToolAnnotations, Icon
 from starlette.applications import Starlette
 from starlette.datastructures import MutableHeaders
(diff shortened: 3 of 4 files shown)


szhygulin:
FYI: the failing `pytest` check (`test_forms_tools.py::test_set_publish_settings_builds_publish_state_body`) is pre-existing on unmodified `main` too (verified via `git stash` before/after this patch) — unrelated to this change, not touched by it.

PR #901, opened 2026-07-06
fix: bug sweep, tech debt remediation, test coverage, and type safety

tuongbeo:
## Description
A large multi-phase pass over the `cloudflare-worker` service (and its supporting docs) covering four areas: a bug sweep across the tool handlers, tech debt remediation, new test coverage, and reducing `as any` usage in favor of typed Google API responses. See commit message for the full itemized list; summary below.

### Bug fixes
- `write_google_doc` append mode inserted new content at the start of the doc instead of the end
- `drive.ts` called `driveRequest()` 24 times without importing it — the worker was failing to type-check entirely
- Chat: malformed search filter for `space_name`; `get_chat_messages` double-flipped already-ordered results
- Forms: `progressBar` sent as an object instead of a boolean; `is_quiz` changes dropped from the `updateMask`
- Tasks: `update_task` left a stale `completed` timestamp when moving a task back to `needsAction`
- Gmail: quoted-reply HTML wasn't escaped, so `<`/`&` in the original message could corrupt the draft
- Contacts: batch update used truthy checks instead of `!== undefined`, so clearing a field with `""` silently no-opped
- Apps Script: deployment updates PUT a partial config, wiping unspecified fields on every update
- Slides: `duplicate_slide` accepted `insertion_index` but never applied it
- `search.ts`: `item.link` could be `undefined` and rendered as the literal string `"undefined"` (caught while adding types)
- docs-engine: nested markdown list items were silently dropped by the parser, and the builder's style-cursor drifted out of sync whenever an item had `subItems`

### Tech debt / infra
- Added CI (type-check + test) on every PR; fixed the empty `dependabot.yml` stub; added a `deploy:office` script
- Bumped `agents`, `@modelcontextprotocol/sdk`, and `@cloudflare/workers-oauth-provider`, resolving a duplicate-SDK install that was breaking type-checking on the two `McpAgent` entry points
- Wired token namespace through `env.TOKEN_NAMESPACE` instead of a hardcoded per-agent string; `/health` now reports a `namespace_mismatch` if the wrangler config and the deployed agent ever disagree
- Deleted `tools/composite.ts` (confirmed via git history it was dead code left over from before `write_google_doc` existed) and its two no-op call sites
- Rewrote `README.md`, which had described a different (Python/self-host) architecture than what's actually in this repo; added `cloudflare-worker/README.md`; fixed `SECURITY.md`'s stale contact/versioning info; updated contact email across README/SECURITY/landing pages

### Test coverage
- Added ~85 tests covering `google-tokens.ts` (refresh/retry/backoff, lock contention, the 5-minute stale-token grace window), `auth/google.ts` and `workers/shared.ts` (delegated OAuth, multi-tenant routing, including a cross-tenant replay-attack regression test), `google.ts`, `index.ts`, and the docs-engine nested-list fix
- `npm run test:coverage` now passes; it was failing outright before (0% coverage on 6 files with configured thresholds)

### Type safety
- Added `google-api-types.ts` with minimal per-service response interfaces and replaced ~76 blind `as any` casts across chat/calendar/forms/tasks/search/gmail/contacts/slides tools with them

## Type of Change
- [x] Documentation update
- [ ] New feature
- [ ] Breaking change

## Testing
- [x] New and existing unit tests pass locally with my changes (`npm test`, 180/180 passing)
- [ ] I have tested this change manually against live Google APIs (not possible in this environment — verified via `tsc --noEmit` and unit tests with mocked fetch/KV)

## Checklist
- [x] My changes generate no new warnings (`tsc --noEmit` clean)

## Additional Notes
Two items were investigated but deliberately **not** changed, noted here for visibility:
- A suspected sheets-engine ordering bug (`total_rows` vs `section_headers`) turned out, on careful re-analysis plus a docs search, to likely not be a real bug — `insertDimension` shifts formatting along with the rows it moves. Left as-is rather than guess at unverifiable live Sheets API behavior.
- `mcp-worker.ts`/`office-agent.ts` remain at 0% test coverage — they import `agents/mcp`, which pulls in the `cloudflare:workers` runtime module that Node's ESM loader can't resolve outside `workerd`. Properly testing them needs `@cloudflare/vitest-pool-workers`, a separate infra investment.

The remaining ~134 `as any` casts (concentrated in `docs.ts`, `drive.ts`, `appsscript.ts`, `sheets.ts` — the most deeply-nested Google API shapes) were not converted in this pass; happy to continue in a follow-up.


tuongbeo:
Opened in error — wrong target repo (this was meant for a fork-internal PR, not upstream). Closing.

PR #907, opened 2026-07-07
fix(gmail): RTL-aware base direction for composed message bodies

seidnerj:
## Problem
Outbound Hebrew/Arabic email composed via the Gmail-web faithful path
rendered left-aligned (LTR). `new_message_html` hardcoded `<div dir="ltr">`,
and even plain-text new messages routed through it, so the HTML alternative
Gmail displays was always LTR.

## Fix
- `base_text_direction(text)` — resolves base paragraph direction using the
  Unicode Bidi first-strong-character rule (`unicodedata.bidirectional`).
- `new_message_html(body_html, direction="ltr")` emits `dir="rtl"` when the
  body is RTL. Default stays `"ltr"` → **LTR output byte-identical**.
- New `direction: "auto" | "ltr" | "rtl"` param (default `"auto"`) on
  `send_gmail_message` / `draft_gmail_message`, threaded through the
  new-message, reply, and forward-note paths.

Embedded opposite-direction runs (English words, `$`/`₪` amounts) render
correctly via the browser's bidi algorithm. Gmail's own attribution/forward
headers stay `dir="ltr"` (their markup, not user content).

## Tests
343/343 gmail tests pass; 17 new. Existing web-compose fidelity + golden
tests unchanged. `ruff format` + `ruff check` clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for sending Gmail messages through either API or SMTP, with automatic transport selection.
  * Improved OAuth callback handling for more reliable sign-in flows and port behavior.
  * Enhanced Gmail compose, reply, draft, and forward handling for more faithful message formatting, attachments, and text direction.

* **Bug Fixes**
  * Improved credential directory checks to avoid race conditions and accidental cleanup.
  * Fixed scope handling for broader Gmail and contacts access when needed.
  * Made optional service authentication failures non-blocking in supported flows.

* **Documentation**
  * Expanded README guidance for OAuth ports, send transport settings, and security behavior.

* **Tests**
  * Added extensive coverage for transport, OAuth, MIME formatting, and attachment behavior.


seidnerj:
Folding this into #878. The RTL fix depends on the Gmail-web compose path introduced there (`gmail_web_mime.py` doesn't exist on `main` yet), so it's now included in that PR (commit 02226a0) rather than as a separate stacked PR. Closing.

PR #914, opened 2026-07-10
fix: raise refresh-lock KV TTL to Cloudflare's 60s minimum

tuongbeo:
## Summary
- `getValidAccessToken()` in `cloudflare-worker/src/google-tokens.ts` acquired a refresh lock in Cloudflare KV with `expirationTtl: 30`, but Cloudflare KV rejects any TTL under 60 seconds.
- This caused every token refresh to fail with `KV PUT failed: 400 Invalid expiration_ttl of 30. Expiration TTL must be at least 60.` whenever the stored Google access token was near expiry — which in turn broke every Office MCP write tool (`write_sheet_values`, `create_drive_file`, etc.), not just Sheets.
- Fix: bump the lock TTL to 60s, the minimum Cloudflare allows.

## Test plan
- [x] `npx vitest run src/__tests__/google-tokens.test.ts` — 21/21 passing
- [ ] Deploy the worker and confirm a write tool call succeeds once the cached access token is near expiry


tuongbeo:
Opened by mistake — gh defaulted the base to this upstream repo instead of my own fork. Closing; the intended fix will be a PR against my fork instead.

Issue #912, opened 2026-07-09
Feature request: add Gmail draft lifecycle tools


gwtaylor:
## Summary

I would like to see the Gmail draft support extended beyond `draft_gmail_message` so that an agent can manage draft lifecycles without crossing into send capability. I think the useful set would be:

- `list_gmail_drafts`
- `get_gmail_draft`
- `update_gmail_draft`
- `delete_gmail_draft`

A supported use case is a workflow where an agent prepares draft replies for review, but the human remains responsible for sending. `draft_gmail_message` gets the first part of that right, but once a draft exists there does not seem to be a way to inspect it as a draft resource, replace its contents, or delete it through gws-mcp.

## Why this matters

The Gmail API model is a bit subtle here. Drafts are stable containers, but the message inside the draft is replaced when the draft is updated. Google's docs describe this explicitly: the contained message cannot be edited in-place, but it can be replaced using `users.drafts.update`, with the draft ID remaining stable and the underlying message ID changing.

That behaviour seems like a good match for agent-assisted draft review. For example, an agent might:

1. Create a draft reply in a thread.
2. Later revise it after the user gives feedback.
3. Replace the existing draft rather than creating a second competing draft.
4. Delete the draft if the user decides not to send it.

Right now the safe workflow is awkward because a revised draft means creating another draft and leaving the old one behind for the user to clean up manually.

### `list_gmail_drafts`

List draft resources for the authenticated user. Ideally this would return enough metadata to let an agent identify the correct draft without pulling full message bodies by default:

- draft ID
- contained message ID
- thread ID
- subject
- recipients
- date / internal date
- snippet

If feasible, query support would be helpful, e.g. `in:drafts subject:(...)`, but I think a simple list operation would already be valuable.

### `get_gmail_draft`

Fetch a specific draft by draft ID. I would suggest supporting a `format` option similar to Gmail's API (`metadata`, `full`, `raw`), since `raw` is useful when the goal is to update the draft while preserving headers and MIME details.

### `update_gmail_draft`

Replace the contents of an existing draft by wrapping `users.drafts.update`.

I think this should mirror the ergonomics of `draft_gmail_message` as much as possible: `to`, `cc`, `bcc`, `subject`, body / HTML body, attachments, `thread_id`, and reply headers where relevant. The tool result should probably be explicit that this is replacement rather than in-place editing: same draft resource, new underlying message.

### `delete_gmail_draft`

[...]

PR #812, opened 2026-05-26
Consolidate draft management into manage_gmail_draft

reinlemmens:
## Summary
Reworked per maintainer feedback to avoid growing the OpenAI tool-payload size.

- Merged `draft_gmail_message` and the new delete capability into a single **`manage_gmail_draft(action="create"|"delete", ...)`** tool, matching the existing `manage_gmail_label` / `manage_gmail_filter` / `manage_doc_tab` convention.
- Net tool count change in `gmail.extended`: **-1** (two tools become one).
- Same Gmail Compose scope; delete is permanent (bypasses Trash) per Gmail API `users.drafts.delete`.
- Validation: `subject`/`body` required for `create`, `draft_id` required for `delete`; raises `UserInputError` otherwise.

## Breaking change
`draft_gmail_message` is removed — callers must switch to `manage_gmail_draft(action="create", ...)`. Keyword arguments are unchanged, so it's mostly a rename + one extra kwarg.

## Test plan
- [x] `uv run pytest tests/gmail/ -q` — **146/146 pass** (existing draft coverage migrated + new delete/validation tests)
- [x] Manually verified draft creation and deletion end-to-end against a live Gmail account
- [x] Rebased on current `main` (includes #889's reply-threading fix, merged cleanly into the create branch)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #920, opened 2026-07-12
feat(gmail): add draft management tools — list, update (with thread attach), delete

MoraesGil:
## Summary

Adds three Gmail draft-management tools that round out the existing `draft_gmail_message`:

- **`list_gmail_drafts`** — paginated draft listing (`drafts().list`) with per-draft Subject + snippet (via `messages().get(format="metadata")`), Gmail search-query filtering, and `nextPageToken` surfacing. Read-only (`readOnlyHint=True`, `is_read_only=True`).
- **`update_gmail_draft`** — full replacement update of an existing draft (`drafts().update`), mirroring `draft_gmail_message`'s signature/reply-context/quote/attachment flow (same private helpers). Adds an opt-in **`attach_to_thread`** flag (requires `thread_id`, else `UserInputError`) that sets `message.threadId` on the update so the draft appears **inside the conversation** in the Gmail UI.
- **`delete_gmail_draft`** — permanent draft deletion (`drafts().delete`), `destructiveHint=True`, docstring warns it does not go to Trash.

All three are registered in `core/tool_tiers.yaml` (`list`/`update` → `extended`, `delete` → `complete`).

## Motivation

Agent workflows that prepare reply drafts for a human to finish (attach files from a phone, review, send) currently have no way to: find which draft is which, fix a draft without creating duplicates, remove superseded drafts, or make a reply draft show up inside its Gmail conversation. We hit all four gaps in production use against a Zendesk-backed support thread and had to fall back to raw API scripts.

## Design notes

- **Why `attach_to_thread` is opt-in:** `draft_gmail_message` deliberately keeps `threadId` off the created draft unless full reply headers were derived (avoids UI-hidden drafts). On *update* we found (validated against the live Gmail API, 2026-07-11) that setting `threadId` makes the draft render inside the conversation view as users expect. The flag keeps the safe default (`False` never sets `threadId`) and makes the UI-attach behavior an explicit choice.
- **ID churn documented:** Gmail may assign a new draft ID and always assigns a new message ID after `drafts().update`. The return string surfaces both new IDs and the docstrings steer callers to `list_gmail_drafts` instead of caching IDs.
- `update_gmail_draft` replicates (rather than refactors) `draft_gmail_message`'s flow to leave the existing, tested tool untouched. Happy to extract a shared helper if maintainers prefer.

## Tests

- 16 new tests in `tests/gmail/test_list_gmail_drafts.py` / `test_update_gmail_draft.py` / `test_delete_gmail_draft.py`, following the existing mock patterns (decorator unwrap, raw-MIME parsing, keyed `side_effect` per message ID).
- `uv run pytest tests/gmail/ -q` → **158 passed** (143 pre-existing + 15 collected new cases), no regressions.
- `uv run ruff check gmail/ core/ tests/gmail/` → clean.

https://claude.ai/code/session_01W8f3k8TDS7PBU59mgk794b

## Summary by CodeRabbit

* **New Features**
  * Added Gmail draft management tools: list drafts (with pagination and optional search), update drafts (plain or HTML, signatures, quoted replies, attachments), and permanently delete drafts.
  * Updating can optionally attach drafts to a conversation thread and derive reply headers for proper threading.
* **Enhancements**
  * Expanded tool access tiers to include draft-related actions for the appropriate Gmail tool tiers.
* **Tests**
  * Added contract tests covering list, update, and delete draft tool behaviors and expected Gmail API calls.

### DIFF
--- a/core/tool_tiers.yaml
+++ b/core/tool_tiers.yaml
@@ -12,12 +12,15 @@ gmail:
     - list_gmail_labels
     - manage_gmail_label
     - draft_gmail_message
+    - list_gmail_drafts
+    - update_gmail_draft
     - list_gmail_filters
     - manage_gmail_filter
 
   complete:
     - get_gmail_threads_content_batch
     - batch_modify_gmail_message_labels
+    - delete_gmail_draft
     - start_google_auth
 
 drive:
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -2626,6 +2626,452 @@ async def draft_gmail_message(
     return f"Draft created{attachment_info}! Draft ID: {draft_id}"
 
 
+@server.tool(
+    title="List Gmail Drafts",
+    annotations=ToolAnnotations(
+        readOnlyHint=True,
+        destructiveHint=False,
+        idempotentHint=True,
+        openWorldHint=True,
+    ),
+)
+@handle_http_errors("list_gmail_drafts", is_read_only=True, service_type="gmail")
+@require_google_service("gmail", GMAIL_COMPOSE_SCOPE)
+async def list_gmail_drafts(
+    service,
+    user_google_email: str,
+    page_size: Annotated[
+        int,
+        Field(
+            description="Maximum number of drafts to return per page (capped at 50). Defaults to 20.",
+        ),
+    ] = 20,
+    page_token: Annotated[
+        Optional[str],
+        Field(
+            description="Token for retrieving the next page of results. Use the next page token from a previous response.",
+        ),
+    ] = None,
+    query: Annotated[
+        Optional[str],
+        Field(
+            description="Optional Gmail search query to filter drafts (same syntax as Gmail search, e.g. 'subject:invoice' or 'to:someone@example.com').",
+        ),
+    ] = None,
+) -> str:
+    """
+    Lists draft emails in the user's Gmail account, showing each draft's subject
+    and snippet for quick identification. Supports pagination and Gmail search
+    query filtering.
+
+    Args:
+        user_google_email (str): The user's Google email address. Required.
+        page_size (int): Maximum number of drafts to return per page (capped at 50). Defaults to 20.
+        page_token (Optional[str]): Token for retrieving the next page of results. Use the
+            next page token from a previous response.
+        query (Optional[str]): Optional Gmail search query to filter drafts.
+
+    Returns:
+        str: The total number of drafts found, followed by one line per draft with
+        the Draft ID, Message ID, Thread ID, Subject, and a short snippet. Includes
+        a pagination token if more results are available.
+
+    Examples:
+        # List the 20 most recent drafts
+        list_gmail_drafts()
+
+        # List drafts matching a search query
+        list_gmail_drafts(query="subject:invoice")
+
+        # Get the next page of results
+        list_gmail_drafts(page_token="EAAY...")
+    """
+    logger.info(
+        f"[list_gmail_drafts] Email: '{user_google_email}', Query: '{query}'"
+    )
+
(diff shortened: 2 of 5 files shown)


gwtaylor:
Thanks for this! It implements most of what I asked for in #912 a few days earlier (it looks like we arrived at the same gap independently; CodeRabbit linked the two). I have the same use case: an agent prepares and revises reply drafts, but a human stays responsible for sending. Without list/update/delete, revising a draft means creating a second draft and leaving the stale one behind for manual cleanup.

One thing worth flagging: there are three earlier open PRs in this space (#699, #781, #812), and in #812 the maintainer asked for this functionality to be built into existing tools rather than added as small purpose-built ones, because the server is at the OpenAI payload max limit. So I suspect this PR, as shaped, will stall the same way those did.

I think there is a middle path that respects that constraint. The draft lifecycle can't be folded entirely into existing tools, because update and delete need the draft resource ID, and `search_gmail_messages` only surfaces message IDs (the message resource doesn't carry its draft ID). But the operations could be consolidated into a single tool, e.g. `manage_gmail_draft` with an `action` parameter (`list` / `update` / `delete`), following the precedent already set by `manage_gmail_label`. That would add one tool rather than three, and could supersede #699, #781, and #812 at the same time.

A small scope question: is there a way to read back a draft's current body? Update here is replacement-based (which matches `users.drafts.update`), so a caller that keeps its own state doesn't strictly need it, but a `get` action (or a `format` option on list) would round out the lifecycle for callers that don't.

123andy:
Nice — these round out draft management well. Heads up on a complementary PR I just opened, #943 (`send_gmail_draft` + draft-forward), so we don't step on each other: it deliberately **does not** add a list-drafts tool, deferring listing to your `list_gmail_drafts`. Between the two, #920 covers list/update/delete and #943 covers send + draft-forward — no overlap.

One small suggestion for `list_gmail_drafts` that pairs directly with a send-a-draft flow: **surface each draft's `thread_id`** (it's already in the `drafts.list` response as `draft.message.threadId`, alongside the `id` you use for the metadata `messages.get`). Rationale: a draft's `draft_id` *and* `message_id` both rotate when the draft is edited in the Gmail UI (an edit can discard-and-recreate the draft resource), but its `thread_id` does not. So `thread_id` is the only stable handle across a "create → user edits → send" flow — and #943's `send_gmail_draft(thread_id=…)` uses exactly that to re-resolve the live draft at send time. Exposing it in the listing would let callers hold the edit-proof handle from the start.

Happy to coordinate ordering if it helps.

MoraesGil (commented):
show me the truth

123andy:
**Correction to my earlier comment** — I need to walk back the reasoning I gave for surfacing `thread_id`.

I claimed a draft's `draft_id` and `message_id` both rotate on edit while `thread_id` stays put, making `thread_id` the "edit-proof" handle. I then tested it against the live Gmail API — editing one reply draft repeatedly on Gmail **web and iOS** — and the result is nearly the opposite:

| Handle | Behavior across edits (web + mobile) |
|---|---|
| `message_id` | **rotates on every edit** — never cache it |
| `draft_id` | **most stable** — unchanged across body edits, a subject change, and adding an attachment; couldn't reproduce rotation at all |
| `thread_id` | **not invariant** — editing the **subject** re-threads the draft (Gmail threads by subject) → new `thread_id` |

So the accurate version: prefer **`draft_id`** as the durable handle, treat **`message_id`** as the one to never cache, and know **`thread_id`** is stable only while the subject is untouched. Surfacing `thread_id` in the listing is still worth doing (handy anchor/disambiguator) — just not as "the edit-proof ID."

On my end this changed the send tool in #943: it now resolves by `draft_id` first and only falls back to a `thread_id` scan, with the subject-edit caveat documented. Flagging here so nobody builds on the wrong assumption — apologies for the overclaim.

PR #921, opened 2026-07-12
fix: accept camelCase tool arguments (calendarId, timeMin, ...) via normalization middleware

### LINKED ISSUE #918: get_events rejects standard camelCase args (calendarId, timeMin, timeMax) — schema uses snake_case with additionalProperties:false
## Summary
The `get_events` tool declares its input schema in **snake_case**
(`calendar_id`, `time_min`, `time_max`, `max_results`) with
`additionalProperties: false`. Any caller passing the conventional Google
Calendar API camelCase field names (`calendarId`, `timeMin`, `timeMax`,
`maxResults`, `singleEvents`, `orderBy`) gets a hard validation error
**before** the request reaches Google — even when authenticated.

This breaks every caller that emits camelCase, including the official
Hermes `google-workspace` skill wrapper (`google_api.py`), which sends
camelCase to the `gws`/MCP layer. The tool itself works fine when called
with the correct snake_case args.

## Environment
- workspace-mcp: 1.22.0 (via `uvx workspace-mcp`)
- pydantic: 2.13.4
- launcher: `uvx workspace-mcp --tools gmail calendar --read-only --single-user`
- OS: macOS 26.5.2 (aarch64)

## Reproduction

```python
import asyncio, os, time
from mcp import ClientSession, StdioServerParameters
from mcp.client.stdio import stdio_client

async def main():
    env = dict(os.environ)
    env["GOOGLE_OAUTH_CLIENT_ID"] = os.environ.get("GOOGLE_OAUTH_CLIENT_ID", "")
    env["GOOGLE_OAUTH_CLIENT_SECRET"] = os.environ.get("GOOGLE_OAUTH_CLIENT_SECRET", "")
    env["OAUTHLIB_INSECURE_TRANSPORT"] = "1"
    env["USER_GOOGLE_EMAIL"] = "john@kmiec.us"
    params = StdioServerParameters(
        command="bash",
        args=["-c", "exec uvx workspace-mcp --tools gmail calendar --read-only --single-user"],
        env=env,
    )
    async with stdio_client(params) as (read, write):
        async with ClientSession(read, write) as session:
            await session.initialize()
            now = time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime())
            end = time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime(time.time() + 86400))
            # camelCase — what most callers (and the Hermes skill) send
            await session.call_tool("get_events", {
                "calendarId": "primary",
                "timeMin": now, "timeMax": end,
                "singleEvents": True, "orderBy": "startTime", "maxResults": 10,
            })

asyncio.run(main())
```

## Observed error
```
WARNING  Invalid arguments for tool 'get_events':
  [{'type':'unexpected_keyword_argument','loc':('calendarId',),'msg':'Unexpected keyword argument','input':'primary'},
   {'type':'unexpected_keyword_argument','loc':('timeMin',), ...},
   {'type':'unexpected_keyword_argument','loc':('timeMax',), ...},
   {'type':'unexpected_keyword_argument','loc':('singleEvents',), ...},
   {'type':'unexpected_keyword_argument','loc':('orderBy',), ...},
   {'type':'unexpected_keyword_argument','loc':('maxResults',), ...}]
ERROR   Error in on_call_tool middleware: 6 validation errors for call[get_events]
```

The validation fires in the server's own pydantic model — it happens
**before** auth is even checked (`token=none` in the same log line), so it
is purely a schema/naming contract problem.

## Works with snake_case
Same call, snake_case, validates and proceeds to the Google API
(only fails here because this throwaway server had no token):

```python
await session.call_tool("get_events", {
    "calendar_id": "primary",
    "time_min": now, "time_max": end,
    "max_results": 10, "user_google_email": "john@kmiec.us",
})
# -> GoogleAuthenticationError: ACTION REQUIRED: Google Authentication Needed
#    (i.e. schema passed; reaches the API)
```

## Root cause
`additionalProperties: false` on the `get_events` input model rejects any
field not in its snake_case `properties`. Callers that use the canonical
Google Calendar camelCase names are rejected outright.

## Suggested fix (either)
1. Accept both spellings — add camelCase aliases (`calendarId`, `timeMin`,
   `timeMax`, `maxResults`, `singleEvents`, `orderBy`) to the input model, or
2. Relax `additionalProperties` / document the snake_case contract prominently
   and have the Hermes `google-workspace` skill emit snake_case.

Most users (and LLM callers) expect the Google Calendar camelCase field
names, so option 1 is the least-surprising.

## Evidence
- Declared input schema (snake_case, `additionalProperties:false`):
  `calendar_id`, `time_min`, `time_max`, `max_results`, `event_id`,
  `query`, `detailed`, `include_attachments`, `user_google_email`.
- Confirmed: camelCase call -> 6 pydantic `unexpected_keyword_argument`
  errors; snake_case call -> passes schema validation, reaches Google API.

Bortlesboat:
Fixes #918

Every tool declares snake_case parameters with `additionalProperties: false`, so callers that mirror the Google API field names (`calendarId`, `timeMin`, `maxResults`, ...) fail schema validation before the request ever reaches Google — including the Hermes google-workspace skill wrapper mentioned in the issue.

This adds a small FastMCP middleware on the `call_tool` path that renames an argument key only when all of these hold:

- the key isn't itself a declared parameter of the tool,
- its snake_case conversion **is** a declared parameter, and
- the caller didn't also pass the snake_case key explicitly.

Ambiguous input (two spellings normalizing to the same parameter, e.g. `iCalUid` + `iCalUID`) is left untouched so validation still rejects it instead of silently picking a winner. Tools that genuinely declare camelCase params are unaffected, and unknown arguments keep failing exactly as before.

Went with the middleware so it covers all tools at once rather than aliasing the calendar ones — happy to scope it down to explicit calendar aliases instead if you'd rather keep argument handling strict.

Tested with the reproduction from #918 through an in-memory client — 8 tests in `tests/core/test_camel_case_middleware.py` covering the camelCase path, snake_case passthrough, explicit-key precedence, collisions, declared-camelCase params, and unknown tools/args. `ruff check` and `ruff format` clean; `tests/core` + `tests/gcalendar` pass locally.

## Summary by CodeRabbit

- **New Features**
  - Tool calls now accept camelCase argument names when corresponding parameters use snake_case.
  - Existing snake_case arguments remain unchanged.
  - Declared camelCase parameters are preserved without modification.
  - Ambiguous or conflicting argument names continue to be rejected safely.
  - Unknown arguments and tools continue to be handled with existing validation behavior.

- **Tests**
  - Added coverage for conversions, validation, conflicts, unknown arguments, and unknown tools.


taylorwilsdon:
Thanks!

Issue #716, opened 2026-04-22
bug: --single-user mode breaks every hour for non-first accounts — refresh-path rebind check


spencer-c-reed:
## Bug Description

In `--single-user` mode with multiple credential files, any account other than the one whose session got bound first silently breaks with "ACTION REQUIRED: Authentication Needed" **exactly one hour after its token was last refreshed**. The refresh itself works — the new access token is successfully fetched from Google and written to the credential file on disk — but the subsequent `OAuth21SessionStore.store_session()` call fails the immutable session-binding check, the caller treats that as "no valid credentials," and the user sees a re-auth prompt they don't actually need.

This is related to but distinct from #433, #506, and #642, which all describe initial call-time mismatches (first-file-alphabetical, case sensitivity, wrong email). This is the **refresh path**: everything is correct at call time, but the hourly background refresh triggers the rebind check on accounts beyond the first.

## Symptoms

In a long-lived MCP session that uses multiple `user_google_email` values in `--single-user` mode:

- **First account called** in the session: works indefinitely (refresh stays within the session binding).
- **Every other account**: works on the first call (initial access token is still valid). ~1 hour later, the next call fails with:
  > ACTION REQUIRED: Google Authentication Needed for Google Gmail for 'accountB@example.com'

The user is forced to re-approve OAuth for accountB. After re-approval, the clock resets and the same failure recurs an hour later. Filing this as Spencer Reed @spencer-c-reed — my agent traced the bug over about 30 minutes of log analysis after repeated re-auth prompts that "kept happening" over multiple days.

## Steps to Reproduce

1. Configure `--single-user` mode with credentials for two Google accounts, both with valid refresh tokens in `~/.workspace-mcp/credentials/` (default `WORKSPACE_MCP_CREDENTIALS_DIR`).
2. Start the workspace-mcp server (e.g., `workspace-mcp --tools gmail --transport streamable-http --single-user`).
3. In a single MCP client session, call `search_gmail_messages(user_google_email="accountA@example.com", ...)`. Works.
4. Call `search_gmail_messages(user_google_email="accountB@example.com", ...)`. Works (access token still valid).
5. Wait until accountB's access token expires (1 hour after it was last refreshed).
6. Call `search_gmail_messages(user_google_email="accountB@example.com", ...)` again. **Fails with "ACTION REQUIRED: Authentication Needed"**, despite accountB's refresh_token being valid and the refresh itself having succeeded.

## Log Evidence

[...]

PR #928, opened 2026-07-15
fix: persist refreshed secondary-account credentials on rebind

### LINKED ISSUE #926: Multi-account stdio (non --single-user): successful token refresh discarded by session rebind guard → spurious re-auth (v1.22.0). #716's fix is single-user-only
## Summary

This is the **non-`--single-user`** variant of #716 (closed/completed). The mechanism is identical — a successful token refresh for a secondary account is thrown away because the post-refresh `store_session()` call trips the immutable session→user rebind guard — but the fix merged for #716 only guards the single-user path (`MCP_SINGLE_USER_MODE == "1"`), so the default multi-user/multi-account stdio setup still reproduces on the current release.

When several Google accounts are driven through one stdio MCP session, any tool call for a *secondary* account whose access token has expired triggers a full re-auth prompt — even though the refresh token is valid, the refresh succeeds, and the rotated token is persisted to the credential store.

## Affected version

**workspace-mcp 1.22.0** (latest; installed as the Claude Desktop `.mcpb` extension).
Transport: **stdio**. Auth: default (`MCP_ENABLE_OAUTH21` unset). **Not** `--single-user`.

## Why #716's fix doesn't cover this

#716's fix lives in `auth/oauth21_session_store.py` `store_session()` (v1.22.0):

```python
# Single-user mode reads credentials directly by email and bypasses the
# session mapping. Keeping the immutable MCP session binding enabled here
# turns a successful token refresh for a second account into a false "rebind" error.
if os.getenv("MCP_SINGLE_USER_MODE") == "1":
    mcp_session_id = None
```

That drop-the-binding guard is gated on `MCP_SINGLE_USER_MODE`. In a **default multi-user** run (tools pass `user_google_email` per call, several accounts share one stdio session) `MCP_SINGLE_USER_MODE` is unset, so `mcp_session_id` is kept and the rebind check a few lines down still raises:

```python
elif self._session_auth_binding[mcp_session_id] != user_email:
    logger.error(f"SECURITY: Attempt to rebind session {mcp_session_id} from {self._session_auth_binding[mcp_session_id]} to {user_email}")
    raise ValueError(f"Session {mcp_session_id} is already bound to a different user")
```

## Observed log sequence (secondary account, expired access token)

```
[get_credentials] Session user primary@gmail.com doesn't match requested secondary@gmail.com; skipping session store
[get_credentials] Credentials not valid. Attempting refresh. User: 'secondary@gmail.com', Session: '<sid>'
[get_credentials] Credentials refreshed successfully. User: 'secondary@gmail.com', Session: '<sid>'
[CREDS] Stored credentials for secondary@gmail.com ...            # refresh persisted to disk
oauth21_session_store.store_session:660 - SECURITY: Attempt to rebind session <sid> from primary@gmail.com to secondary@gmail.com
[get_credentials] Error refreshing credentials: Session <sid> is already bound to a different user. ...
```
→ `get_credentials()` returns `None` → re-auth flow initiated for the secondary account.

Over one real session, a single primary-bound stdio session produced **169** such `session user primary doesn't match requested <secondary>` events against one secondary account.

## Root cause (v1.22.0)

In `auth/google_auth.py` `get_credentials()`, the file-credential refresh branch (~lines 1095–1172):

1. `credentials.refresh(Request())` succeeds.
2. `credential_store.store_credential(user_google_email, credentials)` succeeds — rotated token saved to disk.
3. The code then **unconditionally** calls `store.store_session(..., mcp_session_id=session_id)` (~line 1145).
4. That session is already bound to the primary account, so `store_session` raises `ValueError` (rebind guard, `oauth21_session_store.py:658–665`).
5. The generic `except Exception` around the refresh block turns it into `return None` (~line 1172), which the caller surfaces as "authentication required".

The `store_session()` call here is a **session-cache update**, but its failure discards a refresh that already succeeded and already persisted. This path is reached whenever FastMCP supplies a `session_id`, which it does in stdio.

## Impact

Multi-account stdio users get a spurious re-auth prompt roughly every time a secondary account's access token expires (~hourly), despite valid refresh tokens. Because the refresh *does* persist to disk first, an immediate retry of the same call usually succeeds (it then reads the fresh token and skips refresh) — but the first attempt still prompts.

## Suggested fix

Either broaden #716's guard, or make the post-refresh session-cache update non-fatal so it can never discard a completed refresh. The latter, in `get_credentials()`:

```python
if persist_succeeded or is_stateless_mode():
    try:
        store = get_oauth21_session_store()
        store.store_session(
            user_email=user_google_email,
            ...
            mcp_session_id=session_id,
            issuer="https://accounts.google.com",
        )
    except Exception as session_store_error:
        logger.warning(
            "[get_credentials] Session store update skipped for user %s "
            "(session %s already bound to a different account); proceeding "
            "with refreshed credentials from the credential store. %s",
            user_google_email, session_id, session_store_error,
        )
```

Alternatively, at the top of `store_session()`, drop `mcp_session_id` when the requested `user_email` differs from the session's existing binding (the read path already handles this correctly via the "skipping session store" branch — only the write-back re-introduces the conflict).

## Related

- #716 (closed/completed) — same mechanism, `--single-user` only; its `MCP_SINGLE_USER_MODE` guard doesn't cover the default multi-user path reported here.
- #642 (open) — single-user caller-email override; different condition.
- #449 (closed) — the *other* refresh path (OAuth 2.1 session store) not persisting to file; different branch.

Happy to open a PR with the best-effort wrapper (verified locally: refresh + persist proceed, no spurious re-auth, multi-user rebind protection for genuinely different sessions is unaffected).

mvanhorn:
## Description
In the default multi-user / multi-account stdio setup, a tool call for a secondary Google account whose access token has expired triggers a full re-auth even though the refresh token is valid and the rotated token is already persisted to disk.

The cause is in the post-refresh write path. After a successful `credentials.refresh()` and `store_credential()`, `get_credentials()` calls `store.store_session(..., mcp_session_id=session_id)`. That session is already bound to the primary account, so `store_session()` raised the immutable-rebind `ValueError`, and the surrounding generic `except Exception` turned that into `return None`, discarding the already-successful refresh. The #716 fix only dropped `mcp_session_id` under `MCP_SINGLE_USER_MODE`, so the default multi-user path still reproduced.

This changes `store_session()` so that when a non-null `mcp_session_id` is already bound to a different user, it drops the session binding (sets it to `None`) instead of raising, mirroring what the read path and single-user mode already do. The original session stays bound to its first user (no hijack), and the second account's refreshed credentials are persisted by email. As defense-in-depth, the post-refresh `store_session(...)` call in the file-credential branch of `get_credentials()` is wrapped so any session-cache write failure is logged and the already-persisted credentials are still returned.

Note on the security contract: the previous behavior rejected a cross-user rebind by raising. That still protected the binding, but at the cost of discarding a valid refresh for the second account. The new behavior keeps the same protection (the session is never rebound to a different user) without the data loss. I updated the existing `test_store_session_rejects_mcp_session_rebind_by_default` accordingly (now `test_store_session_keeps_original_binding_on_cross_user_rebind`) to assert the binding is preserved and the second account's credentials are kept. Happy to revisit if you would rather keep the hard raise and address the discard another way.

## Type of Change
- [ ] Documentation update

## Testing

`uv run pytest tests/auth/test_google_auth_refresh_persistence.py tests/auth/test_oauth21_session_store.py` — 15 passed. `ruff check` is clean on the changed files.

## Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas

## Additional Notes
The new regression coverage exercises the happy path (a secondary account with an expired token over a session bound to the primary returns refreshed credentials), the cross-account rebind (binding preserved, second account persisted), and the store-failure path (a non-rebind `store_session` failure no longer discards a successful refresh).

Closes #926

## Summary by CodeRabbit

* **Bug Fixes**
  * Credential refreshes now complete successfully even if session-store persistence encounters an error.
  * Prevented cross-account session binding conflicts from blocking credential storage.
  * Preserved existing session ownership while saving credentials for another account.

* **Tests**
  * Added coverage for credential refresh failures, cross-account sessions, and binding preservation.

### DIFF
--- a/auth/google_auth.py
+++ b/auth/google_auth.py
@@ -1141,19 +1141,27 @@ def get_credentials(
 
                 if persist_succeeded or is_stateless_mode():
                     # Also update OAuth21SessionStore
-                    store = get_oauth21_session_store()
-                    store.store_session(
-                        user_email=user_google_email,
-                        access_token=credentials.token,
-                        refresh_token=credentials.refresh_token,
-                        token_uri=credentials.token_uri,
-                        client_id=credentials.client_id,
-                        client_secret=credentials.client_secret,
-                        scopes=credentials.scopes,
-                        expiry=credentials.expiry,
-                        mcp_session_id=session_id,
-                        issuer="https://accounts.google.com",  # Add issuer for Google tokens
-                    )
+                    try:
+                        store = get_oauth21_session_store()
+                        store.store_session(
+                            user_email=user_google_email,
+                            access_token=credentials.token,
+                            refresh_token=credentials.refresh_token,
+                            token_uri=credentials.token_uri,
+                            client_id=credentials.client_id,
+                            client_secret=credentials.client_secret,
+                            scopes=credentials.scopes,
+                            expiry=credentials.expiry,
+                            mcp_session_id=session_id,
+                            issuer="https://accounts.google.com",  # Add issuer for Google tokens
+                        )
+                    except Exception as session_store_error:
+                        logger.warning(
+                            "[get_credentials] Failed to update OAuth 2.1 session store "
+                            "for refreshed credentials for user %s: %s",
+                            user_google_email,
+                            session_store_error,
+                        )
 
             if session_id and (persist_succeeded or is_stateless_mode()):
                 # Update session cache if it was the source or is active
--- a/auth/oauth21_session_store.py
+++ b/auth/oauth21_session_store.py
@@ -607,6 +607,15 @@ class OAuth21SessionStore:
         with self._lock:
             normalized_expiry = _normalize_expiry_to_naive_utc(expiry)
 
+            if mcp_session_id:
+                bound_user = self._session_auth_binding.get(mcp_session_id)
+                if bound_user and bound_user != user_email:
+                    logger.info(
+                        f"MCP session {mcp_session_id} is already bound to {bound_user}; "
+                        f"storing credentials for {user_email} without changing the session binding"
+                    )
+                    mcp_session_id = None
+
             # Clean up previous session mappings for this user before storing new one
             old_session = self._sessions.get(user_email)
             if old_session:
@@ -654,14 +663,6 @@ class OAuth21SessionStore:
                     self._session_auth_binding[mcp_session_id] = user_email
                     logger.info(
                         f"Created immutable session binding: {mcp_session_id} -> {user_email}"
-                    )
-                elif self._session_auth_binding[mcp_session_id] != user_email:
-                    # Security: Attempt to bind session to different user
-                    logger.error(
-                        f"SECURITY: Attempt to rebind session {mcp_session_id} from {self._session_auth_binding[mcp_session_id]} to {user_email}"
-                    )
-                    raise ValueError(
-                        f"Session {mcp_session_id} is already bound to a different user"
                     )
 
                 self._mcp_session_mapping[mcp_session_id] = user_email
--- a/tests/auth/test_google_auth_refresh_persistence.py
+++ b/tests/auth/test_google_auth_refresh_persistence.py
@@ -1,4 +1,5 @@
 from auth.google_auth import get_credentials
+from auth.oauth21_session_store import OAuth21SessionStore
 
 
 class _RefreshableCredentials:
@@ -20,9 +21,15 @@ class _RefreshableCredentials:
 
 
 class _OAuthSessionStore:
-    def __init__(self, session_credentials=None, session_user="user@example.com"):
+    def __init__(
+        self,
+        session_credentials=None,
+        session_user="user@example.com",
+        store_error=None,
+    ):
         self._session_credentials = session_credentials
         self._session_user = session_user
+        self._store_error = store_error
         self.store_calls = []
 
     def get_user_by_mcp_session(self, session_id):  # noqa: ARG002
@@ -33,6 +40,8 @@ class _OAuthSessionStore:
 
     def store_session(self, **kwargs):
         self.store_calls.append(kwargs)
+        if self._store_error:
+            raise self._store_error
 
 
 class _CredentialStore:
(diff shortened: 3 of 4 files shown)


arthur-bamboo:
Independent reproduction on **v1.24.0**, default multi-account stdio (not `--single-user`), a month after this was opened — the diagnosis above matches ours exactly.

Sequence from the server's own debug log, two Google accounts in one process:

```
Created immutable session binding: <session> -> account-A
[get_credentials] Credentials refreshed successfully. User: 'account-B'
Stored credentials for account-B to .../account-B.json
SECURITY: Attempt to rebind session <session> from account-A to account-B
[get_credentials] Error refreshing credentials: Session ... is already bound to a different user
[search_gmail_messages] ... initiating auth flow
```

Refresh succeeds, the rotated token is persisted, then the rebind guard discards it. The trigger is call **order**, not a clock: whichever account is touched second walls, however healthy its refresh token.

Two notes that may be useful:

1. **A cheap field tell for triage.** The consent URL emitted by this path carries `prompt=select_account`; a genuinely expired or revoked refresh token yields `prompt=consent`. That one parameter separates this bug from real token expiry, from the browser URL alone. We lost time on a wrong "7-day Testing-publishing-status token expiry" theory before that discriminator settled it.

2. **In support of fixing this at `store_session()` rather than at the call site** — your Additional Notes invite that question. A call-site-only guard is insufficient: `save_credentials_to_session()` calls `store_session()` internally, so the same `ValueError` raises from two paths inside `get_credentials()`, and production logs only ever show the first, because it short-circuits to the enclosing `except` before the second is reached. Patching from the log alone fixes one of two. Removing the raise at the root, as you have, covers both.

Happy to test a build if that helps.

mvanhorn:
@arthur-bamboo -- thank you, that log excerpt is the clearest statement of the bug I've seen. `Credentials refreshed successfully. User: 'account-B'` immediately followed by `SECURITY: Attempt to rebind ...` then a fresh auth flow is exactly the failure: the refresh succeeds, the rotated token is written, and then the store throws it away.

Worth noting what's changed upstream since this PR opened. `main` now short-circuits the binding when `MCP_SINGLE_USER_MODE=1`, which fixes one path -- but your repro is default multi-account stdio, *not* `--single-user`, and that path still raises at `oauth21_session_store.py:688`. So the reported bug is still live on current `main`.

Before I rebase, there's a decision for @taylorwilsdon. `main` now has `test_store_session_rejects_mcp_session_rebind_by_default`, which asserts the `ValueError` this PR removes. This PR does not weaken the binding: the session stays bound to account A, and account B's credentials are persisted *without* an `mcp_session_id`, mirroring the read path. What changes is that a valid refresh is no longer discarded to enforce it.

So: is rejecting a cross-user `store_session` the intended contract, with the real fix belonging further up the call chain? Or is that test encoding the bug, in which case I'll rebase and replace it with the binding-preserved version. Happy either way -- I just don't want to overwrite a deliberate test without asking.

PR #935, opened 2026-07-17
docs(create_doc): note content is plain text; point to import_to_google_doc for Markdown

imparpaulo01:
## What
`create_doc` inserts its `content` argument **verbatim as plain text** (a single `insertText` request), so Markdown passed to it lands as literal characters (`#`, `**`, `-`) instead of rendered formatting. The current docstring only describes `content` as *"plain text content to insert"* and doesn't mention that Markdown isn't rendered, nor point to the alternative.

Because a tool's docstring is the description an LLM reads when choosing tools, this omission leads models to pick `create_doc` for Markdown content and produce documents full of literal `#`/`**` characters.

## Change
Docstring-only, **no behavior change**: adds a short note that `content` is inserted verbatim and Markdown is not rendered, and points to `import_to_google_doc` with `source_format="markdown"` for automatic Markdown/HTML/DOCX conversion (that tool already documents importing Markdown `content` directly).

This is a small nudge that should improve tool selection for anyone building on the server.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that document content is inserted as plain text, without Markdown rendering.
  * Added guidance to use the import tool for converting Markdown, HTML, or DOCX into Google Docs format.

### DIFF
--- a/gdocs/docs_tools.py
+++ b/gdocs/docs_tools.py
@@ -395,6 +395,11 @@ async def create_doc(
     """
     Creates a new Google Doc and optionally inserts initial content.
 
+    Note: `content` is inserted verbatim as plain text. Markdown is NOT rendered
+    (characters such as `#`, `**` and `-` appear literally). To create a document
+    from Markdown (or HTML/DOCX/etc.) with automatic conversion, use
+    `import_to_google_doc` with `source_format="markdown"` instead.
+
     After creation, the document body starts at index 1. A new empty doc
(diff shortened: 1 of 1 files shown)


imparpaulo01:
Thanks for the catch — you're right that `source_format` should match the input type. Fixed in 05a61fe: kept `markdown` for Markdown, and clarified `html` for HTML and `docx` (supplied via `file_path`/`file_url`, since it's binary) for DOCX.

PR #938, opened 2026-07-17
feat(telemetry): export FastMCP spans via OTLP when configured

DrFaust92:
## Description

FastMCP is already instrumented with OpenTelemetry (`fastmcp.telemetry`), but it links only `opentelemetry-api`, so every span is a **no-op** until a `TracerProvider` with an exporter is configured. This wires that provider from the standard `OTEL_*` environment variables, so FastMCP's tool-call spans are actually exported to an OTLP collector.

Fully **opt-in** and safe by default:
- No-op unless `OTEL_EXPORTER_OTLP_ENDPOINT` (or `..._TRACES_ENDPOINT`) is set.
- If the `otel` extra isn't installed, it logs a one-line hint and returns — never raises.

### Changes
- **`core/telemetry.py`** — `configure_telemetry()`: gated on the OTLP endpoint; honors `OTEL_EXPORTER_OTLP_PROTOCOL` (gRPC default per spec, or `http/protobuf`), `OTEL_SERVICE_NAME` (defaults to `google-workspace-mcp`), and the rest of the standard `OTEL_*` vars via the SDK. Registers an `atexit` flush.
- **`main.py`** — calls it once at startup (after logging is configured).
- **`pyproject.toml`** — new opt-in `otel` extra (`opentelemetry-sdk`, `opentelemetry-exporter-otlp`); `uv.lock` regenerated.
- **`Dockerfile`** — `uv sync … --extra otel` so the published image ships the SDK (still a no-op unless an endpoint is set).
- **`README.md`** — short "OpenTelemetry Tracing" section.

## Why

Running this server in production (Kubernetes, streamable-http), there's currently no way to see tool-call latency/errors or to correlate a failing Google API call — the app has no metrics/traces of its own, and FastMCP's built-in spans are dormant. This lights them up with zero code changes for users who don't set an endpoint.

## Type of Change

## Testing
Validated locally against the current image's FastMCP (3.4.2):
- `uv sync --frozen --no-dev --extra disk --extra otel` (the exact Docker build command) resolves against the regenerated lock.
- With `OTEL_EXPORTER_OTLP_ENDPOINT` set: `configure_telemetry()` installs a real SDK `TracerProvider`, and a span from `fastmcp.telemetry.get_tracer()` reports `is_recording() == True` (spans export over OTLP).
- With no endpoint: returns `False`, global provider stays the no-op `ProxyTracerProvider`.
- With the endpoint set but SDK absent: logs the hint and returns `False` (no crash).
- `ruff check` / `ruff format --check` clean.

## Notes
- Traces only — no metrics (FastMCP metrics are still upstream WIP).
- Default protocol follows the OTel spec (gRPC); set `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` for the HTTP exporter.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in OpenTelemetry tracing for monitoring server activity.
  * Tracing activates automatically when an OTLP endpoint is configured; remains disabled otherwise.
  * Supports OTLP HTTP and gRPC exporters with standard `OTEL_*` environment variables.
* **Documentation**
  * Updated setup and security/compliance guidance, including OTLP endpoint examples and tracing enable/disable behavior.
* **Tests**
  * Added coverage for tracing enablement, protocol selection/overrides, resource attributes, and graceful failure behavior.
* **Chores**
  * Updated build/runtime dependency installation to include the OpenTelemetry tracing extras.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,9 @@ RUN pip install --no-cache-dir uv
 COPY . .
 
 # Install Python dependencies using uv sync
-RUN uv sync --frozen --no-dev --extra disk
+# --extra otel ships the OpenTelemetry SDK/exporter so tracing can be enabled at
+# runtime via OTEL_* env vars; it stays a no-op unless an OTLP endpoint is set.
+RUN uv sync --frozen --no-dev --extra disk --extra otel
 
 # Create non-root user for security
 RUN useradd --create-home --shell /bin/bash app \
--- a/README.md
+++ b/README.md
@@ -346,6 +346,27 @@ pip install "workspace-mcp[gcs]"
 
 </details>
 
+<details>
+<summary><b>OpenTelemetry Tracing</b> <sub><sup>← optional distributed tracing</sup></sub></summary>
+
+FastMCP is already instrumented with OpenTelemetry, but spans are a no-op until an
+SDK and exporter are configured. Install the `otel` extra (already included in the
+Docker image) and point the server at an OTLP collector — tracing stays off unless
+an endpoint is set.
+
+```bash
+uv sync --extra otel   # if installing from source
+
+export OTEL_EXPORTER_OTLP_ENDPOINT="http://otel-collector:4317"
+export OTEL_EXPORTER_OTLP_PROTOCOL="grpc"   # or "http/protobuf" (default: grpc)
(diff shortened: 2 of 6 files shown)


taylorwilsdon:
Great addition, thank you! I will get this merged today.

PR #949, opened 2026-07-21
feat(calendar): opt-in reminders in get_events + defaultReminders in list_calendars

123andy:
## Summary

The Google Calendar API already returns each event's `reminders` (`useDefault` + `overrides[]`) and each calendar's `defaultReminders`, but the tool output dropped both — even `get_events(detailed=true)` omitted an event's reminders, and `list_calendars` never surfaced a calendar's default notification settings.

This adds an **opt-in** `include_reminders` flag (default `False`) to both tools, so the context-trimming default is preserved while full notification detail is available on request. This mirrors the existing `include_attachments` pattern on `get_events`.

## Changes

- **`get_events(include_reminders=True)`** — surfaces each event's reminder settings in detailed output, on both the single-event and multi-event paths. Only applies when `detailed=True` (same gating as `include_attachments`).
- **`list_calendars(include_reminders=True)`** — surfaces each calendar's default reminders.
- New formatting helpers `_format_reminders` / `_format_reminder_overrides` in `calendar_helpers.py`.

## Behavior

Defaults are unchanged — no output differs unless a caller passes `include_reminders=True`.

Example detailed output with the flag on:
```
- Reminders: popup 10 min before, email 1440 min before
```
```
- Reminders: Using calendar default reminders
```

## Testing

- Unit tests for the `_format_reminders` / `_format_reminder_overrides` helpers (custom overrides, `useDefault`, no-reminders, and malformed-input cases).
- Integration-level tests that drive the real `get_events` / `list_calendars` functions (unwrapped past their auth/error decorators) against a mock service, confirming reminders are surfaced only when `include_reminders=True` and omitted by default.
- Verified end-to-end against the live Google Calendar API: confirmed `events.list` returns reminder `overrides` (not just `events.get`), and that `events.list`, `events.get`, and `calendarList.list` all render correctly for custom-override, `useDefault`, and no-reminder events.

## Notes

This is read-side only and complements the write-side reminder support from #149 (create/modify event reminders); there is no overlap.

### DIFF
--- a/gcalendar/calendar_helpers.py
+++ b/gcalendar/calendar_helpers.py
@@ -100,6 +100,51 @@ def _format_attachment_details(
     return f"\n{indent}".join(attachment_details_list)
 
 
+def _format_reminder_overrides(overrides: List[Dict[str, Any]]) -> str:
+    """Format a list of reminder override objects (method + minutes) for display.
+
+    Each override looks like ``{"method": "popup", "minutes": 10}``. Returns a
+    comma-separated summary such as ``"popup 10 min before, email 1440 min before"``,
+    or ``"None"`` if there are no overrides.
+    """
+    if not overrides:
+        return "None"
+
+    parts = []
+    for override in overrides:
+        method = override.get("method", "unknown")
+        minutes = override.get("minutes")
+        if minutes is None:
+            parts.append(method)
+        else:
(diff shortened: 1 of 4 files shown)


123andy:
Closing — not ready to propose yet, will reopen when ready.

taylorwilsdon:
Closing at your request

PR #952, opened 2026-07-22
feat(gmail): update an existing draft in place via draft_id on draft_gmail_message

Rlsnk:
## Summary

Adds an optional `draft_id` parameter to the existing `draft_gmail_message` tool. When provided, the tool calls `drafts().update` instead of `drafts().create`: the draft's message is replaced wholesale while the draft ID stays stable, so revising a draft no longer leaves a stale duplicate behind for manual cleanup.

**No new tool is added** — this follows the maintainer's request in #812 to build draft management into existing tools rather than adding purpose-built ones (OpenAI payload max limit). It covers the "update" piece of #912; listing/deletion (#699, #781, #812) and send/forward (#943) are intentionally left to their own PRs.

## Behavior

- `draft_id` omitted → exact current behavior (create). Zero change for existing callers.
- `draft_id` provided → full-replacement update via `drafts().update`. All existing machinery flows through the same code path unchanged: reply-context derivation, `In-Reply-To`/`References` headers and `threadId` (a threaded reply draft stays in its conversation after the update), signatures, `quote_original`, attachments (path/base64/URL), and Send-As.
- The return string distinguishes `Draft updated` from `Draft created`.

## Why full replacement

The Gmail API cannot edit a draft's message field-by-field: `users.drafts.update` replaces the contained message, with the draft ID remaining stable while the underlying message ID rotates (this matches the behavior measured in the #920 discussion — `draft_id` is the stable handle). The parameter description and docstring spell this out so agents know to re-supply every field.

## Motivation

Real-world use case (same as #912): an agent prepares and revises reply drafts, a human stays responsible for sending. With create-only, every revision produces a second draft and the stale one has to be deleted by hand in the Gmail UI.

## Tests

- 2 new tests in `tests/gmail/test_draft_gmail_message.py`: update path (asserts `update(userId, id, body)` kwargs, raw message content, and that `create` is never called) and threaded update (asserts `threadId` + derived reply headers survive the update path).
- Full gmail suite: 145 passed. `ruff check` and `ruff format` clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for updating an existing Gmail draft in place using its draft ID.
  * Existing drafts can retain thread context, including reply headers and conversation association.
  * Updated status messages and usage documentation to distinguish between created and updated drafts.

* **Tests**
  * Added coverage for draft updates and preservation of thread context.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -2433,10 +2433,18 @@ async def draft_gmail_message(
             description="Whether to include the original message as a quoted reply. Requires thread_id. Defaults to false.",
         ),
     ] = False,
+    draft_id: Annotated[
+        Optional[str],
+        Field(
+            description="Optional ID of an existing draft to update in place (drafts().update) instead of creating a new draft. The draft's message is fully replaced, so re-supply every field (subject, body, recipients, threading) as it should appear after the update. The draft ID remains stable across updates.",
+        ),
+    ] = None,
 ) -> str:
     """
     Creates a draft email in the user's Gmail account. Supports both new drafts and reply drafts with optional attachments.
     Supports Gmail's "Send As" feature to draft from configured alias addresses.
+    When draft_id is provided, updates that existing draft in place (full message replacement)
+    instead of creating a new one, so revising a draft does not leave stale duplicates behind.
 
     Args:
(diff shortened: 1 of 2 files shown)


Rlsnk:
Apologies for the noise — closing: this was opened by mistake on my side (wrong assumption about the target repo).

PR #957, opened 2026-07-23
fix: streamable-http port probe SO_REUSEADDR + ERROR bind logging

### LINKED ISSUE #946: streamable-http pre-flight port-bind check causes a silent crash-loop on fast restart (SystemExit swallowed, error only at DEBUG)
## Summary

When the server restarts quickly in an environment where the previous process's listener
socket has not yet been released (e.g. a Kubernetes pod whose container restarts in-place,
sharing one network namespace), the streamable-HTTP startup path exits with code 1 and **no
visible log output**, then crash-loops until the port frees. Two independent issues combine to
make this both happen and be undiagnosable.

## Environment

- Version: 1.22.0 (code is unchanged on `main` at time of writing).
- Transport: `--transport streamable-http`, OAuth 2.1 enabled.
- Deployment: container orchestration (Kubernetes) where a restarting process re-enters a
  network namespace that may still briefly hold the old listener on the configured port.
- `sys.stderr` is not a TTY (standard for a containerized service).

## Root cause

In `main.py`, the streamable-HTTP branch performs a pre-flight bind before starting the server
(around lines 851–860):

```python
if args.transport == "streamable-http":
    try:
        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
            s.bind((host, port))
    except OSError as e:
        safe_print(f"Socket error: {e}")
        safe_print(f"❌ Port {port} is already in use. Cannot start HTTP server.")
        sys.exit(1)
    server.run(...)
```

This is fragile in two ways:

1. **The pre-flight socket does not set `SO_REUSEADDR`**, and it duplicates the bind that
   uvicorn performs inside `server.run()` (uvicorn sets `SO_REUSEADDR` and reports bind errors
   clearly). On a fast restart a transient `EADDRINUSE` fails this check even though the real
   server bind moments later would often succeed.

2. **The failure is effectively invisible**, so an operator sees only an exit-1 crash-loop with
   no reason:
   - The messages use `safe_print`, which in a non-TTY process routes to
     `logging.getLogger("main").debug(...)` (see `safe_print`, ~line 196). At the default INFO
     level those lines are dropped.
   - `sys.exit(1)` raises `SystemExit`, which the outer catch-all in `main()`
     (`except Exception as e:` ~line 934, with `logger.error(..., exc_info=True)`) does **not**
     catch, because `SystemExit` derives from `BaseException`. So even the catch-all logs nothing.

Net effect: the process prints the OAuth-provider init line, then exits 1 silently a few seconds
in, repeatedly, until the old socket is released and a boot finally binds.

## Suggested fixes (any one helps; 1 + 2 together resolve it)

1. **Drop the redundant pre-flight bind** and let uvicorn own the bind, or set `SO_REUSEADDR`
   on the pre-flight socket so a lingering socket in a shared namespace does not block startup.
2. **Report the failure at ERROR** via the module logger (not `safe_print`/DEBUG), or raise
   instead of `sys.exit(1)` so the existing `except Exception` catch-all logs it. As written, a
   `SystemExit` at startup produces no diagnostic at all.
3. Optionally, **retry the bind with a short backoff** on transient `EADDRINUSE` at startup,
   which is common on fast in-place restarts.

## Why it matters

For a single-replica deployment this turns any brief restart into a multi-second outage with no
log explaining it — it presents as a mystery crash-loop. Making the bind resilient (1) removes
the crash-loop; making the error visible (2) means that if a bind genuinely fails, the operator
sees why.

stantheman0128:
## Description

Fixes the streamable-http pre-flight port bind that can false-fail on fast restart and crash-loop silently in Kubernetes / non-TTY environments (#946).

Two changes together:

1. **SO_REUSEADDR on both bind probes** (main streamable-http listener and the dual-transport workspace-cli sidecar), matching uvicorn/asyncio `reuse_address` so lingering TIME_WAIT sockets do not produce a false EADDRINUSE before the real server bind.
2. **Log bind failure at ERROR** via `report_preflight_bind_failure` (not only `safe_print`, which is DEBUG when stderr is not a TTY). That makes genuine bind failures visible in container logs. Confirmed at HEAD that the outer `except Exception` does not catch `SystemExit` (it subclasses `BaseException`); ERROR logging is what restores diagnostics.

Related narrower open PR: #744 (SO_REUSEADDR only, opened May). This PR is a completeness superset: same reuseaddr coverage on both probes, plus ERROR-level failure reporting and tests.

Fixes #946

## Type of Change
- [ ] Documentation update

## Testing

## Checklist

## AI assistance

This change was developed with AI assistance (Cursor / Grok), reviewed and verified on Windows 11 by Stan Shih (@stantheman0128).

## Evidence

**Unit tests** (`uv run pytest tests/test_main_port_probe.py -v`):

```
tests/test_main_port_probe.py::test_probe_tcp_bind_succeeds_on_free_port PASSED
tests/test_main_port_probe.py::test_probe_tcp_bind_sets_so_reuseaddr PASSED
tests/test_main_port_probe.py::test_probe_tcp_bind_raises_when_port_held PASSED
tests/test_main_port_probe.py::test_report_preflight_bind_failure_logs_at_error PASSED
tests/test_main_port_probe.py::test_outer_except_exception_does_not_catch_systemexit PASSED
============================== 5 passed in 5.35s ==============================
```

**Live non-TTY simulation** (stderr.isatty()=False, exclusive hold on a port, INFO root logger):

Before (safe_print only; invisible at INFO):
```
DEBUG:main:[MCP Server] Socket error: [WinError 10048] Only one usage of each socket address ...
DEBUG:main:[MCP Server] Port 57898 already in use
(no INFO/ERROR lines from safe_print above)
```

After (`report_preflight_bind_failure`):
```
ERROR:main:Socket error during pre-flight bind: [WinError 10013] An attempt was made to access a socket in a way forbidden by its access permissions
ERROR:main:Port 57898 is already in use. Cannot start HTTP server.
SO_REUSEADDR_SET=True
```

## What was not tested

- Full `workspace-mcp --transport streamable-http` boot against a real Kubernetes shared network namespace (no cluster in this environment). The probe and logging helpers used by that path are covered above.
- End-to-end uvicorn bind after the probe on macOS TIME_WAIT (issue/PR #744 manual scenario); SO_REUSEADDR is asserted in unit tests.

## Additional Notes

Allow edits from maintainers is enabled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup behavior when the main or HTTP port is already in use.
  * Standardized, more informative port-binding error messages, distinguishing occupied ports from other failures.
  * Improved handling of reusable TCP ports during server and HTTP sidecar startup.

* **Tests**
  * Added coverage for free, occupied, reusable, and temporary TCP ports.
  * Verified error logging distinguishes occupied ports from other binding failures.
  * Added regression coverage to ensure system-exit behavior remains unchanged.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,5 +1,6 @@
 import io
 import argparse
+import errno
 import json
 import logging
 import os
@@ -226,11 +227,23 @@ def report_preflight_bind_failure(port: int, error: OSError) -> None:
 
     safe_print alone routes to DEBUG in non-TTY / container environments, so
     operators would otherwise see only a silent exit-1 crash-loop.
+
+    Only label the failure as port contention for errno.EADDRINUSE; other
+    OSError values (permission, address family, unavailable address) keep a
+    generic bind-failure message that includes the original error.
     """
     logger.error("Socket error during pre-flight bind: %s", error)
-    logger.error("Port %s is already in use. Cannot start HTTP server.", port)
     safe_print(f"Socket error: {error}")
-    safe_print(f"❌ Port {port} is already in use. Cannot start HTTP server.")
+    if error.errno == errno.EADDRINUSE:
+        logger.error("Port %s is already in use. Cannot start HTTP server.", port)
+        safe_print(f"❌ Port {port} is already in use. Cannot start HTTP server.")
+    else:
+        logger.error(
+            "Failed to bind HTTP server on port %s: %s",
+            port,
+            error,
+        )
+        safe_print(f"❌ Failed to bind HTTP server on port {port}: {error}")
 
 
 def configure_safe_logging():
@@ -908,11 +921,18 @@ def main():
                     try:
                         # Same SO_REUSEADDR probe as streamable-http above.
                         probe_tcp_bind(http_host, http_port)
-                    except OSError:
-                        logger.warning(
-                            "Port %d in use, workspace-cli HTTP endpoint unavailable",
-                            http_port,
-                        )
+                    except OSError as e:
+                        if e.errno == errno.EADDRINUSE:
+                            logger.warning(
+                                "Port %d in use, workspace-cli HTTP endpoint unavailable",
+                                http_port,
+                            )
+                        else:
+                            logger.warning(
+                                "Failed to bind workspace-cli HTTP endpoint on port %d: %s",
+                                http_port,
+                                e,
+                            )
                         http_available = False
 
                     http_srv = None
--- a/main.py
+++ b/main.py
@@ -208,6 +208,31 @@ def safe_print(text):
         print(text.encode("ascii", errors="replace").decode(), file=sys.stderr)
 
 
+def probe_tcp_bind(host: str, port: int) -> None:
+    """Pre-flight TCP bind probe matching uvicorn/asyncio reuse_address behavior.
+
+    Sets SO_REUSEADDR so a lingering TIME_WAIT socket in a shared network
+    namespace (common on fast in-place container restarts) does not produce a
+    false EADDRINUSE before the real uvicorn bind runs. Raises OSError if the
+    address cannot be bound.
+    """
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        s.bind((host, port))
+
+
+def report_preflight_bind_failure(port: int, error: OSError) -> None:
+    """Log a failed pre-flight bind at ERROR (visible when stderr is not a TTY).
+
+    safe_print alone routes to DEBUG in non-TTY / container environments, so
+    operators would otherwise see only a silent exit-1 crash-loop.
+    """
+    logger.error("Socket error during pre-flight bind: %s", error)
+    logger.error("Port %s is already in use. Cannot start HTTP server.", port)
+    safe_print(f"Socket error: {error}")
+    safe_print(f"❌ Port {port} is already in use. Cannot start HTTP server.")
+
+
 def configure_safe_logging():
     """Replace console handlers with ASCII-safe formatters for Windows compatibility."""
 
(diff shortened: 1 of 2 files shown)


stantheman0128:
Addressed both CodeRabbit notes in efbde8d:

1. `report_preflight_bind_failure` and the sidecar probe path only say "already in use" for `errno.EADDRINUSE`; other `OSError` values get a generic bind failure that includes the original error.
2. Successful-probe tests now bind port `0` (OS ephemeral) instead of `_free_port()`, and there is a non-`EADDRINUSE` (`EACCES`) case that asserts the port-contention wording is not emitted.

`pytest tests/test_main_port_probe.py` : 6 passed.

MMannanAkmal:
Independent check of this branch at efbde8d against current main (54b1c56, v1.26.0):

- Merges cleanly, no conflicts, despite the 378 commits since the base.
- `uvx ruff@0.15.22 check` and `format --check` pass on main.py and tests/test_main_port_probe.py. The red ruff job here predates the RUFF_VERSION pin in .github/workflows/ruff.yml, so a push should turn it green.
- pytest passes, including tests/test_main_port_probe.py.
- Reproduced the original failure on Linux: with server-side TIME_WAIT sockets left on the listen port, a bare bind fails with EADDRINUSE and the pre-fix startup exits 1 with nothing in the log. probe_tcp_bind with SO_REUSEADDR binds in the same state, and a live listener is still rejected.

One gap: the tests here assert that SO_REUSEADDR is set, not that a TIME_WAIT port is accepted. I opened https://github.com/stantheman0128/google_workspace_mcp/pull/1 against this branch with a deterministic test for that case. Merging it there carries the proof along and re-triggers CI.

stantheman0128:
Merged the TIME_WAIT test from https://github.com/stantheman0128/google_workspace_mcp/pull/1 as 3304e9c.

That was the missing proof. Existing tests only asserted SO_REUSEADDR is set. The new one leaves a server-side TIME_WAIT, checks a bare bind still fails with EADDRINUSE, then expects probe_tcp_bind to succeed.

`pytest tests/test_main_port_probe.py`:
- Windows: 6 passed, 1 skipped (bare bind is not blocked here)
- WSL/Linux: 7 passed

`uvx ruff@0.15.22 check` and `format --check` are clean on the two files. This push should also pick up the ruff pin now on main.

PR #959, opened 2026-07-23
fix(sheets): reject non-letter column values in `_column_to_index`

### LINKED ISSUE #958: `_column_to_index` accepts any string — `delete_columns=["B2"]` silently deletes column AL
**Describe the bug**

`_column_to_index` in `gsheets/sheets_helpers.py` returns `None` only for an empty string. Every
other input is fed through the base-26 arithmetic and yields a plausible-looking index, so a cell
reference or a range resolves to a real — but completely unrelated — column.

```python
def _column_to_index(column: str) -> Optional[int]:
    """Convert column letters (A, B, AA) to zero-based index."""
    if not column:
        return None
    result = 0
    for char in column.upper():
        result = result * 26 + (ord(char) - ord("A") + 1)
    return result - 1
```

All five call sites in `gsheets/sheets_tools.py` (lines 1571, 1762, 1831, 1996, 2114) guard the
result with `if col_idx is None: raise UserInputError(...)`. Because the helper can never return
`None` for a non-empty string, **those guards are unreachable** and the bad index flows straight
into the request body.

The worst case is `resize_sheet_dimensions(delete_columns=[...])`: it builds a `deleteDimension`
request against the wrong index, permanently deletes a column the caller never named, and reports
success — `"deleted columns: B2"`.

**Startup Logs**

N/A — this does not depend on server configuration or transport. The defect is in a pure helper
function and reproduces without starting the server (see below).

**To Reproduce**

1. Open a spreadsheet that has data in column AL.
2. Call `resize_sheet_dimensions(spreadsheet_id="<id>", delete_columns=["B2"])` — i.e. pass a cell
   reference where a bare column letter was expected.
3. The tool returns success: `"deleted columns: B2"`.
4. Column **AL** has been permanently deleted. Column B is untouched.

The underlying helper can be checked directly, with no credentials and no server:

```python
from gsheets.sheets_helpers import _column_to_index
```

| input      | returned index | column actually targeted |
|------------|----------------|--------------------------|
| `"A"`      | 0              | A (correct)              |
| `"AA"`     | 26             | AA (correct)             |
| `""`       | `None`         | rejected (correct)       |
| `"A1"`     | 10             | **K**                    |
| `"B2"`     | 37             | **AL**                   |
| `"C3"`     | 64             | **BM**                   |
| `"A:B"`    | 521            | **TB**                   |
| `"Sheet1"` | 229493716      | **SHEESK**               |
| `"5"`      | **-12**        | negative `startIndex`    |

**Expected behavior**

Anything that is not a bare column letter should be rejected with the `UserInputError` the call
sites already try to raise, and no `batchUpdate` request should be issued.

**Screenshots**

N/A.

- OS: not platform-specific (reproduced on Linux)
- Container: N/A — reproduces in any deployment
- Version: 1.22.1 (`main`, `b92f5b7`)

**Additional context**

These tools are driven by an LLM. Writing `"B2"` or `"A:B"` where a bare column letter is expected
is a very ordinary model mistake, and the tool description does not constrain the format. The
failure is silent and destructive: the user loses a column of real data, and the tool response tells
them the operation succeeded. It's the same class of problem as #923 (`update_script_content`
silently deleting omitted files).

The other four call sites are less severe but still wrong: `insert_columns_at` inserts at the wrong
position (shifting every column to its right), and `column_sizes` / `auto_resize_columns` /
`hide_columns` / `unhide_columns` silently act on an unrelated column.

The existing test `test_invalid_column_letter_raises_error` only covers `column_sizes={"": 200}` —
the one input the old code did reject — which is probably why this went unnoticed.

Happy to open a PR: a format check in the helper fixes all five call sites at once and makes the
guards that are already there start firing.

Unrelated but adjacent, spotted in the same function — happy to file separately if useful:
`delete_rows` / `delete_columns` are sorted descending but never de-duplicated, so
`delete_rows=[5, 5]` deletes row 5 and then the row that shifted up into its place, destroying two
rows while reporting `"deleted rows: 5, 5"`.

glitch-ux:
## Description

Fixes #958.

`_column_to_index` (`gsheets/sheets_helpers.py`) returned `None` only for an empty string. Any other
input ran through the base-26 arithmetic and produced a plausible-looking index:

```python
if not column:
    return None
result = 0
for char in column.upper():
    result = result * 26 + (ord(char) - ord("A") + 1)
return result - 1
```

| input   | index | column targeted |
|---------|-------|-----------------|
| `"A1"`  | 10    | K               |
| `"B2"`  | 37    | **AL**          |
| `"C3"`  | 64    | BM              |
| `"A:B"` | 521   | TB              |
| `"5"`   | -12   | negative index  |

All five call sites in `gsheets/sheets_tools.py` (1571, 1762, 1831, 1996, 2114) guard with
`if col_idx is None: raise UserInputError(...)`, so every one of those guards was unreachable.
The practical consequence is that `resize_sheet_dimensions(delete_columns=["B2"])` issues a
`deleteDimension` for index 37 and **permanently deletes column AL**, then reports success.

## Fix

Validate the format in the helper, so the guards the call sites already have start firing:

```python
COLUMN_LETTERS_RE = re.compile(r"^[A-Za-z]+$")

if not column or not COLUMN_LETTERS_RE.match(column):
    return None
```

Deliberately scoped to the helper — no call-site changes, so all five paths are fixed at once and
the existing per-site error messages are preserved. `_parse_a1_part` is unaffected: it only ever
passes the `[A-Za-z]*` capture group from `A1_PART_REGEX`.

## Type of Change

- [ ] Documentation update

## Testing

- [ ] I have tested this change manually — *not against a live spreadsheet (no credentials); the
  defect and the fix are both verified against the shipped helper and through
  `_resize_sheet_dimensions_impl` with a mocked Sheets service, per the repo's
  "mock Google APIs, never hit live services" rule.*

Added to `tests/gsheets/test_resize_sheet_dimensions.py`:

- `test_delete_columns_rejects_non_letter_values` — parametrized over `B2`, `A1`, `C3`, `5`, `A:B`,
  `Sheet1`; asserts `UserInputError` **and** that no `batchUpdate` call is made, since the point is
  that nothing gets deleted.
- `test_insert_columns_at_rejects_cell_reference` — the other destructive/positional call site.
- `test_multi_letter_column_still_resolves` — pins `"AA"` → `startIndex` 26 so the new check can't
  over-reject.

Note the pre-existing `test_invalid_column_letter_raises_error` only covered `column_sizes={"": 200}`,
i.e. the one input the old code did reject — which is why this went unnoticed.

Verification:

```
# all 7 new tests fail on main without the source change
7 failed, 29 passed

# with the fix
1291 passed, 2 skipped, 3 warnings in 7.38s   (baseline on main: 1283 passed, 2 skipped)

uvx ruff check          -> All checks passed!
uvx ruff format --check -> 2 files already formatted
```

## Checklist

## Additional Notes

Scope kept deliberately narrow. Two adjacent things I did **not** touch, happy to follow up on
either if you'd like:

- `delete_rows` / `delete_columns` sort descending but never de-duplicate, so `delete_rows=[5, 5]`
  deletes two rows while reporting `"deleted rows: 5, 5"`.
- No upper bound on the resulting index (Sheets caps at column ZZZ). Out-of-range values fail loudly
  at the API rather than corrupting data, so I left it alone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for spreadsheet column inputs to ensure only bare column letters (e.g., `A`, `AA`) are accepted.
  * Malformed values like cell references (`B2`), ranges, numeric strings, and other non-column fragments are now rejected instead of being misinterpreted.
  * Column operations now avoid sending spreadsheet update requests when validation fails, preventing unintended changes.

* **Tests**
  * Added coverage for invalid/valid column formats and confirmed no update requests are made on failure.

### DIFF
--- a/gsheets/sheets_helpers.py
+++ b/gsheets/sheets_helpers.py
@@ -29,7 +29,7 @@ def _column_to_index(column: str) -> Optional[int]:
     # check the arithmetic below still produces a plausible index -- "B2" gives
     # 37, i.e. column AL -- and the callers' `if col_idx is None` guards never
     # fire, so a delete_columns request silently targets a column nobody named.
-    if not column or not COLUMN_LETTERS_RE.match(column):
+    if not column or not COLUMN_LETTERS_RE.fullmatch(column):
         return None
     result = 0
     for char in column.upper():
--- a/gsheets/sheets_helpers.py
+++ b/gsheets/sheets_helpers.py
@@ -19,11 +19,17 @@ MAX_GRID_METADATA_CELLS = 5000
 
 A1_PART_REGEX = re.compile(r"^([A-Za-z]*)(\d*)$")
 SHEET_TITLE_SAFE_RE = re.compile(r"^[A-Za-z0-9_]+$")
+COLUMN_LETTERS_RE = re.compile(r"^[A-Za-z]+$")
 
 
 def _column_to_index(column: str) -> Optional[int]:
     """Convert column letters (A, B, AA) to zero-based index."""
-    if not column:
+    # Callers pass raw tool arguments straight through, so a cell reference or a
+    # range can arrive where a bare column letter was expected. Without a format
+    # check the arithmetic below still produces a plausible index -- "B2" gives
+    # 37, i.e. column AL -- and the callers' `if col_idx is None` guards never
+    # fire, so a delete_columns request silently targets a column nobody named.
+    if not column or not COLUMN_LETTERS_RE.match(column):
         return None
     result = 0
     for char in column.upper():
--- a/tests/gsheets/test_resize_sheet_dimensions.py
+++ b/tests/gsheets/test_resize_sheet_dimensions.py
@@ -583,7 +583,7 @@ async def test_delete_rows_non_integer_value_raises_user_input_error():
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("column", ["B2", "A1", "C3", "5", "A:B", "Sheet1"])
+@pytest.mark.parametrize("column", ["B2", "A1", "C3", "5", "A:B", "Sheet1", "AA\n"])
 async def test_delete_columns_rejects_non_letter_values(column):
     """Test delete_columns rejects values that are not bare column letters."""
     mock_service = create_mock_service()
--- a/tests/gsheets/test_resize_sheet_dimensions.py
+++ b/tests/gsheets/test_resize_sheet_dimensions.py
@@ -6,6 +6,7 @@ Tests column/row resizing, auto-resize, freeze, and hide/unhide operations.
 
 import pytest
 from unittest.mock import Mock
+import re
 import sys
 import os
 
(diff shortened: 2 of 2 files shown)


glitch-ux:
Good catch from CodeRabbit — fixed with .fullmatch() and added the "AA\n" case in 45282d3.

taylorwilsdon:
Merged #964 thanks

PR #964, opened 2026-07-25
fix(gsheets): reject cell references in column letter parameters

### LINKED ISSUE #958: `_column_to_index` accepts any string — `delete_columns=["B2"]` silently deletes column AL

syf2211:
## Summary

Validate `_column_to_index` inputs so only bare column letters (e.g. `A`, `AA`) are accepted. Cell references like `B2` or ranges like `A:B` now return `None`, allowing existing `UserInputError` guards in `resize_sheet_dimensions` and related tools to reject bad input before any `batchUpdate` is issued.

## Motivation

Fixes #958. When an LLM passes a cell reference where a bare column letter is expected (e.g. `delete_columns=["B2"]`), the old helper silently resolved to column **AL** (index 37) and permanently deleted the wrong column while reporting success.

## Changes

- Add `COLUMN_LETTER_REGEX` and reject non-letter inputs in `_column_to_index`
- Add unit tests in `tests/gsheets/test_sheets_helpers.py`
- Add integration regression test ensuring `delete_columns=["B2"]` raises `UserInputError` without calling the API

## Tests

```bash
python3 -m pytest tests/gsheets/test_sheets_helpers.py tests/gsheets/test_resize_sheet_dimensions.py -q
```

- 31/31 passed

## Notes

- No behavioral change for valid bare column letters (`A`, `Z`, `AA`, `AL`, etc.)
- `_parse_a1_part` is unaffected — it still extracts only the letter group from A1 notation before calling `_column_to_index`

Fixes #958

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for spreadsheet column-letter inputs, rejecting values with trailing whitespace/newlines and non-letter formats.
  * Prevents cell-style references (e.g., column+row) from being accepted where only bare column letters are expected.
  * Improved handling of A1-style parts and safer quoting for sheet titles, including titles containing trailing newlines.
* **Tests**
  * Added/expanded coverage for the stricter validation and quoting behavior.

### DIFF
--- a/gsheets/sheets_helpers.py
+++ b/gsheets/sheets_helpers.py
@@ -24,7 +24,7 @@ COLUMN_LETTER_REGEX = re.compile(r"^[A-Za-z]+$")
 
 def _column_to_index(column: str) -> Optional[int]:
     """Convert column letters (A, B, AA) to zero-based index."""
-    if not column or not COLUMN_LETTER_REGEX.match(column):
+    if not column or not COLUMN_LETTER_REGEX.fullmatch(column):
         return None
     result = 0
     for char in column.upper():
@@ -40,7 +40,7 @@ def _parse_a1_part(
     Supports anchors like '$A$1' by stripping the dollar signs.
     """
     clean_part = part.replace("$", "")
-    match = pattern.match(clean_part)
+    match = pattern.fullmatch(clean_part)
     if not match:
         raise UserInputError(f"Invalid A1 range part: '{part}'.")
     col_letters, row_digits = match.groups()
@@ -173,7 +173,7 @@ def _quote_sheet_title_for_a1(sheet_title: str) -> str:
     If the sheet title contains special characters or spaces, it is wrapped in single quotes.
     Any single quotes in the title are escaped by doubling them, as required by Google Sheets.
     """
-    if SHEET_TITLE_SAFE_RE.match(sheet_title or ""):
+    if SHEET_TITLE_SAFE_RE.fullmatch(sheet_title or ""):
         return sheet_title
     escaped = (sheet_title or "").replace("'", "''")
     return f"'{escaped}'"
--- a/gsheets/sheets_helpers.py
+++ b/gsheets/sheets_helpers.py
@@ -19,11 +19,12 @@ MAX_GRID_METADATA_CELLS = 5000
 
 A1_PART_REGEX = re.compile(r"^([A-Za-z]*)(\d*)$")
 SHEET_TITLE_SAFE_RE = re.compile(r"^[A-Za-z0-9_]+$")
+COLUMN_LETTER_REGEX = re.compile(r"^[A-Za-z]+$")
 
 
 def _column_to_index(column: str) -> Optional[int]:
     """Convert column letters (A, B, AA) to zero-based index."""
-    if not column:
+    if not column or not COLUMN_LETTER_REGEX.match(column):
         return None
     result = 0
     for char in column.upper():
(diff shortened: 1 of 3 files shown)


taylorwilsdon:
Thanks!

PR #976, opened 2026-07-27
Fix inspect_doc_structure for tabbed Docs without tab_id

### LINKED ISSUE #975: inspect_doc_structure returns zero elements for tabbed Google Docs without tab_id
## Summary

`inspect_doc_structure` can return an empty body structure for tabbed Google Docs when `tab_id` is not provided, even though the document has content and the tool output includes available tabs.

This is important because the tool documentation tells callers to use `inspect_doc_structure` to find safe insertion indexes. When it reports `total_elements: 0` / `total_length: 0`, downstream agents can choose unsafe low indexes such as `1`, causing inserts to land at the top of a document instead of the intended append position.

## Version tested

- `google_workspace_mcp` v1.22.2
- Commit: `59b534201fd16f58b175fc564df6adf4f6b4de71`
- Transport: streamable HTTP
- Auth mode: external OAuth bearer-token mode

## Reproduction shape

1. Use a native Google Doc that has tab metadata/content. A single visible tab is enough.
2. Call `inspect_doc_structure` without `tab_id`, for example:

```json
{
  "document_id": "<redacted>",
  "detailed": true
}
```

3. Observe output similar to:

```json
{
  "total_elements": 0,
  "paragraphs": 0,
  "total_length": 0,
  "tabs": [
    { "title": "Tab 1", "tab_id": "t.0" }
  ]
}
```

4. Call `get_doc_as_markdown` for the same document. It returns the actual document content.
5. If `inspect_doc_structure` is called with the concrete `tab_id`, the tab body can be inspected correctly.

## Likely cause

Google Docs API tabbed content lives under `tabs[].documentTab.body` when `documents.get(..., includeTabsContent=true)` is used. The top-level `document.body` can be empty/ambiguous for tabbed documents.

In `gdocs/docs_tools.py`, `inspect_doc_structure` currently initializes:

```python
target_content = doc.get("body", {})
```

When no `tab_id` is supplied, the function later finds the first document tab for headers/footers/documentStyle, but does not appear to also set `target_content` / `analysis_doc["body"]` from the first `documentTab.body` before calling `parse_document_structure(analysis_doc)`.

So the returned structure can be based on an empty top-level body while `tabs` correctly reports that document tabs exist.

## Expected behavior

When `tab_id` is omitted for a tabbed Google Doc, one of these would be safer:

- Default to the first document tab and set `analysis_doc["body"]` from `first_tab_doc["body"]` before parsing.
- Or return per-tab structures so callers can choose the correct tab explicitly.

In either case, the output should make clear which tab was inspected, e.g. `inspected_tab_id` / `inspected_tab_title`, so callers do not treat a tab-specific index as global.

## Why this matters

`inspect_doc_structure` is the main safety tool for discovering insertion positions before calls such as `modify_doc_text`, `create_table_with_data`, or `batch_update_doc`. Returning zero length for a non-empty tabbed doc can lead to invalid or wrong-position writes.

## References

Google's Docs API tabs guide notes that tab content is included under `document.tabs` / `documentTab` when `includeTabsContent=true` is used:

https://developers.google.com/workspace/docs/api/how-tos/tabs

Ofori01:
## Description

Fixes `inspect_doc_structure` for tabbed Google Docs when callers do not provide `tab_id`.

When `documents.get(..., includeTabsContent=True)` returns tabbed document content under `tabs[].documentTab.body`, the current no-`tab_id` path copies the first tab's headers/footers/style for analysis but leaves `analysis_doc["body"]` set to the top-level document body. For tabbed Docs, that top-level body can be empty, so the tool can report `total_length: 0` and zero elements even when the first tab has body content.

This PR makes the existing first-document-tab fallback also use the first tab's body and named ranges before parsing structure.

Fixes #975.

## Type of Change

- [ ] Documentation update

## Testing

Commands run:

```bash
uv run pytest tests/gdocs/test_header_footer_guardrails.py -q
uv run ruff check gdocs/docs_tools.py tests/gdocs/test_header_footer_guardrails.py
```

## Checklist

.

### DIFF
--- a/gdocs/docs_tools.py
+++ b/gdocs/docs_tools.py
@@ -1462,6 +1462,8 @@ async def inspect_doc_structure(
 
         first_tab_doc = first_document_tab(doc.get("tabs", []))
         if first_tab_doc:
+            analysis_doc["body"] = first_tab_doc.get("body", {})
+            analysis_doc["namedRanges"] = first_tab_doc.get("namedRanges", {})
             analysis_doc["headers"] = first_tab_doc.get("headers", {})
             analysis_doc["footers"] = first_tab_doc.get("footers", {})
             analysis_doc["documentStyle"] = first_tab_doc.get("documentStyle", {})
--- a/tests/gdocs/test_header_footer_guardrails.py
+++ b/tests/gdocs/test_header_footer_guardrails.py
@@ -338,7 +338,23 @@ class TestHeaderFooterGuardrails:
                 {
                     "tabProperties": {"tabId": "t.0"},
                     "documentTab": {
-                        "body": {"content": []},
+                        "body": {
+                            "content": [
+                                {
+                                    "startIndex": 1,
+                                    "endIndex": 19,
+                                    "paragraph": {
+                                        "elements": [
+                                            {
+                                                "textRun": {
+                                                    "content": "Tabbed body text"
+                                                }
+                                            }
+                                        ]
+                                    },
+                                }
+                            ]
+                        },
                         "headers": {
                             "hdr-tab-1": {
                                 "content": [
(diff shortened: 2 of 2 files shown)


Ofori01:
@taylorwilsdon could you review this when you have a chance? It fixes the tabbed Docs structure issue reported in #975.

taylorwilsdon:
Thanks!

PR #977, opened 2026-07-28
feat(sheets): add manage_sheet_table for native Sheets tables

alexnbg88:
## Description

Adds `manage_sheet_table`, which wraps the Sheets `addTable` / `updateTable` / `deleteTable` requests.

The toolset can already read native tables (`list_sheet_tables`) and append rows to them (`append_table_rows`), but there is no way to create one. Both existing tools need a `tableId` that already exists, so a table has to be made by hand in the Sheets UI before either becomes usable. This closes that gap.

**Actions**

- **create** — name a range as a table, with optional typed columns and banded row colors. The range must include the header row.
- **update** — rename, resize, retype or recolor an existing table.
- **delete** — remove a table by id.

**Details worth calling out**

- The `update` field mask is built from the arguments actually supplied, never sent as `"*"`. An update that only renames a table cannot silently clear its column properties.
- `columnIndex` is table-relative, not sheet-relative. It is inferred from list position when omitted, since that is the part of this API most easily got wrong by hand.
- `DROPDOWN` columns take a `values` list and build a `TableColumnDataValidationRule` with a `ONE_OF_LIST` condition. The API only permits column validation on `DROPDOWN` columns, so supplying `values` for any other type is rejected up front with a message explaining why, instead of being sent and coming back as an opaque 400.
- All argument validation happens before any network call.

**Tier registration**

Registered under `sheets: complete` in `tool_tiers.yaml`. I've also added `tests/gsheets/test_sheets_tool_tiers.py`, because a tool registered with `@server.tool` but missing from `tool_tiers.yaml` is silently filtered out of every `--tool-tier` deployment with no startup error or warning. I hit exactly that while writing this. The test now catches it for any future Sheets tool, and catches the reverse case of a tier entry naming a tool that no longer exists.

**Enum drift guard**

`TABLE_COLUMN_TYPES` is an allowlist, so it is only useful while it matches the API. `test_table_column_types_drift.py` pins it to the discovery document bundled with `google-api-python-client`, treating the two drift directions differently on purpose:

- A value we allow that the API does not define **fails** the test. It defeats the allowlist and produces exactly the 400 the allowlist exists to prevent.
- A type the API has added that we do not list only **warns**. That is a missing feature rather than a defect, and failing on it would break routine dependency-bump PRs.

Happy to drop either test file if you'd rather not have them.

## Type of Change

- [ ] Documentation update

## Testing

21 new tests covering create/update/delete, column typing and index inference, dropdown rule construction, banding colors, JSON-encoded input, field-mask narrowness, and argument validation.

Manually verified against the live Sheets API on two Google accounts: create a table, read it back and confirm the table, the `CURRENCY` column, the `DROPDOWN` column and its stored choices; rename via `updateTable` and confirm the column properties survived the narrow field mask; delete the table and confirm removal. 14/14 checks on each account.

**Note on the existing suite:** `tests/core/test_telemetry.py` has 7 failures on a clean checkout of `main` (v1.22.2) in my environment, before any of my changes. This PR does not affect them. Everything else passes: 1365 passed with that file excluded. `ruff check` and `ruff format --check` are clean.

## Checklist

## Additional Notes

Purely additive: 1007 insertions, 0 deletions, no existing behavior changed.

Example — a client pipeline with a currency column and a status dropdown:

```
manage_sheet_table(
    action="create",
    spreadsheet_id="...",
    table_name="Pipeline",
    range_name="Sheet1!A1:C50",
    column_properties=[
        {"columnName": "Client", "columnType": "TEXT"},
        {"columnName": "Fee",    "columnType": "CURRENCY"},
        {"columnName": "Stage",  "columnType": "DROPDOWN",
         "values": ["Open", "Won", "Lost"]},
    ],
    header_color="#4285f4",
)
```

## Summary by CodeRabbit

* **New Features**
  * Added support for creating, updating, and deleting native Google Sheets tables, including typed columns and dropdown validation.
  * Added table configuration for naming/ranges and header/footer/banding row colors.
  * Enabled the table management capability in the Sheets “complete” tool tier.

* **Bug Fixes**
  * Improved input validation for actions, required parameters, column types, column indexes, dropdown values, and color handling.

* **Tests**
  * Added extensive tests covering request payloads, field update behavior, error cases, tool-tier consistency, and column type drift protection.

### DIFF
--- a/core/tool_tiers.yaml
+++ b/core/tool_tiers.yaml
@@ -95,6 +95,7 @@ sheets:
     - list_spreadsheet_comments
     - manage_spreadsheet_comment
     - manage_conditional_formatting
+    - manage_sheet_table
 
 chat:
   core:
--- a/gsheets/sheets_helpers.py
+++ b/gsheets/sheets_helpers.py
@@ -1127,3 +1127,148 @@ async def _fetch_grid_metadata(
         )
 
     return hyperlink_section, notes_section
+
+
+# ---------------------------------------------------------------------------
+# Native Sheets tables (addTable / updateTable / deleteTable)
+# ---------------------------------------------------------------------------
+
+# Column types accepted by TableColumnProperties.columnType. Mirrors the
+# Sheets v4 ColumnType enum; COLUMN_TYPE_UNSPECIFIED is deliberately omitted
+# because it is the "do not use" default.
+TABLE_COLUMN_TYPES = {
+    "DOUBLE",
+    "CURRENCY",
+    "PERCENT",
+    "DATE",
+    "TIME",
+    "DATE_TIME",
+    "TEXT",
+    "BOOLEAN",
+    "DROPDOWN",
+    "FILES_CHIP",
+    "PEOPLE_CHIP",
+    "FINANCE_CHIP",
+    "PLACE_CHIP",
+    "RATINGS_CHIP",
+}
+
+TABLE_ROW_COLOR_FIELDS = {
+    "header_color": "headerColorStyle",
+    "footer_color": "footerColorStyle",
+    "first_band_color": "firstBandColorStyle",
+    "second_band_color": "secondBandColorStyle",
+}
+
+
+def _parse_table_column_properties(
+    column_properties: Optional[Union[str, List[dict]]],
+) -> Optional[List[dict]]:
+    """
+    Normalize table column definitions into Sheets TableColumnProperties.
+
+    Accepts a list of dicts or a JSON-encoded list. Each entry supports:
+      - columnName (str): the column header label.
+      - columnType (str): one of TABLE_COLUMN_TYPES.
+      - columnIndex (int, optional): table-relative index; defaults to position.
+      - values (list, optional): dropdown choices. Only valid for DROPDOWN,
+        where the API accepts a validation rule whose condition is ONE_OF_LIST.
+    """
+    if column_properties is None:
+        return None
+
+    parsed = column_properties
+    if isinstance(parsed, str):
+        try:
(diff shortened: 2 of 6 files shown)


alexnbg88:
Thanks — one fixed, one intentional.

**Fixed: the `replies` indexing.** Good catch, this was a real bug. `response.get("replies", [{}])[0]` reads as defensive but isn't: the dict default only applies when the key is *absent*, so `{"replies": []}` slipped straight through to `[0]` and raised `IndexError`. The neighbouring shapes (`{}`, `[{}]`, `[{"addTable": {}}]`) were all handled, which is exactly why it went unnoticed — the test mock always returned a populated reply. Now reads the first reply defensively and degrades to `table_id=None`, with a parametrised test over all four shapes. I confirmed the empty-`replies` case fails with `IndexError` before the change and passes after, while the other three pass throughout.

**Not changing: the `parsed_columns` asymmetry.** The difference between `if parsed_columns:` on create and `if parsed_columns is not None:` on update is deliberate, and aligning them would be a small regression.

On **update**, `None` and `[]` must mean different things: `None` leaves column properties untouched, `[]` clears them. That distinction is what drives the field mask — `columnProperties` is only added to `fields` when the caller actually supplied something, so an update that only renames a table can't wipe its columns.

On **create** there is no prior state to preserve, so `None` and `[]` both mean "no typed columns". Switching to `is not None` would send `columnProperties: []` to `addTable`, which is a meaningless empty array in the request.

I've added a comment at the create branch explaining this, since the asymmetry is a reasonable thing to trip over when reading the two branches side by side.

Issue #915, opened 2026-07-10
Feature: permanent Drive file delete (files.delete) — a hard-delete tool distinct from trash-via-update_drive_file


szhygulin:
## Summary

Add a Drive tool to **permanently** delete a file — `service.files().delete(fileId=...)` — i.e. a hard-delete that bypasses Trash. Today `update_drive_file(trashed=True)` gives a reversible *soft* delete (→ Trash), but there is no way to permanently remove a specific file programmatically.

## Background / why this isn't #133

#133 ("Why no delete for all tools") established that delete for Gmail/Calendar/Tasks is handled via soft-delete *modify* calls. For Drive that soft-delete equivalent — `update_drive_file` with `trashed=True` — already exists and works well.

The gap here is the **hard** delete. A trashed Drive file lingers in Trash until Google's ~30-day auto-purge, and the only way to remove it sooner is Drive's "empty trash," which is all-or-nothing (it purges *every* trashed file, not a specific one). There is currently no per-file permanent delete anywhere in the Drive tool surface — the only `files().<op>` calls in `gdrive/drive_tools.py` are `list` / `update` / `export_media` / `get_media`.

## Request

A new tool (e.g. `delete_drive_file`) that calls:

```python
service.files().delete(fileId=file_id, supportsAllDrives=True).execute()
```

returning success / not-found. It would mirror `update_drive_file`'s structure (single `fileId`, shared-drive support). No new OAuth scope is needed — the existing `https://www.googleapis.com/auth/drive` scope already permits `files.delete`.

## Safety / opt-in

`files.delete` is irreversible — no Trash, not recoverable — so it deserves more friction than the reversible trash path:

- The tool description should state the irreversibility explicitly and point to `update_drive_file(trashed=True)` as the reversible default.
- It fits an MCP `destructiveHint: true` annotation (cf. #609) and/or placement in a high/destructive tool tier so it stays opt-in rather than in the default surface.

## Notes

- The `delete` verb is already used in the Drive tool file for `permissions().delete` (ACL revoke), so the client/error-handling patterns are already present.
- Scope is one new tool; the reversible `update_drive_file(trashed=True)` remains the recommended default.

Thanks for the great project!

taylorwilsdon:
Hmm... honestly, this scares me a little because not every client respects destructiveHint effectively and the current delete still deletes it, just requires a human to perma-purge. Will mull this over and curious what other folks think.

PR #978, opened 2026-07-28
fix(drive): exclude trashed files from search_drive_files by default

martiant:
## Description

`search_drive_files` returns files that are in the trash. `list_drive_items` does not — it builds `f"'{folder_id}' in parents and trashed=false"` ([drive_tools.py](https://github.com/taylorwilsdon/google_workspace_mcp/blob/main/gdrive/drive_tools.py)) — and neither does the Drive web UI. Search is the outlier because it passes the caller's query through verbatim and appends nothing.

The disagreement is actively misleading rather than merely inconsistent. `update_drive_file(trashed=True)` is the only delete path exposed over MCP, so the natural verification loop is trash → search → confirm gone. Today that loop reports failure on a successful trash, and I've watched an agent conclude from it that the trash write path was broken when it was fine.

Verified live against v1.22.2, file `1Nf2i87…`:

- trashed via `update_drive_file(trashed=true)`; Drive API confirmed `{"trashed": true, "explicitlyTrashed": true}`
- `search_drive_files("name contains 'trash-test'")` → **still returns the file**
- `search_drive_files("name contains 'trash-test' and trashed=false")` → `No files found`

### The change

- New `include_trashed: bool = False` parameter on `search_drive_files`.
- When it's `False`, `and trashed=false` is appended to the final query.
- The query is user-supplied and may already contain its own `trashed = true/false` clause, so the existing trashed regex in `DRIVE_QUERY_PATTERNS` is hoisted to `TRASHED_CLAUSE_PATTERN` and reused to detect that case. An explicit clause always wins and is never double-added.
- `include_trashed=True` restores the previous pass-through behavior exactly.
- Ordering is safe with the existing `file_type` filter: `((fullText contains 'x') and trashed=false) and mimeType = '…'`.

**Behavior change, intentional:** callers who relied on the old implicit inclusion of trashed results now need `include_trashed=True` or an explicit `trashed = true` clause. Happy to flip the default to `True` and make this purely additive if you'd rather not change existing behavior in a point release — but I think matching `list_drive_items` and the Drive UI is the right default, and the current behavior mostly surprises people.

## Type of Change
- [x] Documentation update

## Testing

7 new tests in `tests/gdrive/test_drive_tools.py`:

| Test | Asserts |
|------|---------|
| `…excludes_trashed_by_default` | free-text → `(fullText contains 'budget') and trashed=false` |
| `…excludes_trashed_for_structured_query` | structured → `(name contains 'report') and trashed=false` |
| `…include_trashed_leaves_query_untouched` | opt-in → query passed through verbatim |
| `…respects_explicit_trashed_clause` (×3) | `trashed=true`, `trashed = true`, `TRASHED=FALSE` all pass through unchanged |
| `…trashed_filter_composes_with_file_type` | trashed + mimeType filters compose correctly |

```
uv run pytest tests/gdrive tests/core -q
232 passed
```

`uvx ruff check` error count is unchanged from `main` (no new lint findings); `uvx ruff format --check` is clean on all touched files.

## Checklist

## Additional Notes

Related to #915 (permanent Drive file delete) — same corner of the API. That issue is about the missing hard-delete path; this PR is about the observation path lying about the soft-delete path that already exists.

Docs updated: the `search_drive_files` parameter table and the "Drive Search Query Operators" section in `skills/managing-google-workspace/references/drive.md`.

## Summary by CodeRabbit

* **New Features**
  * Drive searches now exclude trashed files by default.
  * Added an `include_trashed` option to include trashed items.
  * If the query contains an explicit `trashed = true/false` predicate, it overrides the flag.
* **Bug Fixes**
  * Improved handling of `trashed=...` inside quoted text so it’s treated as data, not a filter.
* **Documentation**
  * Updated Drive search docs with “Trash” guidance and override behavior.
* **Tests**
  * Expanded coverage for default trash filtering, explicit predicate variations, and quoted edge cases.

### DIFF
--- a/gdrive/drive_helpers.py
+++ b/gdrive/drive_helpers.py
@@ -175,6 +175,10 @@ def format_permission_info(permission: Dict[str, Any]) -> str:
     return base
 
 
+# Matches an explicit trashed clause anywhere in a Drive query, e.g. "trashed = true".
+# Used both for structured-query detection and to avoid double-adding a trashed filter.
+TRASHED_CLAUSE_PATTERN = re.compile(r"\btrashed\s*=\s*(true|false)\b", re.IGNORECASE)
+
 # Precompiled regex patterns for Drive query detection
 DRIVE_QUERY_PATTERNS = [
     re.compile(r'\b\w+\s*(=|!=|>|<)\s*[\'"].*?[\'"]', re.IGNORECASE),  # field = 'value'
@@ -182,7 +186,7 @@ DRIVE_QUERY_PATTERNS = [
     re.compile(r"\bcontains\b", re.IGNORECASE),  # contains operator
     re.compile(r"\bin\s+parents\b", re.IGNORECASE),  # in parents
     re.compile(r"\bhas\s*\{", re.IGNORECASE),  # has {properties}
-    re.compile(r"\btrashed\s*=\s*(true|false)\b", re.IGNORECASE),  # trashed=true/false
+    TRASHED_CLAUSE_PATTERN,  # trashed=true/false
     re.compile(r"\bstarred\s*=\s*(true|false)\b", re.IGNORECASE),  # starred=true/false
     re.compile(
         r'[\'"][^\'"]+[\'"]\s+in\s+parents', re.IGNORECASE
--- a/gdrive/drive_helpers.py
+++ b/gdrive/drive_helpers.py
@@ -176,8 +176,10 @@ def format_permission_info(permission: Dict[str, Any]) -> str:
 
 
 # Matches an explicit trashed clause anywhere in a Drive query, e.g. "trashed = true".
-# Used both for structured-query detection and to avoid double-adding a trashed filter.
-TRASHED_CLAUSE_PATTERN = re.compile(r"\btrashed\s*=\s*(true|false)\b", re.IGNORECASE)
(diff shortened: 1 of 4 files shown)


martiant:
Addressed the review finding in 42a4675.

`name contains 'trashed=false'` — a search for a *filename* containing that text — matched the raw `TRASHED_CLAUSE_PATTERN`, so the default filter was skipped and trashed files came back despite `include_trashed=False`. That is the bug this PR fixes, reintroduced through a legal query. Good catch.

`has_explicit_trashed_clause()` now blanks out Drive query string literals (both quote styles, backslash escapes honoured) before looking for the predicate, and the skip decision uses it. A real clause still wins when a quoted look-alike sits beside it: `trashed=true and name contains 'trashed=false'` is correctly treated as explicit.

`TRASHED_CLAUSE_PATTERN` stays in `DRIVE_QUERY_PATTERNS` unchanged — a false positive there is harmless, since any query carrying a quoted value already matches several other patterns and is structured either way.

4 more tests (11 total): three quoted-look-alike spellings through `search_drive_files`, plus a direct unit test of the detector covering escapes and the both-present case. `uv run pytest tests/gdrive tests/core -q` → 236 passed. `uvx ruff check` error count still identical to `main`; `ruff format --check` clean.

taylorwilsdon:
Good catch, thanks! Will get this merged.

PR #980, opened 2026-07-28
fix(gcalendar): emit the same event metadata from both get_events branches

wdrwilson:
## Description

`get_events` formats its detailed output in two separate blocks, and they'd drifted apart.

The `event_id` branch prints `Color ID`. The ranged branch (`time_min`/`time_max`) doesn't.
Neither prints `recurringEventId`, `eventType` or `status`, even though all four are
already on the API response, since `events().list()` runs with no field mask.

I ran into this auditing event colours over a date range. The only way to do that today is
one `event_id` lookup per event, and because colour only appears under `detailed=True`,
every one of those also returns the event's full description. My calendar is mostly
Calendly and Zoom dial-in boilerplate, so reading one field per event costs a few hundred KB.

The same gap bites twice more. `recurringEventId` is what you need to update a series
instead of a single instance, and no read path returns it at all. And `events().list()`
includes `workingLocation` and `outOfOffice` entries by default, which without `eventType`
look like ordinary meetings.

Both branches now print all four. `colorId` always renders, defaulting to the string
`"None"`, which is what the `event_id` branch already did. The other three render only when
present and non-default, so an ordinary one-off confirmed meeting picks up a single extra
line. `detailed=False` output is untouched.

This changes what the formatter prints and nothing else. No extra API calls, no new
parameters, no signature changes.

## Type of Change

- [ ] Documentation update

## Testing

New file `tests/gcalendar/test_get_events_detailed_fields.py`, 15 cases. They cover each
new field, the compact-output rules, the unchanged `detailed=False` path, and a parity
check that runs the same event through both branches and asserts they agree on all four
fields, so a field added to one and not the other fails here.

`tests/gcalendar/` goes 91 to 106 passing. Reverting the source change fails 12 of the 15;
the 3 that still pass are the no-regression guards. Deleting the `Event Type` line from
just one branch fails the parity case and nothing else, which is what it's there for.

## Checklist

`ruff format --check` and `ruff check` are both clean on `gcalendar/calendar_tools.py` and
the new test file.

## Additional Notes

One judgement call worth flagging. I left the two formatting blocks separate rather than
factoring out a shared helper. They use different indentation and a different header line,
so collapsing them looked like more churn than this fix justifies, and it would have made
the diff much harder to review. The parity test is there to catch the drift that
duplication invites. Happy to do the refactor instead if you'd rather.

## Summary by CodeRabbit

* **New Features**
  * Detailed calendar event results now consistently include the event’s Color ID.
  * Recurring Event ID, non-default Event Type, and non-confirmed Status are shown when applicable.
* **Bug Fixes**
  * Missing Color IDs are displayed clearly as `None`.
  * Confirmed/default events omit Event Type, Status, and Recurring Event ID; other statuses are rendered.
* **Tests**
  * Added coverage to ensure single-event and date-range detailed outputs format event metadata consistently.


taylorwilsdon:
Thanks!

wdrwilson:
Amazing thanks!

PR #982, opened 2026-07-30
fix(gappsscript): default update_script_content to safe merge mode

### LINKED ISSUE #923: update_script_content silently DELETES files omitted from the files list — the tool description does not say so
**Summary**

update_script_content replaces the entire file set of an Apps Script project. Any file not included in the files argument is deleted. This is Google's documented behaviour for projects.updateContent, so it is not a bug in the API call itself.

The bug is in the tool description, which is what an LLM reads before deciding how to call it:

_"Updates or creates files in a script project."_

That phrasing strongly implies a partial update — add or modify the files you pass, leave the rest alone. It is the opposite of what happens. There is no mention anywhere in the docstring, the argument descriptions, or the tool title that omitted files are destroyed.

For an MCP server — where the caller is a language model reasoning from the description — a misleading description is not a documentation nit. It is the interface.

**Impact**

This footgun fired five separate times across a few days of agent-driven work on a 14-project Apps Script codebase, in three different sessions, by agents that had read the tool description and reasonably concluded they could push a single changed file:

A push of only appsscript wiped every other file in the project.
A push of 3 files (of 8) deleted shell, Code_progress, view_progress, view_overview, view_tcu.
A push omitting index deleted the entire UI of two live dashboards.

In each case the agent happened to still hold full copies from an earlier read and restored them. Had it not, the source would have been gone — Apps Script HEAD has no undo, and unversioned work is unrecoverable.

The destructiveHint=True annotation is set, which is good, but annotations are metadata; the docstring is the part the model actually reasons from, and it currently says the opposite of the truth.

**Suggested fixes**

Minimum (documentation). Rewrite the docstring so the semantics are unmissable:

python"""
REPLACES ALL FILES in a script project.

⚠️ DESTRUCTIVE: This is a full-project replacement, not a partial update.
Any file that exists in the project but is NOT included in `files` will be
PERMANENTLY DELETED. To change one file, you must first call
get_script_project() and pass back the complete file set with your edit applied.

Args:
    files: The COMPLETE list of files the project should contain after this
           call. Omitted files are deleted. Each item: {name, type, source}.
"""

Better (make the safe path the default). Add a merge: bool = True parameter:

merge=True (default): fetch the current project content, overlay the supplied files by name, and push the union. Callers get intuitive partial-update semantics and cannot destroy files by omission.
merge=False: current behaviour — full replacement — for callers who genuinely want to delete files.

This keeps full-replacement available while making the destructive path an explicit opt-in rather than the silent default. Given the caller is an LLM working from a one-line description, defaulting to safe seems strongly preferable.

Optional guardrail. When merge=False and the supplied file list is a strict subset of the existing files, return a warning naming the files that will be deleted, or require an explicit confirm_deletions=True.

**Environment**

taylorwilsdon/google_workspace_mcp
Tool: update_script_content
Underlying API: [projects.updateContent](https://developers.google.com/apps-script/api/reference/rest/v1/projects/updateContent) — replacement semantics are correct and expected; only the tool's description is wrong.

PS google_workplace_mcp Is awsone! Keep up the great work.

syf2211:
## Summary

`update_script_content` now defaults to `merge=true`, overlaying supplied files onto the current Apps Script project by file name instead of replacing the entire project. This prevents accidental deletion of omitted files when LLM callers perform partial updates.

## Motivation

Fixes #923. Google Apps Script `projects.updateContent` replaces the full project file set, but the tool description implied partial updates. Agents repeatedly deleted unrelated files when pushing only the files they changed.

## Changes

- Add `_merge_script_files` helper and `merge` parameter (default `true`) to `update_script_content`
- `merge=true`: fetch current content, overlay updates by file name, push merged set
- `merge=false`: preserve prior full-replacement semantics for intentional deletions
- Clarify tool docstring and update README / Apps Script reference docs
- Add regression tests for merge overlay, file creation, and full-replacement mode

## Tests

- `uv run pytest tests/gappsscript/test_apps_script_tools.py -q` → 26 passed
- `uv run ruff check gappsscript/apps_script_tools.py tests/gappsscript/test_apps_script_tools.py` → clean

## Notes

- **Behavior change:** callers who relied on deletion-by-omission must pass `merge=false` with the complete desired file set
- `merge=true` adds one `getContent` round-trip before `updateContent`
- File deletion is still available via `merge=false`; no silent delete path in merge mode

Fixes #923

## Summary by CodeRabbit

* **New Features**
  * Script content updates now merge with existing project files by default.
  * Added an option to replace the entire project, including deleting omitted files.
  * Improved file matching and validation for updates involving duplicate names or file types.

* **Bug Fixes**
  * Prevented concurrent updates to the same script from overwriting each other.

* **Documentation**
  * Updated tool descriptions, examples, and reference documentation to explain merge and replacement behavior.


taylorwilsdon:
Thanks!

PR #1000, opened 2026-08-04
fix: handle Drive resumable upload 308 responses

jdowning10:
## Summary

Drive resumable uploads use `308 Resume Incomplete` plus a `Range` header to acknowledge intermediate chunks. `httplib2` treats 308 as a redirect by default, so responses without a redirect `Location` header fail with `Redirected but the response is missing a Location: header` before the Google API client can continue the upload.

This removes only 308 from the redirect codes on the credentialed `httplib2.Http` instance. Other redirect handling remains unchanged.

## Tests

- `uv run --frozen pytest tests/auth/test_httplib2_timeout.py` — 4 passed
- `uv run --frozen ruff check auth/google_auth.py tests/auth/test_httplib2_timeout.py`
- `git diff --check`

The regression test verifies that 308 is removed while 300, 301, 302, 303, and 307 remain enabled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Google Drive resumable uploads by correctly handling `308 Resume Incomplete` responses.
  * Preserved other configured HTTP redirect behaviors while processing upload responses.


taylorwilsdon:
Thanks!

PR #1002, opened 2026-08-05
fix(gdrive): read Shared Drive permissions via permissions.list

drew-miraxis:
## Problem

`get_drive_file_permissions` and `check_drive_file_public_access` read the inline
`permissions` field returned by `files.get()`. For items in a **Shared Drive**, the
Drive API does not populate that field — nor the `shared` boolean — even when
`supportsAllDrives=True` is passed. As a result every Shared Drive file is reported
as private / "not shared with 'anyone with the link'", even when it is shared
"Anyone with the link". My Drive files are unaffected because `files.get()` does
populate their inline permissions.

## Reproduction

1. Put a file in a Shared Drive (not "My Drive") and share it "Anyone with the link → Viewer".
2. Call `get_drive_file_permissions(file_id=...)` (or `check_drive_file_public_access`).
3. Result: `Shared: False`, "No additional permissions (private file)", "❌ NOT shared with 'Anyone with the link'".

Verified directly against the Drive API on such a file:
`files.get(fields="…permissions…,shared", supportsAllDrives=True)` returns neither
`permissions` nor `shared`, while `permissions.list(fileId, supportsAllDrives=True)`
returns the full set including `{type: anyone, role: reader}`.

## Fix

When the file is in a Shared Drive (`driveId` present), resolve permissions via a
paginated `permissions.list(..., supportsAllDrives=True)` instead of the empty
inline field, and derive the "Shared" state from the resolved permission set rather
than the `shared` boolean. My Drive behaviour is unchanged (inline permissions
remain the source, no extra API call). Adds `list_all_permissions` and
`derive_shared_state` helpers in `drive_helpers.py`.

## Test

`tests/gdrive/test_drive_tools.py`: adds a Shared Drive case (fallback surfaces the
`anyone` grant) and a My Drive case (inline path, no extra `permissions.list` call).
The `tests/gdrive` suite passes (98 tests).

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Google Drive permission reporting for Shared Drive files.
  * Shared status and public-link access are now determined more accurately.
  * Permission details, including organizer access, are displayed consistently across Shared Drive and My Drive files.
  * Permission results now include complete access details, even when they are not included in the initial file information.
  * Permission reporting now handles large permission sets and paginated results more reliably.

### DIFF
--- a/gdrive/drive_helpers.py
+++ b/gdrive/drive_helpers.py
@@ -46,6 +46,55 @@ def check_public_link_permission(permissions: List[Dict[str, Any]]) -> bool:
     )
 
 
+def list_all_permissions(service, file_id: str) -> List[Dict[str, Any]]:
+    """
+    Return the complete permission set for a file, including Shared Drive items.
+
+    files.get() does NOT populate the inline `permissions` field (nor the `shared`
+    boolean) for items that live in a Shared Drive, even when supportsAllDrives=True
+    is passed. permissions.list() is the only reliable source in that case. This
+    helper paginates and always sets supportsAllDrives=True.
+    """
+    permissions: List[Dict[str, Any]] = []
+    page_token = None
+    while True:
+        resp = (
+            service.permissions()
+            .list(
+                fileId=file_id,
+                supportsAllDrives=True,
+                pageSize=100,
+                pageToken=page_token,
+                fields=(
+                    "nextPageToken, permissions(id, type, role, emailAddress, "
+                    "domain, expirationTime, permissionDetails, allowFileDiscovery)"
+                ),
+            )
+            .execute()
+        )
+        permissions.extend(resp.get("permissions", []))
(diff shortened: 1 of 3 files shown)


drew-miraxis:
Good catch @coderabbitai — confirmed. `check_drive_file_public_access` keyed the fallback on `file_metadata['driveId']`, but `driveId` wasn't in its `files.get()` field selection, so the Shared Drive fallback never ran there. Fixed in 1ae6a70: added `driveId` to the requested fields and switched the displayed Shared state to `derive_shared_state(...)` over the resolved permissions. Also added a Shared Drive test for this tool (`get_drive_file_permissions` already had one). Full `tests/gdrive` suite passes (99).

taylorwilsdon:
Thanks!

Issue #997, opened 2026-08-04
send_gmail_message retries a non-idempotent send, so a lost 200 becomes a duplicate delivery


siliroid:
### What I think is happening

`gmail/gmail_tools.py` passes `num_retries=GOOGLE_API_WRITE_RETRIES` (= **3**, `core/utils.py`) to `users().messages().send()` at two sites — `:2429` and `:2553`.

```python
sent_message = await asyncio.to_thread(
    service.users().messages().send(userId="me", body=send_body).execute,
    num_retries=GOOGLE_API_WRITE_RETRIES,
)
message_id = sent_message.get("id")
...
return f"Email sent! Message ID: {message_id}"
```

Gmail's send is not idempotent and accepts no idempotency key. And googleapiclient's retry predicate is unconditional on status — `googleapiclient/http.py`:

```python
def _should_retry_response(resp_status, content):
    # Retry on 5xx errors.
    if resp_status >= 500:
        return True
    # Retry on 429 errors.
    if resp_status == _TOO_MANY_REQUESTS:
        return True
```

So the retry fires precisely when the **response** failed — which is not the same event as the **send** failing. A message Gmail already accepted, whose 200 is lost to a 503 or a dropped connection, gets sent again. And again.

⇒ **Up to four deliveries, and the caller is told `Email sent!` exactly once.** The agent above reports one email to the user. The recipient got two.

### Why I think it is worth your time rather than a nitpick

I went and read how the retries got there before writing this. #742, merged by you, closing #731 — juliandickie's batched real-user report whose item ① was *"Auto-retry on transient HTTP 500/502/503/504 (Drive uploads)"*, off a production session where a Drive upload returned `internalError` and succeeded on retry.

That was the right fix for that problem. The release note on #742 describes the change as centralized retries across *"Drive uploads, PDF export, Gmail send/draft"* — three operations named in one breath, and only one of them is unsafe to repeat.

**The retry was added to buy reliability, and on this one call it is the thing that costs reliability.** Upload twice and you get one file. Send twice and somebody's client hears from you twice.

### The smaller one at the same two sites

`message_id = sent_message.get("id")` — if a response body ever arrives without an `id`, this returns the literal string `Email sent! Message ID: None` and nothing raises. Weak on its own, since Gmail returns an id on a 200 in practice, but it is the same shape one line away.

### Scope, honestly

[...]

PR #1010, opened 2026-08-07
fix(gmail): retry 429/5xx rate-limited sub-requests in batch fetches

### LINKED ISSUE #1009: get_gmail_messages_content_batch does not retry 429 rate-limited sub-requests
**Describe the bug**
`get_gmail_messages_content_batch` does not retry sub-requests that fail with HTTP 429 (rate limiting). The only retry helper (`_fetch_message_with_retry`) catches `ssl.SSLError` specifically; any other exception — including `HttpError` 429 — falls through to the generic `except Exception` branch and is returned immediately as an error stub. Because the batch tool emits `⚠️ Message <id>: <HttpError 429 ...>` *inside a successful 200 response*, the failure is invisible to any client-side retry-on-error logic: the caller sees exit 0 and never learns a subset of messages were not fetched.

Under normal load (e.g. two back-to-back 25-message batches), Gmail's per-user quota is exceeded and several sub-requests in the second batch fail with `rateLimitExceeded` / `userRateLimitExceeded`. The repo already has the detection helpers (`_is_quota_or_rate_limit_error`, `_http_error_status`, `GMAIL_QUOTA_ERROR_MARKERS`) but they are not used in the batch/read path.

**To Reproduce**
1. Call `get_gmail_messages_content_batch` with 25 message IDs (batch A) — succeeds.
2. Immediately call it again with 25 more IDs (batch B).
3. Batch B returns `Retrieved 25 messages` but 4+ entries are `⚠️ Message <id>: <HttpError 429 ... "Too many concurrent requests for user.">`.

**Expected behavior**
Sub-requests that fail with a retryable status (429 / 5xx / quota markers) should be retried with exponential backoff, mirroring the existing SSL-retry pattern and the 401 redo-loop behavior. Only the failed IDs should be re-fetched in a follow-up request, then merged with the successful results.

**Platform**
- OS: Linux
- Container: n/a
- Version: v1.23.1 (also reproduced the degraded capture through the opencode MCP client)

**Additional context**
The `googleapiclient` batch path has no `num_retries` support (`BatchHttpRequest.execute()` takes no retry parameter), so this must be handled in the server. The helpers `gmail/gmail_helpers.py:_is_quota_or_rate_limit_error` and `_http_error_status` already exist and are the natural building blocks. See also #997 (retry semantics for writes) — this issue is scoped to *read* batch fetching, which is idempotent and safe to retry.

lukaszliniewicz:
## Description

Fixes #1009 — `get_gmail_messages_content_batch` / `get_gmail_threads_content_batch`
silently lose messages when Gmail rate-limits individual sub-requests.

**The problem:** batch fetches only retried `ssl.SSLError`. HTTP 429 (rate limit) / 5xx
sub-request failures surfaced as `⚠️ Message <id>: <HttpError 429 ...>` stubs *inside a
successful batch response* — invisible to any client-side retry-on-error logic, so
partial fetches were quietly dropped under quota pressure (e.g. two back-to-back
25-message batches trip the per-user rate limit).

**The fix:**
- New `_is_retryable_error()`: retryable = SSL error, HTTP 429 / 5xx, or a
  quota/rate-limit marker in the body (reuses `_http_error_status` +
  `_is_quota_or_rate_limit_error` from `gmail_helpers`).
- `_fetch_message_with_retry` now retries retryable errors with 1s/2s/4s backoff
  (was SSL-only).
- After each batch chunk, collect sub-requests that failed with retryable errors,
  sleep a short backoff, then re-fetch **only those IDs** and merge — mirroring the
  existing 401 redo-loop pattern in `googleapiclient`.
- Reads are idempotent, so this retry is safe (deliberately NOT applied to sends;
  see #997).

## Type of Change
- [ ] Documentation update

## Testing
      (`tests/gmail/test_gmail_batch_retry.py`, 7 cases: 429/5xx retryable, 404 not,
      403 quota-marker retryable vs plain-forbidden not, non-HTTP not)
- [x] New and existing unit tests pass locally with my changes (`169 passed`)
- [x] ruff clean

## Checklist

## Additional Notes

Quota context (researched from official docs): `messages.get` = 20 quota units, and
batch sub-requests count individually (n per batch). A 25-message batch ≈ 500 units.
The per-user sustained budget is ~100 units/sec (new model) or 250/sec (legacy), so
back-to-back 25-batches exceed it quickly; the exponential backoff added here makes
occasional 429s degrade gracefully rather than dropping messages. The concurrent-
request limit ("Too many concurrent requests for user") is shared across all clients
accessing a mailbox and is the variant I observed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gmail message and thread retrieval reliability during temporary service errors, rate limits, quota limits, and SSL issues.
  * Automatically retries eligible failures and selectively re-fetches unsuccessful batch requests.
  * Preserved clear handling for non-retryable errors, such as missing resources and ordinary permission failures.

* **Tests**
  * Added coverage for retryable and non-retryable Gmail API error scenarios.

### DIFF
--- a/gmail/gmail_helpers.py
+++ b/gmail/gmail_helpers.py
@@ -2,12 +2,14 @@
 
 from __future__ import annotations
 
+import asyncio
 import html
 import logging
+import ssl
 from collections import Counter
 from datetime import datetime, timezone
 from email.utils import getaddresses, parseaddr, parsedate_to_datetime
-from typing import Any, Literal, Optional
+from typing import Any, Callable, Iterable, Literal, Mapping, Optional
 
 from fastmcp.exceptions import ToolError as ToolExecutionError
 from googleapiclient.errors import HttpError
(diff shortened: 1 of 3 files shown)


lukaszliniewicz:
Both the major issue and the nitpicks are already addressed in the current commit (`8408ea2`):

- **Major:** `fetch_thread_with_retry` was nested in the `except` block → `UnboundLocalError` on a 429 after a successful batch. Fixed (moved out) + regression test.
- **Nitpicks:** added the missing `ssl.SSLError` test, and switched to per-ID call counts (`msg-1`: 1, `msg-2`: 2; `missing`: 1).

The cited line numbers are from the previous revision — looks like a stale review, not new findings.

PR #1013, opened 2026-08-10
fix(gmail): resolve correct filename for attachments nested under multipart/signed

### LINKED ISSUE #1014: get_gmail_attachment_content saves every attachment as smime.p7s on S/MIME-signed messages
**Describe the bug**
`get_gmail_attachment_content` downloads the correct attachment bytes but
reports/saves the wrong filename whenever the source message is S/MIME
signed. Every attachment on such a message — regardless of which
`attachment_id` is passed — gets saved and reported as `smime.p7s`. Verified
by matching downloaded file size/content against the actual attachment; data
is not corrupted, only mislabeled.

Root cause: the metadata re-fetch in `get_gmail_attachment_content`
(`gmail/gmail_tools.py`) uses a `fields` mask that only requests one level of
`parts`:

```
fields="payload(parts(filename,mimeType,body(attachmentId,size)),body(attachmentId,size),filename,mimeType)"
```

Gmail's partial-response field masks are not recursive, so nested `parts`
beyond that level are silently dropped. On an S/MIME-signed message the MIME
tree is:

```
multipart/signed
├── multipart/mixed        (parts stripped by the fields mask — nparts=0 in the response)
│   └── application/pdf    (the real attachment — invisible after truncation)
└── application/pkcs7-signature  (smime.p7s — the only part still visible)
```

With only `smime.p7s` visible, the existing "exactly one attachment visible"
fallback (used because ephemeral attachment IDs don't always match) applies
its name to whatever was actually downloaded.

I confirmed this by dumping the raw API response with and without the
`fields` mask against a real S/MIME-signed message (bank statement email)
in my own inbox — the restricted response's `multipart/mixed` part came back
with `parts` completely absent, while the unrestricted response shows the
PDF nested inside it as expected.

**Startup Logs**
Not applicable — this is a per-call bug in `get_gmail_attachment_content`'s
metadata re-fetch, not a startup-time issue. Repro is at the tool-call level
(see below).

**To Reproduce**
1. Send/receive an S/MIME-signed email with a non-signature attachment (e.g.
   a bank statement with a PDF attachment, S/MIME signed) into a Gmail
   account.
2. Call `get_gmail_message_content` to list the message's attachments — note
   the actual attachment (e.g. `statement.pdf`) and its `attachment_id`,
   distinct from the `smime.p7s` signature part.
3. Call `get_gmail_attachment_content(message_id=..., attachment_id=<the
   PDF's id>)`.
4. Observe the response reports `Filename: smime.p7s` / `Saved filename:
   smime_<uuid>.p7s`, even though the downloaded byte size matches the PDF,
   not the signature part.

**Expected behavior**
The reported/saved filename should match the requested attachment
(`statement.pdf`), not the unrelated `smime.p7s` signature part.

 - OS: Ubuntu 24.04
 - Container: not applicable (ran via `uv run`, stdio transport)
 - Version: v1.23.1 (also reproduced against current `main`)

**Additional context**
Fix + regression test: #1013

The regression test mocks a `fields`-aware Gmail API client (a plain
`Mock(return_value=...)` ignores the `fields` kwarg entirely and can't
reproduce this class of bug, since it always returns the same canned
response regardless of what mask was requested).

bcelary:
## Description
`get_gmail_attachment_content` re-fetches message metadata to resolve the
attachment's filename/MIME type, using a `fields` mask that only requested
one level of `parts`:

```
fields="payload(parts(filename,mimeType,body(attachmentId,size)),body(attachmentId,size),filename,mimeType)"
```

Gmail's partial-response field masks are **not recursive** — any part nested
deeper than that level is silently dropped from the response. On S/MIME-signed
messages the real attachment sits two levels deep
(`multipart/signed` → `multipart/mixed` → the attachment), while the sibling
`smime.p7s` signature part is at the top level. With the mask in place, the
extractor only ever sees `smime.p7s`, and the existing "exactly one
attachment visible" fallback then labels *every* downloaded attachment as
`smime.p7s`, regardless of which attachment was actually requested. The
downloaded bytes are correct — only the reported/saved filename is wrong,
which is a silent misfiling risk for anything that trusts the filename.

**Fix shape**: I initially dropped the `fields` mask entirely, but traced it
back to #415, which added it deliberately (see the code comment it
introduced: *"Use format="full" with fields to limit response to attachment
metadata only"*) — an unrestricted `format="full"` fetch inlines the full
base64 `body.data` for every `text/plain`/`text/html` part, which is real
weight for image-heavy HTML mail and isn't needed just to resolve a
filename. Measured against real messages in my own inbox: full payload
9.9–53KB vs. 0.7–2.5KB with the original (buggy) mask.

So instead of removing the mask, this rebuilds it recursively to depth 6
(`_attachment_metadata_fields()`), which comfortably covers real-world MIME
nesting (S/MIME, forwarded messages, `multipart/related`, etc.) while
keeping the response small — 1.5–4KB in the same measurements, still
5–13x smaller than the full payload. Anything nested deeper than 6 levels
falls back to the existing size/last-resort matching, same as before.

Verified against a real S/MIME-signed message: the attachment id mismatch
reproduced pre-fix and resolved post-fix (filenames + downloaded byte sizes
match the actual attachments), and the recursive mask returns the same MIME
tree shape as an unrestricted fetch for that message.

## Type of Change

## Testing

Added `test_resolves_correct_filename_for_nested_smime_attachment`, with a
`fields`-aware mock that truncates to whatever depth the `fields` kwarg
actually requests (parsed from the mask string), rather than a hardcoded
depth — so it stays meaningful against the real depth-6 mask instead of only
testing a mock frozen to the old buggy shape. A plain `Mock(return_value=...)`
ignores the `fields` kwarg entirely and can't exercise this bug at all.
Confirmed the test fails against a synthetic depth-1 mask (reproducing the
original bug exactly) and passes at depth-6. Full `tests/gmail/` suite:
163/163 passing. `ruff check` / `ruff format --check` clean.

## Checklist

## Additional Notes
Fixes #1014 — filed alongside this PR with the full root-cause writeup and
repro steps.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed attachment downloads for deeply nested and S/MIME Gmail messages.
  - Attachments now retain their correct filenames, extensions, and contents instead of being incorrectly identified as `smime.p7s`.
  - Improved attachment metadata retrieval while avoiding unnecessary message body data, resulting in more efficient downloads.

- **Tests**
  - Added coverage for resolving nested PDF attachments in multipart and S/MIME messages.
  - Updated formatting assertions without changing expected behavior.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -2090,24 +2090,7 @@ async def get_gmail_messages_content_batch(
 
 
 def _attachment_metadata_fields(depth: int) -> str:
-    """Build a `fields` mask for messages.get() that mirrors a MIME part tree
-    down to `depth` levels of nesting, keeping only filename/mimeType/
-    attachmentId/size at every level (never `body.data`).
-
-    Gmail's partial-response field masks are not recursive: a bare
-    `parts(filename,mimeType,body(attachmentId,size))` only covers one level,
-    so any attachment nested deeper (e.g. multipart/signed S/MIME messages,
-    where the real attachment sits inside a nested multipart/mixed part and
-    only the top-level pkcs7-signature part is visible at depth 1) silently
-    disappears from the response — see the regression this guards against.
-    The mask still exists to keep the response small: an unrestricted
-    format="full" fetch inlines the full base64 body.data for every
-    text/plain and text/html part, which can be tens to hundreds of KB for
-    image-heavy HTML mail, none of which is needed just to resolve a
-    filename. depth=6 comfortably covers real-world MIME nesting (signed,
-    forwarded, multipart/related, etc.); anything deeper falls back to the
-    same size/last-resort matching used when the fields mask doesn't help.
-    """
+    """Build a metadata-only fields mask for a MIME tree of the given depth."""
     node = "filename,mimeType,body(attachmentId,size)"
     for _ in range(depth):
         node = f"filename,mimeType,body(attachmentId,size),parts({node})"
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -1918,6 +1918,34 @@ async def get_gmail_messages_content_batch(
     return final_output
 
 
+def _attachment_metadata_fields(depth: int) -> str:
+    """Build a `fields` mask for messages.get() that mirrors a MIME part tree
+    down to `depth` levels of nesting, keeping only filename/mimeType/
+    attachmentId/size at every level (never `body.data`).
+
+    Gmail's partial-response field masks are not recursive: a bare
+    `parts(filename,mimeType,body(attachmentId,size))` only covers one level,
+    so any attachment nested deeper (e.g. multipart/signed S/MIME messages,
+    where the real attachment sits inside a nested multipart/mixed part and
+    only the top-level pkcs7-signature part is visible at depth 1) silently
+    disappears from the response — see the regression this guards against.
+    The mask still exists to keep the response small: an unrestricted
+    format="full" fetch inlines the full base64 body.data for every
+    text/plain and text/html part, which can be tens to hundreds of KB for
+    image-heavy HTML mail, none of which is needed just to resolve a
+    filename. depth=6 comfortably covers real-world MIME nesting (signed,
+    forwarded, multipart/related, etc.); anything deeper falls back to the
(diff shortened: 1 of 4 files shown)


taylorwilsdon:
Thanks!

PR #1035, opened 2026-08-18
feat(gmail): complete draft lifecycle — update, delete, and send drafts, zero new tools

123andy:
## Description

Completes the Gmail **draft lifecycle** by surfacing three Gmail API methods that currently have no route through the server — `drafts.update`, `drafts.delete`, and `drafts.send` — with **zero new tools**. Supersedes #943, reworked after the tool-count concern raised there: instead of adding a `send_gmail_draft` tool, every capability lands as a parameter on an existing tool.

- **`draft_gmail_message`** gains `action="create" | "update" | "delete"` (+ `draft_id`):
  - `update` → `drafts.update` (replaces the draft wholesale, preserving the draft ID and thread — matching the API's semantics).
  - `delete` → `drafts.delete`, with the docstring warning that deletion is **permanent** (the API bypasses Trash), and a name-aware rejection when content args are passed to `delete` ("Did you mean `action='update'`?").
- **`send_gmail_message`** gains `draft_id`: sends an existing draft **as-is** via `drafts.send`. Combining `draft_id` with any content/addressing argument is rejected with guidance to `draft_gmail_message(action='update')` first — so there is never ambiguity about what got sent.
- Guidance fix: `modify_gmail_message_labels` now warns when TRASH is added to a draft's message that **trashing the message does not remove the draft** — a real user-confusion case (duplicate-draft reports).
- Forward-message internals refactored (`_build_forward_message`) so drafts and sends share one forward-construction path.

Compose-then-send matters for agentic use: the model drafts, a human reviews in Gmail, then a single `send_gmail_message(draft_id=…)` commits it — and because tool gating (tier configs / allowlists) is **per tool name**, "can draft" and "can send" remain independently grantable.

### Relationship to #812

#812 reaches draft **deletion** by consolidating `draft_gmail_message` into a new `manage_gmail_draft` name (net −1 tool). This PR keeps the existing, widely-deployed tool names — renames break per-tool-name allowlists and tier configs already pointing at `draft_gmail_message`/`send_gmail_message` — and reaches **net 0** by folding the verbs into the existing tools; it also covers `update` and draft **send**, which #812 does not. If the `manage_*` naming convention is preferred, this content rebases naturally onto that shape — happy to defer to the maintainer on naming and coordinate with #812 rather than duplicate it.

## Type of Change

## Testing

21 new tests (`test_draft_lifecycle.py`, `test_send_draft_via_send_message.py`): exact API-call assertions for update/delete/send, validation ordering, 404/400 → actionable errors, threading preservation, and the trash-draft notice. Manually validated end-to-end against a deployed streamable-http instance (create → update → send-by-id, and the delete path).

## Checklist

## Additional Notes
Opening as a draft while our internal review finishes; will mark ready shortly. Once this lands I'll close #943 in its favor.

## Summary by CodeRabbit

* **New Features**
  * Send existing Gmail drafts by draft ID.
  * Create, update, forward, and permanently delete drafts.
  * Forward messages with signatures and attachments.
  * Combine forwarded and manually added attachments.

* **Bug Fixes**
  * Added clear handling for missing, stale, or invalid draft IDs.
  * Prevented incomplete forwarded attachments from creating drafts.
  * Clarified that trashing a draft’s message does not delete the draft.


123andy:
Both CodeRabbit findings addressed in 9df8b91:

- **compose scope at the tool boundary** — agreed and fixed as suggested. The decorator is back to `["gmail_read", GMAIL_SEND_SCOPE]`, so ordinary/reply/forward sends require nothing new from existing credentials. The `draft_id` path now translates Google's 403 on `drafts.send` into re-auth guidance instead: anyone drafting through this server already holds `gmail.compose` (`draft_gmail_message` declares it, so it's part of the same consent), which leaves only pre-drafts credentials able to hit that 403.
- **send-later hint nested under `threadId`** — fixed; the `send_gmail_message(draft_id=…)` instruction is now emitted unconditionally with the draft ID.

Tests added for both (403 → `UserInputError` with `gmail.compose` guidance; full suite green).

BosTheCoder:
@123andy Do you still plan on working on this?👀Would be super useful!

123andy:
@BosTheCoder yes — thanks for the nudge. Rebuilt on current `main` (v1.27.1) in `0fb08a2`, and the two outstanding CodeRabbit nitpicks are fixed (`UserInputError` for the failed-attachment path; the reused `draft_body` split into `draft_request_body`).

### Merge notes worth a maintainer's eye

- **`draft_gmail_message` needed manual reassembly** — both sides had rewritten the same block. Every `main` behaviour is preserved: default Send-As (#1054), bare-newline HTML (#1132), reply targeting (#1065), scrubbed logging (#1037). The action dispatch sits in front of them.

- **Fixed an argument-count bug** in the `_derive_reply_headers` call: the 4th parameter (`target`) was omitted, which silently reverted #1065's fix to the last message in the thread.

- **Restored a dropped import.** `HttpError`'s only `gmail_tools` use moved to `gmail_helpers` on `main`, so the merge removed the import while three `except HttpError` clauses still depended on it. Restored, and the three hand-rolled status lookups now use `_http_error_status()`.

### One question, so this doesn't sit on your queue

CodeRabbit's pre-merge check flagged scope: #943 had scoped update/delete to #920, and this PR consolidates send + update + delete into one change. **Would you rather review this as one PR, or split into (1) send-an-existing-draft via `draft_id` and (2) the update/delete lifecycle?** Happy either way.

### Unrelated, but possibly useful

`uv.lock` on `main` is stale: it pins `workspace-mcp 1.27.0` while `pyproject.toml` says `1.27.1`, so `uv run` re-locks on any contributor's machine. The incidental bump is reverted out of this PR to keep the diff to Gmail — it probably wants its own regeneration.

### Why this is still worth landing

`main`'s `draft_gmail_message` still has no `action` or `draft_id` parameter and ends at `"Draft created!"` — update, delete and send-by-id don't exist yet. And v1.26.0 added a docstring telling callers to reach past the tool and `call users.drafts.send with the draft ID`, which is exactly the gap this closes. Still zero new tools.

123andy:
Both CodeRabbit findings are fixed: an updated draft now stays in its reply thread, and a `drafts.send` 403 is classified before re-authentication is suggested.

The PR has grown since — preserve-on-update, `clear_fields`, `action="list"`, and live testing against Gmail. The description is rewritten to cover it, including the related PRs and the schema cost.

123andy:
Both fixed in `e9115e1`. A 400 from `drafts.send` no longer claims the draft is missing — it carries Google's reason and gives the remedy for an unsendable draft as well as for a bad ID. The size-cap refusal on the header-patch path is now a `UserInputError`, matching the forward-attachment path.

PR #1036, opened 2026-08-18
fix(drive,gmail): stop advertising client-side file_path on remote transports; honest boundary errors

123andy:
## Description

On a remote (`streamable-http`) deployment, `file_path` resolves on the machine the **server** runs on — a container with no view of the caller's disk — so a client-supplied path can never work. Today the parameter is still advertised, and a doomed call falls through to a bare "Path does not exist", which reads as a typo rather than a topology mismatch. This PR makes remote deployments **stop advertising parameters that cannot work**, and makes the failure honest for clients that cached the old schema:

1. **Schema-level hiding**: the three `import_to_google_*` tools and `update_drive_file` set `exclude_args=stdio_only_args("file_path")` — on `streamable-http` the parameter is simply absent from the tool schema; on stdio nothing changes. This is the same deployment-adaptive-schema idiom v1.24 established for hiding `user_google_email` under managed identity.
2. **Runtime guard as defense-in-depth** (for clients with a cached schema): `_import_with_conversion` rejects `file_path` in remote mode *before* touching the Drive API, with an error that names the working alternatives for that specific tool (`content` where supported, `file_url`).
3. **Honest boundary errors**: `validate_file_path` and the Gmail attachment-failure guidance are transport-aware — remote callers are told the path "does not exist **on the MCP server**" and pointed at inline `content`; server-operator advice (`ALLOWED_FILE_DIRS`, move-the-file) is only shown on stdio, where the operator is the caller.
4. **Startup-order fix**: `main.py` now calls `set_transport_mode()` immediately after transport resolution, *before* tool imports — required for any decoration-time schema adaptation to see the real transport (`fastmcp_server.py` already did this).
5. **Decoration-time CI coverage**: a subprocess test imports the tool module with the transport pre-set to `streamable-http`, because FastMCP validates `exclude_args` at decoration time and only when non-None — an in-process suite running under the stdio default can never catch a stale exclusion.

### Relationship to #871 / #872 / #873

#872 (and #871 for create/update) solve the *capability* side: routing large/binary uploads around the server via resumable upload URLs, plus a runtime rejection of `file_path`/`file_url` in remote mode. This PR is the *advertising/honesty* layer and is complementary: schema hiding composes cleanly with a `return_upload_url` route (the hidden param stays hidden; the new route is what remote clients see). Where the two overlap — the runtime rejection inside the import tools — I'm happy to rebase on #872 and drop the duplicated guard if it lands first. #873's local-path guard for Gmail attachments is likewise adjacent to the transport-aware attachment guidance here.

## Type of Change

## Testing

New suite `tests/gdrive/test_import_remote_guard.py`: all three import tools reject `file_path` remotely before any API call with alternative-naming errors; inline `content` still works remotely; `file_path` still works end-to-end on stdio; attachment advice is transport-aware; the `stdio_only_args` helper; and the subprocess decoration-time import test. Schema output verified in both transports against the real FastMCP schema-generation path (param present on stdio, absent on streamable-http).

## Checklist

## Additional Notes
FastMCP ≥ 2.14 marks `exclude_args` deprecated in favor of `Depends()` dependency injection; it still works, and the existing `user_google_email` hiding rides the same mechanism — migrating both to DI together would be a natural follow-up, out of scope here. Opening as a draft while our internal review finishes; will mark ready shortly.

## Summary by CodeRabbit

* **New Features**
  * Improved remote imports and attachments by guiding callers toward inline content or reachable URLs instead of server-local file paths.
  * Remote tool schemas now hide parameters intended only for local standard-input/output usage.
* **Bug Fixes**
  * Enhanced file validation and permission errors with transport-specific guidance.
  * Prevented unsupported remote file-based imports before external service requests are made.
* **Tests**
  * Added coverage for remote file handling, inline content workflows, schema behavior, and transport-specific errors.


123andy:
CodeRabbit's finding was correct and is fixed in f74ac26: `update_drive_file` had the schema-level hiding but reached `_resolve_import_media` outside `_import_with_conversion`, so a cached-schema client sending `file_path` would have had the server read its own filesystem. It now gets the same remote-mode `UserInputError` guard (before any Drive call), pointing at `content`/`file_url`, with a regression test asserting rejection happens before any API or file access.

axsaucedo:
Independent second-site reproduction of the problem this PR fixes.

We run this same server image behind a brokered remote write path (an internal enterprise agent platform proxies the `streamable-http` connection; the deploying org, not the caller, controls the container filesystem). The `import_to_google_doc` description's advice to "prefer `file_path` for files on disk" is exactly backwards for that topology: a caller-supplied path can never resolve, and the resulting `Path does not exist` on a file that plainly exists locally reads as a typo, not a topology mismatch. It cost a first-time user several retries before landing on `content` as the working alternative.

We currently carry a client-side override of this tool's description (rewriting it to say `file_path` is server-side only and to use `content` for local files) specifically to work around this. This PR — hiding `file_path` from the schema on remote transports plus the runtime guard and transport-aware error — would let us delete that override entirely, since it fixes the problem at the source rather than downstream of it.

Wanted to add this as a second data point that the failure mode here is real and reproduces outside the original reporter's setup, in case it's useful for moving the review forward.

123andy:
Updated to current `main`; conflicts resolved, suite green.

- **Conflicts** were all in `gdrive/drive_tools.py`. Upstream's `base64_content` / `base64_sha256` and this PR's guard are both kept. The import docstrings keep the newer goal of keeping file contents out of the caller's context, stated per transport: remote servers prefer `file_url`, local servers prefer `file_path`.
- **`base64_content`** reached the import tools after this branch was cut, so the remote `file_path` rejection was stale: `import_to_google_slides` told callers to host a file that can be sent inline. The rejection now names only the routes its tool has, checked against the tool's signature. On `update_drive_file` it runs before the mode check and offers `content` for text formats only.
- **The runtime guard is reachable.** A new test calls through FastMCP's client: `exclude_args` hides `file_path` from the schema, but the parameter stays in the function signature, so an argument sent anyway still reaches the function.

CodeRabbit's two new findings are in `core/utils.py` code this PR does not touch; they came in with the merge of `main`.

taylorwilsdon:
Thanks for this. The case you're describing is valid, but I actually suspect for the overwhelming number of deployments (albeit with far fewer users total) using streamable-http doesn't mean the server is on another machine. Our own README uses claude mcp add --transport http workspace-mcp http://localhost:8000/mcp as the standard setup. In that case the server and the client share a filesystem, and file_path works today. This PR would remove file_path from the schema for those users and reject it at runtime. 

The only routes left are file_url or inline content, which pushes whole files through the model's context. The new error messages would also tell localhost users something false: that the server can't see their filesystem. And they'd hide the ALLOWED_FILE_DIRS advice, which is the actual fix for them. I'm going to refactor this to make it something we can merge without breaking a bunch of workflows.

123andy:
Thanks. An explicit setting is the right call; transport was the wrong signal for "the server can see the caller's disk". It fits the hosted deployment that motivated this, which runs behind a gateway with OAuth 2.1 off, so the stateless-mode shortcut does not apply there and the explicit `WORKSPACE_MCP_DISABLE_LOCAL_FILES=true` is what it will set. Checked locally on this branch: with the setting on, `file_path` drops out of the schema on streamable-http; unset, nothing changes.

On the routes that remain once local files are off: on a hosted server, the way to avoid pushing whole files through the model's context is to keep the bytes off the server entirely. Resumable upload URLs do that for uploads (#871, #872), and signed download URLs do it for the other direction (#888). Those can be brought onto current `main` if that helps review.

123andy:
CodeRabbit flagged this on #1160, which builds on it: `local_file_access_enabled()` compares `WORKSPACE_MCP_DISABLE_LOCAL_FILES` without stripping whitespace, so `" true "` leaves local files enabled — failing open. `.strip().lower()` would close it; the stacked PR picks it up on rebase.

123andy:
**FastMCP 4 removed `exclude_args`, so this branch fails at startup on a fresh install.** `pyproject.toml` allows `fastmcp>=3.4.7`; a fresh install now resolves 4.0.x and fails with:

```
TypeError: FastMCP.tool() got an unexpected keyword argument 'exclude_args'
```

v1.28.0 without this PR runs fine on 4.0.5, so only the four `exclude_args` uses added here are affected.

Two commits are ready on this branch, each independently droppable:

- `878b5d3` hides `file_path` by rewriting the signature, the same way `require_google_service` already hides `user_google_email`. Behaviour is unchanged, except that a stale client sending a hidden `file_path` is now rejected by FastMCP's validation before the tool runs; the in-tool guards stay as a backstop. Checked on FastMCP 3.4.7 and 4.0.5.
- `f0747b5` strips whitespace from `WORKSPACE_MCP_DISABLE_LOCAL_FILES`, so `" true "` no longer leaves local files enabled (flagged by CodeRabbit on #1160).

123andy:
A live run with `WORKSPACE_MCP_DISABLE_LOCAL_FILES=true` showed `file_path` correctly absent from the schemas while the shipped prose still named it (`update_drive_file`'s description and its `source_format` line, and the "prefer file_path" sentence on the three `import_to_google_*` tools), inviting a model to call a parameter that does not exist. The wording is now static and honest in both configurations (local-only paths are described as offered on a local server, without naming the parameter), pinned by a subprocess test that checks the shipped text under both settings.

taylorwilsdon:
Think we've got it there - thanks!

PR #1037, opened 2026-08-18
fix: keep user free-text out of INFO/ERROR logs (queries, find/replace text, subjects, titles)

123andy:
## Description

Several tools log free text the user typed at INFO — and once at ERROR — where it lands in durable logs rather than the conversation that owns it:

- **Search queries** logged verbatim by all seven search paths: `search_gmail_messages`, `search_drive_files` (in up to three lines per call), `search_docs`, `search_contacts` (twice), `get_events` (twice), Chat `search_messages`, and `search_custom`. Gmail/Drive/Contacts queries routinely carry names, addresses, and sensitive terms.
- **Free text**: `find_and_replace_doc` logged the full find and replace strings.
- **Subjects/titles/names**: Gmail send/draft subjects, event/calendar summaries, doc/form/sheet/slides/task/Apps-Script titles, sheet tab names, contact-group names.
- **Addressing beyond the caller**: a forward's recipient list; freebusy calendar IDs (typically other people's addresses).
- **URLs**: `create_drive_file(fileUrl=…)` logged the URL verbatim — query strings can carry signed-URL secrets (the Gmail attachment path already redacts these via `_redact_url`).
- **The ERROR-path re-leak**: `handle_http_errors` logged the raw `HttpError` with `exc_info` on every failed Google call, and `HttpError.__str__` embeds the full request URI — `…/messages?q=<the user's search terms>` — so every query scrubbed at INFO leaked right back on any 4xx.

**What this PR does:**

1. INFO keeps operational metadata only — who, which resource, how much (`query_len=`, `subject_len=`, or the ID where one already names the line). Values leave INFO at ~45 sites across 11 modules.
2. Where the text has real debugging value (find/replace, search queries), a `logger.debug` companion keeps it available — and a new **`WORKSPACE_MCP_LOG_LEVEL`** env var makes DEBUG actually reachable (`basicConfig` previously hard-coded INFO with no override). Default unchanged.
3. `handle_http_errors` now logs the error with URL query strings redacted (`?<query-redacted>`, endpoint path kept) at ERROR, with the full exception + traceback at DEBUG. The exception raised to the *caller* still carries the full text — it's their own request.
4. `tests/test_log_hygiene.py` pins the principle with caplog tests (positive anchors so they can't pass vacuously, DEBUG-side assertions so "moved, not deleted" is verified) plus a unit test for the URI scrubber.

**Why this matters even single-user:** the repo already treats log leakage as a bug — `main.py` suppresses httpx INFO logs *because they leak access tokens in URLs*, and review guidance lists "no secrets or PII logged" as a top priority. `configure_file_logging` persists INFO+ to `~/.google_workspace_mcp/logs/` by default, so these values get a durable on-disk record beyond the conversation even on a laptop. And multi-user `streamable-http` is a first-class deployment mode (the helm chart ships in-repo), where server logs commonly flow to aggregators with a different audience than the data's owner.

**Deliberately kept:** the authenticated user's own email (the principal — operationally load-bearing) and attachment/file *names* in transfer paths (metadata-tier; happy to tighten if preferred).

Merge-order note: independent of #1035/#1036 apart from a trivial rebase on one `draft_gmail_message` log line, whichever lands second.

## Type of Change

## Testing

Full suite: 1695 passed. `ruff check` and `ruff format --check` clean.

## Checklist

## Additional Notes
The commit history shows the sweep hardening in stages (an adversarial review pass caught a `None`-query crash, missed sites, and the unreachable-DEBUG/ERROR-path gaps before this was opened); happy to squash on merge.

## Summary by CodeRabbit

* **Security & Privacy**
  * Sensitive search queries, titles, filenames, message details, URLs, attendee information, and contact data are excluded from standard logs.
  * Error logs redact URL query parameters while preserving diagnostic context.
  * Detailed values remain available at debug level when needed.

* **New Features**
  * Gmail label updates can optionally verify applied changes and report unresolved or unchanged messages.

* **Configuration**
  * Log verbosity uses `WORKSPACE_MCP_LOG_LEVEL`, safely defaulting to `INFO` for missing or invalid values.

* **Tests**
  * Added coverage validating redaction and log-level behavior across searches, replacements, attachments, URLs, and HTTP errors.


123andy:
All four CodeRabbit findings addressed in 0a10a49:

- **`create_sheet` `len(sheet_name)` on an Optional param** — real crash, fixed with a guard. Also re-audited every remaining `len()` site this PR introduced: all others take required `str` parameters.
- **`time_filter` at INFO** — agreed; it's a caller-provided string. INFO now logs `has_time_filter`, the value moved to a DEBUG companion.
- **`create_drive_file` logged URL** — now built via `urlsplit` with both query *and* fragment stripped (either can carry signed-URL secrets).
- **`WORKSPACE_MCP_LOG_LEVEL` validation** — good catch: `getattr(logging, <env value>)` can resolve to a non-level attribute (e.g. `BASIC_FORMAT`) and crash `basicConfig` at startup. The level name is now checked against an explicit allowlist with an INFO fallback.

Full suite 1699 passed; ruff clean.

taylorwilsdon:
Thanks!

PR #1042, opened 2026-08-20
fix(gmail): include Reply-To in message and thread header output

### LINKED ISSUE #1043: get_gmail_message_content and get_gmail_thread_content drop the Reply-To header
## Summary

`GMAIL_METADATA_HEADERS` (`gmail/gmail_helpers.py`) whitelists `Subject`, `From`, `To`, `Cc`, `Message-ID`, `In-Reply-To`, `References`, `Date`, `List-Unsubscribe`, `Precedence`, `List-Id` — but not `Reply-To`. Both `get_gmail_message_content` and `get_gmail_thread_content` pass this list to the Gmail API's `metadataHeaders` param, so `Reply-To` never comes back in either `"metadata"` or `"full"` format, for a single message or a thread.

Reply-To is already fetched and used internally for reply-target resolution (the reply-header derivation helpers used by `send_gmail_message`/`draft_gmail_message`) — it's just never surfaced to a caller reading a message.

## Impact

An agent reading a message or thread has no way to see that replies should route to a different address than the sender — the same class of problem #963 raised for `To`/`Cc` (fixed in #970), just for `Reply-To` instead.

## Fix

#1042 adds `Reply-To` to `GMAIL_METADATA_HEADERS` and surfaces it in both the message and thread formatters, following the same explicit-presence convention #970 established for `To`/`Cc`.

mickey-mikey:
## Description

`GMAIL_METADATA_HEADERS` (`gmail/gmail_helpers.py`) never included `Reply-To`, so it was excluded from the Gmail API's `metadataHeaders` filter before reaching any formatting code. `get_gmail_message_content`, `get_gmail_messages_content_batch`, `get_gmail_thread_content` and `get_gmail_threads_content_batch` all silently dropped it as a result.

Reply-To was already being fetched and used internally for reply-target resolution (the reply-header derivation helpers), it just was never surfaced to the caller reading a message.

Follows the explicit-presence convention already established for To/Cc by #970 — no output change when Reply-To is absent from a given message.

## Type of Change

## Testing
- [x] New and existing unit tests pass locally with my changes (`pytest tests/gmail/` — 213 passed)

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Gmail message and thread details now display the `Reply-To` header when available.
  * Gmail metadata collection now includes the `Reply-To` header for more complete email details.

* **Bug Fixes**
  * Improved handling of messages without a `Reply-To` header, ensuring empty values are omitted.

* **Tests**
  * Added coverage to verify `Reply-To` headers are included, formatted correctly, and excluded when absent.

Closes #1043

### DIFF
--- a/gmail/gmail_helpers.py
+++ b/gmail/gmail_helpers.py
@@ -34,6 +34,7 @@ GMAIL_METADATA_HEADERS = [
     "From",
     "To",
     "Cc",
+    "Reply-To",
     "Message-ID",
     "In-Reply-To",
     "References",
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -296,6 +296,7 @@ def _format_message_header_lines(
     """Format standard Gmail message headers for response output."""
     subject = headers.get("Subject", "(no subject)")
     sender = headers.get("From", "(unknown sender)")
+    reply_to = headers.get("Reply-To", "")
     to = headers.get("To", "")
     cc = headers.get("Cc", "")
     rfc822_msg_id = headers.get("Message-ID", "")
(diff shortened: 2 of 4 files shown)


taylorwilsdon:
Thanks!

PR #1044, opened 2026-08-20
Report and author event boundaries in their own timezones

seidnerj:
Google returns each event boundary as `{"dateTime", "timeZone"}`, and the two do not agree by default: `events.list`/`get` normalize every `dateTime` in a response into one zone (the `timeZone` request param, defaulting to the calendar's own) while leaving each boundary's IANA `timeZone` untouched.

That makes cross-timezone events lossy in both directions:

**Read.** A flight landing 17:50 `Europe/Amsterdam` comes back as `2026-08-21T18:50:00+03:00` on an `Asia/Jerusalem` calendar. Correct instant, but the formatter echoed only the normalized offset and dropped `timeZone`, so the arrival's real local time was unrecoverable from the output.

**Write.** `create`/`update` stripped the caller's explicit offsets (correct on its own, so the IANA name resolves the DST-aware offset) then stamped one `timezone` onto *both* boundaries. Authoring that same flight wrote the arrival as 17:50 Israel time: an hour off, no error, API reports success.

## Changes

- `EventBoundary` (frozen dataclass) + `parse_event_boundary()` resolve each boundary into a tz-aware moment in its own zone. `_format_event_time()` is now a thin wrapper, so all existing call sites benefit. The instant never changes, only its presentation.
- `start_timezone` / `end_timezone` on `manage_event` and both impls, each falling back to `timezone`. Both paths build boundaries through one shared `_build_time_boundary()`.

## Compatibility

Conversion is gated on a resolvable IANA `timeZone`. Without one the returned value is echoed verbatim, preserving `Z` rather than `+00:00`, as `test_get_events_detailed_fields.py` already pins. Omitting the new params reproduces existing behavior exactly.

## Tests

29 new; full suite 1722 passed / 2 skipped; ruff clean. Coverage includes the case most likely to break: Israel and the EU switch DST on different dates, so for two days each spring the TLV/AMS gap is 2 hours rather than 1.

## Summary by CodeRabbit

* **New Features**
  * Added support for separate timezones on event start and end boundaries when creating or updating events.
  * Improved handling of all-day events, daylight-saving transitions, timezone conversions, and weekday display.
  * Preserved explicit time offsets and supplied timezone names when conversion is unavailable.

* **Bug Fixes**
  * Added safe fallbacks for malformed or unrecognized event data.
  * Validated timezone names before creating or updating timed events.
  * Prevented missing update boundaries from being added unintentionally.

### DIFF
--- a/gcalendar/calendar_helpers.py
+++ b/gcalendar/calendar_helpers.py
@@ -54,13 +54,16 @@ class EventBoundary:
     """The dateTime/date string exactly as Google returned it."""
 
     local_date: datetime.date
-    """Calendar date at this boundary, in :attr:`timezone` when one is known."""
+    """Calendar date at this boundary, in its own zone when that zone was resolved."""
 
     moment: Optional[datetime.datetime] = None
-    """Timezone-aware instant, converted into :attr:`timezone`. None for all-day."""
+    """Timezone-aware instant, converted when :attr:`timezone_resolved` is true."""
 
     timezone: Optional[str] = None
-    """IANA zone name from the boundary, when Google supplied a resolvable one."""
+    """IANA zone name exactly as Google supplied it, even if locally unresolvable."""
+
+    timezone_resolved: bool = False
+    """Whether :attr:`moment` was converted into :attr:`timezone`."""
 
     is_all_day: bool = False
     is_exclusive_end: bool = False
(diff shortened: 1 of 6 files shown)


seidnerj:
Thanks - the positional-contract finding was correct and is fixed in `276b712`: both parameters moved to the end of `manage_event`, `_create_event_impl` and `_modify_event_impl`, behind a bare `*` so they are keyword-only. That restores every prior binding and prevents the same breakage when parameters are added later. Two regression tests cover it, both confirmed to fail against the previous ordering.

On the two follow-ups, I've taken one and am declining the other.

**Validate `timezone` before stripping the offset** - taken, with a different remedy, in `f42f6d7`. An unresolvable name is now rejected up front with a `ValueError` naming the offending zone. I deliberately did not fall back to `{"dateTime": time_value}`: silently dropping the zone the caller asked for books the event in a different one with no error, which is worse than the opaque rejection it replaces. All-day boundaries carry no zone and skip validation.

**Restore RFC3339 normalization when no timezone is supplied** - declining. There is no prior normalization on this path to restore. `_correct_time_format_for_api` is applied only to query time-ranges (`get_events`, the OOO and focus-time list tools, `query_freebusy`), never to event body boundaries. Before this PR the create path set `effective_start = start_time` unchanged when no timezone was given, producing exactly the `{"dateTime": time_value}` that `_build_time_boundary` produces now - the refactor is behavior-preserving here. Adding normalization would be a behavior change for existing callers and belongs in its own PR.

taylorwilsdon:
Thanks!

Issue #597, opened 2026-03-20
Gmail tools hang indefinitely on VPS without IPv6 due to httplib2 AAAA resolution


alvarosanchezuy:
## Summary
All Gmail tools hang indefinitely without error on VPS instances where IPv6 is disabled or non-functional.

## Environment
- workspace-mcp version: 1.14.3
- Python version: 3.12
- OS: Debian (VPS, IPv6 disabled via sysctl)
- Used via: OpenClaw google-workspace plugin (execFile + uvx + --cli mode)

## Symptom
`gmail_search`, `gmail_read`, `gmail_send` and all other Gmail tools hang indefinitely. No error is thrown, no timeout is triggered. The process simply never returns.

Debug logging shows execution reaches `service.users().messages().list(...).execute()` and blocks there forever:
```
INFO:auth.google_auth:[search_gmail_messages] Successfully authenticated gmail service for user: nova@neuronal.uy
INFO:gmail.gmail_tools:[search_gmail_messages] Email: 'nova@neuronal.uy', Query: 'is:unread', Page size: 10
<hangs here indefinitely>
```

## Root Cause
The Google API Python client uses `httplib2` for HTTP requests. `httplib2` resolves DNS via `getaddrinfo()` and prefers IPv6 (AAAA records) by default. On VPS instances where IPv6 is configured at the DNS level but non-functional at the network level, `httplib2` attempts the IPv6 connection and blocks indefinitely — there is no connection timeout in httplib2's default configuration.

This is a **silent failure**: no exception, no log, no timeout.

## Verification
```bash
# IPv6 non-functional:
curl -6 https://gmail.googleapis.com/  # exit code 28 (timeout)

# IPv4 works fine:
curl -4 https://gmail.googleapis.com/  # responds immediately

# Python getaddrinfo returns IPv6 first (default):
python3 -c "import socket; print([r[0].name for r in socket.getaddrinfo('gmail.googleapis.com', 443, socket.AF_UNSPEC, socket.SOCK_STREAM)])"
# ['AF_INET6', 'AF_INET6', 'AF_INET6', 'AF_INET', 'AF_INET', 'AF_INET']

# Direct curl to Gmail API with same token works instantly:
curl -s -m 10 -H "Authorization: Bearer $TOKEN" \
  "https://gmail.googleapis.com/gmail/v1/users/me/messages?q=is:unread&maxResults=3"
# Returns results in <1s
```

## Workaround Applied
[...]

PR #1046, opened 2026-08-20
fix: run external token validation off the event loop

pavel-at-stacks:
## Description
In external OAuth mode, `verify_token` calls `get_user_info()` directly. This is a synchronous HTTPS call to Google's userinfo API, and `verify_token` runs on the event loop. The server validates a token on every request, so every request blocks the loop for one Google round-trip. One slow call freezes the whole server: it cannot answer `initialize` or `list_tools`, and it cannot deliver tool results until the call returns (up to the 30s socket timeout).

We hit this behind a LiteLLM gateway. Under parallel load, tool calls and `list_tools` timed out together at the gateway's 60s cap, while the server itself logged no errors.

The fix runs the call through `asyncio.to_thread`. This is the same pattern `google_auth.py` already uses for the same function.

## Type of Change
- [ ] Documentation update

## Testing
- [ ] I have added tests that prove my fix is effective or that my feature works

`tests/auth`: 201 passed. `ruff check` and `ruff format` (0.15.22) are clean.

## Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas

## Additional Notes
Related earlier reports of stalls on this call path: #835 (missing socket timeout), #597 (IPv6 resolution hang). Both made the freeze longer; this change stops one stalled validation from blocking all other requests.

## Summary by CodeRabbit

* **Performance**
  * Improved responsiveness during external sign-in by isolating token validation from other application operations.
  * Added controlled validation capacity to prevent excessive concurrent authentication requests from overwhelming the service.
* **Bug Fixes**
  * Preserved existing token validation behavior and error handling.
  * Ensured authentication resources are released during normal shutdown and startup failures.


taylorwilsdon:
Thanks!

PR #1047, opened 2026-08-21
fix(gmail): stop reporting batch label changes that Gmail never made

123andy:
## The bug

`batch_modify_gmail_message_labels` returns `Labels updated for N messages` where **N is the length of the input list** — not a count of anything Gmail did. `users.messages.batchModify` answers `204 No Content` and **silently ignores IDs it does not recognise**, so a wrong or stale ID produces a confident success indistinguishable from a real one, on every retry, forever.

The easiest way to hit it: pass a **thread** ID where a **message** ID belongs (`draft_gmail_message` returns both; several tools return only thread IDs). Nothing is modified and the tool says it worked. Field report from our deployment: an agent called this three times on the same wrong ID, reported the target trashed each time, and it wasn't — a human caught it, not the tooling. For an agent caller a false success is worse than an error, because it suppresses the retry that would have fixed it. Blast radius is any bulk archive / mark-read / label sweep.

## The fix

After the `batchModify` call, read the messages back (`messages.get(format="minimal")`, batched with the same `new_batch_http_request` + sequential-fallback pattern the batch read tools use) and classify every ID:

- **Applied** — end state matches the request
- **Unchanged** — message exists, labels don't match
- **No such message** — the ID was silently ignored (reported by name, with a hint that a THREAD id may have been passed where a MESSAGE id is required)
- **Could not verify** — the read-back itself failed; no claim either way

```
Label changes for 2 message ID(s): Added labels: TRASH
Applied: 1/2
No such message (1): 1a025da58e17ce23
  These IDs do not resolve to a message in this mailbox, so Gmail ignored them. A common cause is passing a THREAD id where a MESSAGE id is required.
```

`verify=True` is the default — a destructive tool whose failure mode is a confident lie should be correct by default. `verify=False` opts out for very large sweeps and then says plainly that it recorded the request without verifying (`NOT VERIFIED: ...`), never claiming success.

## Tests

7 tests in `tests/gmail/test_batch_modify_label_verification.py`; 6 of them fail against the unpatched tool (the seventh pins that `batchModify` itself is still called once with all IDs — verification is additive, not a replacement). Covers: the thread-ID repro, mixed batches, removal-didn't-take, read-back failure → unverified, `verify=False` wording, and the sequential fallback when the batch endpoint is unavailable. Full suite passes.

Verified live against a deployed streamable-http instance: the exact thread-ID repro that motivated this now returns `Applied: 0/1` with the offending ID named.

## Summary by CodeRabbit

* **Bug Fixes**
  * Gmail label updates now verify whether requested changes were applied.
  * Results distinguish successfully updated, unchanged, missing, and unverifiable messages.
  * Invalid or stale IDs are no longer reported as successful updates.
  * Temporary verification failures are retried when possible.

* **New Features**
  * Added an option to skip verification when needed, with results clearly marked as unverified.
  * Verification results now provide clearer per-message outcome summaries.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -3930,9 +3930,7 @@ async def _verify_batch_label_changes(
 
         for mid in chunk_ids:
             entry = results.get(mid)
-            if not entry or (
-                entry.get("data") is None and entry.get("error") is None
-            ):
+            if not entry or (entry.get("data") is None and entry.get("error") is None):
                 statuses[mid] = "unverified"
                 continue
             error = entry.get("error")
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -3856,8 +3856,8 @@ async def _verify_batch_label_changes(
 
     Returns a per-ID status: ``applied`` (end state matches the request),
     ``not_applied`` (message exists, labels do not match), ``unresolved``
-    (no such message — the ID was silently ignored), or ``unverified``
-    (the read-back itself failed, so nothing can be claimed either way).
+    (the ID does not currently resolve to a message), or ``unverified`` (the
+    read-back itself failed, so nothing can be claimed either way).
     """
     wanted = set(add_label_ids or ())
     unwanted = set(remove_label_ids or ())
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
Thanks!

PR #1051, opened 2026-08-22
fix(drive): preserve shortcuts for metadata updates

lukaszliniewicz:
## Description
Fixes a Drive shortcut targeting bug in `update_drive_file`.

`update_drive_file` currently resolves Drive shortcuts before every update. That is appropriate for content operations, but unsafe for resource-local metadata changes: for example, calling `update_drive_file(file_id=<shortcut>, trashed=True)` sends `files.update()` to the shortcut target and trashes the underlying file instead of the shortcut.

This change makes shortcut handling depend on the kind of update:

- metadata-only updates (rename, move, trash, star, description, etc.) operate on the supplied Drive resource itself;
- content replacement keeps the existing behavior of following the shortcut to the underlying file;
- mixed content + metadata calls continue to apply both to the resolved target, preserving the historical content-update behavior.

The implementation keeps the existing resolver API backward-compatible by adding a keyword-only `follow_shortcuts=True` option and opting out only for metadata-only `update_drive_file` calls.

## Type of Change
- [ ] Documentation update

## Testing

Regression coverage verifies both sides of the contract: metadata-only updates preserve the shortcut ID, while content replacement still follows shortcuts.

Local validation:
- `uv sync --extra test --frozen`
- `uv run pytest` — 1694 passed, 2 skipped
- `uvx ruff@0.15.22 check`
- `uvx ruff@0.15.22 format --check`
- `uv run pytest tests/gdrive` — 135 passed
- `git diff --check`

## Checklist

## Additional Notes
The bug was observed against a real Drive shortcut: attempting to trash the shortcut instead trashed its target. The target was restored immediately. I have left the manual-testing checkbox unchecked because I did not want to mutate another live shortcut merely to exercise the patched path; the regression test models that exact `files.update()` target selection.

## Summary by CodeRabbit

* **New Features**
  * Added an option to preserve Google Drive shortcut information without following the shortcut to its target.
  * Metadata-only updates now apply directly to shortcuts, while content replacements continue updating the target file.

* **Bug Fixes**
  * Improved shortcut handling for file updates and metadata changes.
  * Prevented unsupported or mixed shortcut updates that could produce invalid results.
  * Reduced sensitive details in logs and error messages, including file names, URLs, and queries.
  * Added coverage for shortcut resolution and file update behavior.

### DIFF
--- a/gdrive/drive_helpers.py
+++ b/gdrive/drive_helpers.py
@@ -431,11 +431,16 @@ async def resolve_drive_item(
     *,
     extra_fields: Optional[str] = None,
     max_depth: int = 5,
+    follow_shortcuts: bool = True,
 ) -> Tuple[str, Dict[str, Any]]:
     """
-    Resolve a Drive shortcut so downstream callers operate on the real item.
+    Fetch Drive item metadata and optionally resolve shortcuts to their targets.
 
-    Returns the resolved file ID and its metadata. Raises if shortcut targets loop
+    By default shortcuts are followed so content-oriented callers operate on the real
+    item. Set ``follow_shortcuts=False`` for resource-local metadata mutations (rename,
+    move, trash, star, description, etc.) that must act on the shortcut itself.
+
+    Returns the selected file ID and its metadata. Raises if shortcut targets loop
     or exceed max_depth to avoid infinite recursion.
     """
     current_id = file_id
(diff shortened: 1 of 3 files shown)


taylorwilsdon:
Thanks!

PR #1055, opened 2026-08-25
feat(gdocs): support Google Docs suggested edits

### LINKED ISSUE #861: [feat request] Google Docs: Edit text using Suggestion mode
**Is your feature request related to a problem? Please describe.**
Since I can't use git with Google Docs, I need a way of reviewing the agent changes before accepting them. Or at least to see clearly what changed. I find the suggestion mode in Docs would be adequate for this. 

**Describe the solution you'd like**
Being able to use the Docs API to do edits in suggestion mode, instead of direct edits. That way we can see what changed in the file before accepting the change. 

**Describe alternatives you've considered**

- Git doesn't work for Google Docs, AFAIK. 
- Tell the agent to add the changes in comments, and then copy pasting them myself. 
- Direct edits, but then I lose track of what's changed and forces me to re-read everything looking for anything that might have been wrong. 

**Additional context**
Docs API supports Suggestions. 

PS. There is no community/Discord to ask questions about this project, is it?

mweholt:
## Summary

Adds support for Google Docs  **suggested edits** ([Work with comments and suggestions](https://developers.google.com/workspace/docs/api/how-tos/suggestions)) ((resolves #861 ):

- **`batch_update_doc` gains `suggest_mode`** — applies the whole batch as pending suggestions (`WriteControl.writeMode = SUGGEST`) instead of direct edits. Operations the API refuses in suggest mode (`insert_doc_tab`, `delete_doc_tab`, `update_doc_tab`, `create_named_range`, `delete_named_range`, `update_table_column_properties`, `create_header_footer`, and the `documentFormat` / `useEvenPageHeaderFooter` / `useFirstPageHeaderFooter` document-style fields) are rejected client-side with an actionable message rather than an opaque API error.
- **`list_doc_suggestions`** (new) — reads a document with `suggestionsViewMode=SUGGESTIONS_INLINE` and returns each pending suggestion's ID, type (insertion / deletion / style / paragraph-style / bullet / positioned-object), and a text preview. Covers body, headers, footers, and footnotes across the main document and every tab, and reads suggestion markers off paragraphs, paragraph elements, tables, rows, cells, and section breaks — a suggestion that isn't listed has no reachable ID, so it can't be accepted or rejected either.
- **`manage_doc_suggestions`** (new) — accept / reject / delete suggestion threads by ID via `AcceptSuggestionRequest` / `RejectSuggestionRequest` / `DeleteSuggestionRequest`.

### On the silent-fallthrough warning

These APIs are part of the Workspace Developer Preview Program, which exposes **no endpoint to query enrollment**. The failure mode that matters is silent: an unenrolled project may ignore `writeControl.writeMode` and apply "suggested" edits **directly**, so a caller believes they proposed a change when they actually made it.

`batch_update_doc` therefore diffs suggestion IDs before/after a suggest-mode batch and **appends a warning to that response** when the expected suggestion doesn't appear.

This is deliberately advisory and never a gate. The diff is a heuristic with real blind spots — an operation can withdraw an existing suggestion instead of adding one rather than growing the ID set, `find_replace` legitimately matches nothing, `insert_text` can be a no-op, and the walk stops recursing at depth 5. Misjudging a *working* project is both likelier and more damaging than the case being guarded against, so nothing is ever disabled on the strength of it. A per-user set of confirmed-working callers is kept purely to skip the two extra reads on later calls; a negative verdict is never stored, so it can't outlive its response or leak between users in multi-user deployments.

Happy to drop this piece if you'd rather not carry it — it's cleanly separable from the three tools.

## Test plan

- [x] `uv run ruff check .` — clean
- [x] `uv run pytest` — 1801 passed, 2 skipped
- [x] `tests/gdocs/test_suggest_mode_enrollment.py` (38 tests) covers detection plus what it must *not* do: no latching across calls, no leaking between users, `manage_doc_suggestions` unaffected by a prior warning
- [x] Live end-to-end against a real document on a Preview-enrolled project:
  - suggest-mode insert → suggestion created, no false warning
  - `find_replace` **with** a match → suggestion created, no false warning
  - `find_replace` with **zero** matches → no spurious warning (`NO_UPDATES_REQUESTED`)
  - insert into a **header** via `segment_id` → detected, no false warning
  - `list_doc_suggestions` → surfaced all three, including the header suggestion
  - `accept` merged one, `reject` discarded another, third left pending
  - suggestion-type coverage checked against ground truth by diffing the tool's
    output with a direct `documents.get` — bullet, paragraph-style, table, and
    inline-object suggestions all reported, and a previously-unreachable bullet
    suggestion accepted cleanly

Docstring coverage on diff-scoped functions: 60/60.

## Summary by CodeRabbit

* **New Features**
  * Added support for creating document changes as suggested edits.
  * Added tools to list, accept, reject, or delete document suggestions.
  * Improved suggestion details, including author, creation time, location, and change type.
  * Expanded suggestion detection across tables, headers, tabs, styles, and nested content.
  * Added clearer summaries of suggestions returned during batch updates.
* **Bug Fixes**
  * Improved confirmation of whether suggested edits were successfully saved.
  * Added clearer warnings when suggestion status cannot be confirmed.
  * Allowed header and footer creation in suggestion mode.


mweholt:
Took a few revisions with CodeRabbit but everything checks out now.

taylorwilsdon:
> Took a few revisions with CodeRabbit but everything checks out now.

It's honestly pretty solid with the profile I've got applied, if you put it in relaxed mode and tell it to ignore the type of nit picky garbage a lot of ai code review puts out you get mostly actionable stuff. Thanks for the contribution! Will get this in.

taylorwilsdon:
I took a pass at it and got it merge ready but when I was actually testing the functionality, I quickly became a little hesitant to merge this directly since the Workspace developer preview program isn't even possible to self-enroll - all you can do is submit a google form and hope they pick you, which is only likely if you're a real enterprise customer and not just an end user which a huge number of workspace mcp users are. Let me mull this over for a bit and look into whether google plans to actually ship this API broadly soon-ish.

mweholt:
@taylorwilsdon My understanding is they approve just about everyone. My workspace is for a small, non-software company and submitted the form and got approved in less than a day. But totally understandable if you want to hold off until the feature makes it out of preview.

kohanyirobert:
> My understanding is they approve just about everyone

Same experience. Would love this feature.

materemias:
i got approved in a few hours, their own beta mcp-s require it as well, so i think it will be the supported path

PR #1059, opened 2026-08-25
fix(utils): preserve Word text fidelity in extract_office_xml_text

warnes:
## Description

`extract_office_xml_text` loses text fidelity for Word documents in three ways that share one symptom: **a search of the extracted text finds nothing, and the caller cannot tell that apart from the term genuinely being absent.**

**1. A word split across runs is joined with a space.** Word splits a single token across runs routinely — spell-check state, formatting, tracked changes. The current loop strips each run and joins with `" "`, so:

```xml
<w:p><w:r><w:t>Project</w:t></w:r><w:r><w:t>X2024</w:t></w:r></w:p>
```

extracts as `Project X2024`, and a search for `ProjectX2024` returns nothing.

**2. Paragraph boundaries are discarded** — every paragraph in a member is joined with `" "`, running headings into body text.

**3. Only `word/document.xml` is read** — text in a header, footer, footnote or endnote (confidentiality banners, running titles, document numbers) is invisible to every caller.

### Changes

- Group `w:t` runs by their containing `w:p`; join runs of one paragraph with **no separator**. Runs are no longer stripped individually — under `xml:space="preserve"` those spaces are the real spacing between words.
- Join paragraphs within a member with newlines. Spreadsheets keep space-joining; they have no paragraphs.
- Include `word/header*.xml`, `footer*.xml`, `footnotes.xml`, `endnotes.xml` after the body, matched exactly so `settings.xml` is not scraped.
- Keep a flat-scan fallback for `w:t` outside any paragraph (some shapes and tables), so nothing previously found is lost.

## Type of Change
- [ ] Documentation update

## Testing

`tests/core/test_extract_office_xml_text.py`, 12 tests.

**Six fail against the current implementation and pass after the change**: the split-token case, both paragraph-boundary cases, and the three header/footer/footnote cases. I verified this by reverting `core/utils.py` and re-running — a test that passes either way proves nothing.

The other six are regression guards that pass both before and after: `xml:space` spacing, paragraph trimming, orphan `w:t` outside a paragraph, `settings.xml` not being scraped, corrupt input still returning `None`, and **spreadsheet extraction unchanged**.

```
uv run --extra test python -m pytest tests   ->  1788 passed, 2 skipped
uvx ruff check <changed files>               ->  clean
uvx ruff format --check <changed files>      ->  clean
```

## Checklist

## Additional Notes

One behaviour change worth flagging: extracted Word text now contains newlines between paragraphs where it previously contained spaces. Anything matching against the old single-line shape would need adjusting — I did not find such a caller in-repo.

Changes are confined to `core/utils.py`; no logic was added to any `*_tools.py` file.

## Summary by CodeRabbit

* **New Features**
  * Improved Word text extraction to include headers, footers, footnotes, and endnotes.
  * Preserved paragraph structure, run spacing, tabs, and line breaks in Word and PowerPoint content.
  * Added support for extracting text from tables, text boxes, and other structured office content.
  * Added fallback extraction when paragraph-based content is unavailable.

* **Bug Fixes**
  * Improved handling of duplicate markup, malformed office files, and corrupt archives.
  * Preserved existing spreadsheet cell text extraction behavior.

### DIFF
--- a/core/utils.py
+++ b/core/utils.py
@@ -407,29 +407,102 @@ def extract_office_xml_text(file_bytes: bytes, mime_type: str) -> Optional[str]:
                         # Runs are not stripped individually either: under
                         # xml:space="preserve" the leading/trailing spaces of a run
                         # are the real spacing between words.
-                        paragraphs = []
-                        for para in xml_root.iter():
-                            if not para.tag.endswith("}p"):
+                        # Every text node belongs to exactly one LINE, and every
+                        # line is assembled by the same rule: concatenate its runs
+                        # with no separator, strip once.
+                        #
+                        # Word splits a single token across runs routinely (spell
+                        # check state, formatting, tracked changes, hyperlinks), so
+                        # stripping each run and joining with " " turns
+                        # "ProjectX2024" into "Project X2024" and a search for the
+                        # identifier finds nothing.
+                        #
+                        # A line is normally a paragraph (w:p / a:p). Paragraphs
+                        # NEST — a text box inside a body paragraph carries its own
+                        # w:p — so a node is assigned to its CLOSEST paragraph
+                        # ancestor, never to every ancestor. Text with no paragraph
+                        # ancestor at all still forms a line, owned by its nearest
+                        # non-run container, so it is neither dropped nor split.
+                        # ElementTree has no parent links, hence the parent map.
+                        parents = {
+                            child: parent
+                            for parent in xml_root.iter()
+                            for child in parent
+                        }
+
+                        def _line_owner(node):
+                            """Closest paragraph ancestor, else nearest non-run container."""
+                            cur = parents.get(node)
+                            while cur is not None:
+                                if cur.tag.endswith("}p"):
+                                    return cur
+                                if not cur.tag.endswith("}r"):
+                                    parent = parents.get(cur)
+                                    while parent is not None:
+                                        if parent.tag.endswith("}p"):
+                                            return parent
+                                        parent = parents.get(parent)
+                                    return cur
+                                cur = parents.get(cur)
+                            return None
+
+                        # Markup-compatibility: Word writes a text box TWICE —
+                        # once under mc:Choice for modern readers and once under
+                        # mc:Fallback (VML) for old ones. Both carry the same
+                        # text, so scraping the subtree blindly emits it twice and
+                        # the duplicate is indistinguishable from a genuinely
+                        # repeated paragraph. Per the spec a consumer takes the
+                        # first Choice it understands, else the Fallback; we cannot
+                        # evaluate the Requires attribute, so take the first Choice
+                        # when there is one and ignore the rest of the alternatives.
+                        skip = set()
+                        for alt in xml_root.iter():
+                            if not alt.tag.endswith("}AlternateContent"):
                                 continue
-                            runs = [
-                                t.text
-                                for t in para.iter()
-                                if t.tag.endswith("}t") and t.text
-                            ]
-                            if runs:
-                                line = "".join(runs).strip()
-                                if line:
-                                    paragraphs.append(line)
-                        if paragraphs:
-                            member_texts.extend(paragraphs)
-                        else:
-                            # Fall back to the flat scan for text carried outside
-                            # any paragraph element (some shapes and tables).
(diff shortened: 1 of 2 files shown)


warnes:
**Review round 1 — thanks, both findings were real.** I verified each by reproducing the output before changing anything.

**The orphan-text fix I pushed first was wrong.** Making the flat scan always-on fixed the drop you identified, but it carried the flat scan's own fidelity bug into every document:

```
<w:tbl><w:r><w:t>Project</w:t></w:r><w:r><w:t>X2024</w:t></w:r></w:tbl>
  -> 'Project\nX2024'    the split-token bug this PR exists to fix
xml:space runs
  -> 'hello\nworld'      real spacing destroyed
orphan text
  -> emitted after all paragraphs, losing its document position
```

Both paths now share one rule instead of two: every text node belongs to exactly one line, and every line is `"".join(runs).strip()`. A line is the closest paragraph ancestor — paragraphs nest, so closest, never every ancestor — else the nearest non-run container. Lines emit in first-appearance order, which is document order.

**`mc:AlternateContent` was duplicating every text box.** Pre-existing, not introduced by this PR. Word 2010+ writes the box twice, under `mc:Choice` for modern readers and `mc:Fallback` (VML) for old ones, so the text appeared twice and the duplicate was indistinguishable from a genuinely repeated paragraph. Per the markup-compatibility rule a consumer takes the first Choice it understands, else the Fallback. `Requires` cannot be evaluated here, so the first Choice wins and the remaining alternatives are ignored. **A document carrying only a Fallback still yields its text** — a dedup that silently dropped content would be worse than the duplication, and there is a test pinning that.

**`w:tab` / `w:br` inside a run** now yield `\t` and `\n`. `"".join` otherwise fused `A<tab>B` into `AB` — the same false-token class this PR is about. Tab *stop* definitions under `w:pPr/w:tabs` are correctly not treated as content.

**Test gaps a mutation pass exposed (12 -> 25 tests):**

- **The ancestor walk was never exercised.** Every previous fixture put `w:t` exactly two levels under its paragraph, so replacing the walk with a fixed two-level lookup passed the entire suite. Added `w:hyperlink` and `w:ins` nesting — `w:ins` is this PR's flagship regression reappearing one level down.
- **The old "table" fixture was invalid OOXML** that Word cannot emit (`w:r` directly under `w:tbl`), so tables were only ever tested through an impossible document. Replaced with a real `w:tbl > w:tr > w:tc > w:p`.
- **PowerPoint `a:p` was untested**, and the split-token bug was live there.
- Added: tabs, breaks, tab-stop definitions, whitespace-only paragraphs, and both `mc:AlternateContent` shapes.

**Two existing tests could not fail.** The paragraph-trim test passed with `.strip()` deleted, because the whole output is stripped anyway — the padded paragraph is now in the middle. Header assertions used containment that could not distinguish separators; now exact equality.

Each fix was verified by disabling it individually and confirming the matching tests fail. Full suite: 1801 passed, 2 skipped. `ruff check` and `ruff format --check` clean.

warnes:
Note: the two commits on this branch were force-pushed once to reword their messages — no content changed, verified by comparing tree hashes before and after (identical). The detail that was in those messages now lives in the review response above, where it belongs: a commit message should carry what a reader needs to understand the code years later, not the story of the review.

warnes:
**Correction to something I claimed on both PRs.** I said these two are independent and can be merged in either order. That is wrong, and I found it by merging both onto `upstream/main` and running the suite.

They have no textual conflict — git merges them clean — but they disagree on one assertion:

- #1059 `tests/core/test_extract_office_xml_text.py::test_corrupt_zip_returns_none` asserts a corrupt archive returns `None`
- #1060 makes exactly that case raise `OfficeXmlExtractionError`

So **whichever merges second turns that test red.** The fix is one test: it should follow #1060's contract — a corrupt file is a finding, not an absence — and assert `pytest.raises(OfficeXmlExtractionError)` instead.

Nothing needs changing before either merges on its own; each is green against `upstream/main` today. I am flagging it so the second merge is not a surprise, and I am happy to push that one-test update to whichever PR you take second, once you have decided the order.

taylorwilsdon:
Not sure why the maintainer edits check keeps skipping, spun off to https://github.com/taylorwilsdon/google_workspace_mcp/pull/1071

PR #1060, opened 2026-08-25
fix(utils): distinguish an unreadable Office file from an empty one

warnes:
## Description

`extract_office_xml_text` returns `None` both when a file **cannot be read** and when it is read fine but **holds no text**. Callers cannot act on that distinction because it is not available to them.

Both call sites treat `None` as "no text" and fall through to a UTF-8 decode, so a corrupt `.docx` is reported to the user as:

```
[Binary or unsupported text encoding for mimeType '...' - N bytes]
```

That reads as *"this is a binary file"*, not *"this file is damaged"* — sending the reader after a format question instead of a data-integrity one. Upload pipelines do produce corrupt Office files, and classifying them as unsupported hides the real failure.

### Changes

- Add `OfficeXmlExtractionError`, raised when the file is not a valid ZIP, when its XML will not parse, or on an unexpected failure inside extraction. The original exception is chained with `raise ... from e`.
- **Keep `None` for the readable-but-empty case**, so that part of the contract is unchanged.
- Update both call sites to report the file as damaged instead of falling through to the binary branch.

## Type of Change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

Marked breaking deliberately: a function that previously never raised now can. See Additional Notes — I am happy to make it additive instead.

## Testing

`tests/core/test_office_extraction_errors.py`, 9 tests: unreadable inputs raise (not-a-zip, truncated, empty bytes), the message names the mime type, the cause is chained, and readable-but-empty still returns `None` — including a valid ZIP with no `word/document.xml`, and a document whose only text is whitespace.

```
uv run --extra test python -m pytest tests   ->  1785 passed, 2 skipped
uvx ruff check <changed files>               ->  clean
uvx ruff format --check <changed files>      ->  clean
```

## Checklist

## Additional Notes

**A trade-off I would rather surface than hide.** This changes the contract of a function that previously never raised. Both in-repo callers are updated and `core.utils` is not a documented public API, but an external importer relying on `None` would be affected.

If you would prefer it non-breaking, the same distinction can be made additively — an opt-in `raise_on_error: bool = False` parameter, default preserving today's behaviour exactly. Say the word and I will rework it; I chose the exception because it makes the failure impossible to ignore by accident, which is the property that was missing.

**On code organisation:** the caller changes are in `gdrive/drive_tools.py` and `gdocs/docs_tools.py`, which CONTRIBUTING flags as a common revision reason. I modified the existing inline fallback in place rather than adding new logic there, to keep the diff minimal. If you would rather this extraction-and-fallback block moved into `drive_helpers.py` / a docs helper, I am glad to do that — it looked like a larger refactor than a bug fix should carry unannounced.

Independent of #1059; the two can be merged in either order.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of damaged or invalid Office documents.
  * Document retrieval now reports a specific unreadable-file message instead of a misleading encoding or unsupported-format error.
  * Readable Office documents without text continue to be handled correctly.

* **Tests**
  * Added coverage for invalid, empty, and successfully readable Office documents.

### DIFF
--- a/core/utils.py
+++ b/core/utils.py
@@ -439,9 +439,15 @@ def extract_office_xml_text(file_bytes: bytes, mime_type: str) -> Optional[str]:
                 except OfficeXmlExtractionError:
                     raise
                 except KeyError:
-                    # The member simply is not in the archive. That is ABSENT, not
-                    # damaged: a readable file with nothing to extract yields None,
-                    # which is the distinction this whole change exists to keep.
+                    # ABSENT is not DAMAGED, and the difference is the contract:
+                    # this function returns None for a readable file with no
+                    # extractable text, and raises only when the file cannot be
+                    # read. A member that is simply not in the archive is the
+                    # former.
+                    #
+                    # Do NOT merge this into the generic handler below. That one
+                    # raises, so folding these together would report every archive
+                    # lacking an optional member as corrupt.
                     logger.info(
                         f"Member '{member}' not present in {mime_type} file; skipping."
                     )
--- a/core/utils.py
+++ b/core/utils.py
@@ -352,7 +352,13 @@ def extract_office_xml_text(file_bytes: bytes, mime_type: str) -> Optional[str]:
                         "No sharedStrings.xml found in Excel file (this is optional)."
                     )
                 except ET.ParseError as e:
+                    # Absent sharedStrings is optional (KeyError above). MALFORMED
+                    # is not: every t="s" cell resolves through it, so continuing
+                    # would silently produce wrong cell text.
                     logger.error(f"Error parsing sharedStrings.xml: {e}")
+                    raise OfficeXmlExtractionError(
+                        f"sharedStrings.xml is not parseable XML: {e}"
+                    ) from e
                 except (
                     Exception
                 ) as e:  # Catch any other unexpected error during sharedStrings parsing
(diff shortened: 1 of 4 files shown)


warnes:
**Review — confirmed, and fixed.** I reproduced it before changing anything: a malformed `word/document.xml` was caught per-member, logged, and fell through to `if not pieces: return None`, so `extract_office_xml_text` returned `None` for a damaged file. Exactly the conflation this PR set out to remove, surviving inside the PR.

Now raising `OfficeXmlExtractionError`:
- a member whose XML will not parse
- an unexpected failure reading a member
- a malformed `sharedStrings.xml` — every `t="s"` cell resolves through it, so continuing would silently produce wrong cell text

**My first attempt broke the distinction the PR is about.** Converting the generic member handler to raise also caught `KeyError`, so a readable ZIP with no `word/document.xml` started reporting as *damaged* rather than *empty*. An existing test caught it. `KeyError` is now handled separately and skipped, with a test either side — absent and malformed are different answers, and the code says so explicitly so it doesn't get "simplified" back.

**Tests 9 -> 13**: malformed document XML raises, the message names the member, malformed sharedStrings raises, and absent sharedStrings stays optional.

**One inconsistency I left, deliberately, rather than have you find it.** The generic `except Exception` around `sharedStrings.xml` still logs and continues; I converted only its `ParseError`. Widening that catch felt like scope a bug-fix PR should not take on its own — happy to include it if you would rather it were uniform.

Full suite 1789 passed, 2 skipped. `ruff check` and `ruff format --check` clean.

Both branches were force-pushed once to reword commit messages only — the trees are byte-identical, verified by comparing tree hashes before and after.

warnes:
**Correction to something I claimed on both PRs.** I said these two are independent and can be merged in either order. That is wrong, and I found it by merging both onto `upstream/main` and running the suite.

They have no textual conflict — git merges them clean — but they disagree on one assertion:

- #1059 `tests/core/test_extract_office_xml_text.py::test_corrupt_zip_returns_none` asserts a corrupt archive returns `None`
- #1060 makes exactly that case raise `OfficeXmlExtractionError`

So **whichever merges second turns that test red.** The fix is one test: it should follow #1060's contract — a corrupt file is a finding, not an absence — and assert `pytest.raises(OfficeXmlExtractionError)` instead.

Nothing needs changing before either merges on its own; each is green against `upstream/main` today. I am flagging it so the second merge is not a surprise, and I am happy to push that one-test update to whichever PR you take second, once you have decided the order.

warnes:
Closing in favour of #1101, which is the same change rebuilt on current `main`.

This branch had drifted 114 commits behind while carrying 3 of its own, and `mergeable` was `CONFLICTING` — the conflict is in `core/utils.py`, which #1071 rewrote when it landed #1059's work. Rather than force-push a branch that has already been reviewed, I opened a fresh PR so the review history here stays intact and readable.

**#1101 is not a replay of this diff.** Rebuilding against current `main` surfaced two paths this version missed:

- `word/document.xml` is now parsed *ahead of* the member loop, so its failure never reached the member-level handler.
- `_word_related_text_targets` returned `[]` on a malformed relationships part, silently dropping every header, footer and note — a different door onto the same conflation.

Both are covered there, with a negative control: neutering all 9 raise sites fails 12 of 19 tests with `DID NOT RAISE`, and the 7 that still pass are the readable-but-empty and success-path cases that must pass either way.

**Also retracting something I said here.** I commented on both this PR and #1059 that the two were not independent and could not be merged in either order. That is now stale — #1059's work reached `main` via #1071, so there is nothing left to sequence. #1101 stands alone against current `main`. Apologies for leaving an outdated ordering constraint on the record; a reviewer acting on it would have been looking for a dependency that no longer exists.

Thank you for spinning off #1071 to get #1059's fix in despite the maintainer-edits check misbehaving. #1101 has `maintainer_can_modify=true`, so that check should pass on it.

PR #1061, opened 2026-08-25
feat(auth): expose OAuthProxy token expiry knobs as env vars

DrFaust92:
## Description

FastMCP's `GoogleProvider` accepts `token_expiry_threshold_seconds` and `fastmcp_access_token_expiry_seconds`, but `configure_server_for_http()` passes neither, so a deployment cannot reach either without patching the server. This reads both from the environment and forwards them, following the `WORKSPACE_MCP_ALLOWED_CLIENT_REDIRECT_URIS` pattern a few lines above:

- `WORKSPACE_MCP_OAUTH_PROXY_TOKEN_EXPIRY_THRESHOLD_SECONDS`
- `WORKSPACE_MCP_OAUTH_PROXY_ACCESS_TOKEN_EXPIRY_SECONDS`

Unset, empty, non-integer and negative values keep FastMCP's own defaults (`0` and `None`), so existing deployments are unaffected.

Why they are worth reaching: `OAuthProxy` rotates its client-facing refresh token one-time-use with no grace window, so when two refreshes race on the same token the loser is rejected with `invalid_grant` and the user is pushed through a full re-authentication (PrefectHQ/fastmcp#4265, PrefectHQ/fastmcp#4901). Raising the FastMCP access-token lifetime above Google's hour makes refreshes rarer, and a non-zero threshold refreshes early rather than at the edge — both narrow the window. On our deployment this currently shows up as roughly 14 `Refresh token not found ... already rotated` warnings a day across independent clients.

## Type of Change

## Testing

Three cases covered: defaults preserved when unset, values forwarded when set, unusable values ignored.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional environment-based configuration for OAuth access-token lifetime and early refresh timing.
  * Settings now apply consistently across supported OAuth providers.
  * Valid values are applied automatically; leaving them unset preserves default behavior.
* **Bug Fixes**
  * Invalid, negative, empty, or out-of-range values are safely ignored.
  * Reduced concurrent token-refresh race frequency without adding a grace period to one-time-use refresh tokens.
* **Documentation**
  * Documented configuration options, supported ranges, defaults, and usage guidance.

### DIFF
--- a/.env.oauth21
+++ b/.env.oauth21
@@ -16,6 +16,28 @@ OAUTH2_ENABLE_DEBUG=false
 OAUTH2_ENABLE_LEGACY_AUTH=true
 
 # ---------------------------------------------------------------------------
+# FastMCP OAuth Proxy Token Lifetime Tuning (OAuth 2.1, optional)
+#
+# These settings mitigate client refresh-token races by making client-facing
+# access-token refreshes less frequent and refreshing Google's upstream access
+# token shortly before expiry. They do not add a grace period to FastMCP's
+# one-time-use client refresh tokens.
+#
+# Leave both unset to preserve FastMCP's defaults. Invalid or out-of-range values
+# are ignored with a warning.
+#
+# Refresh Google's upstream access token early. Allowed range: 0-300 seconds.
+# A small value such as 30-120 seconds is recommended; values near Google's
+# access-token lifetime would otherwise cause a refresh on every request.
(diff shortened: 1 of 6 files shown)


taylorwilsdon:
Created a new helper module for the oauth logic and took a pass at this, will get in today - thanks! 

Hopefully not a permanent solution but happy to expose a workaround for the time being.

PR #1062, opened 2026-08-25
fix(gmail): prevent duplicate delivery after an ambiguous send

### LINKED ISSUE #997: send_gmail_message retries a non-idempotent send, so a lost 200 becomes a duplicate delivery

justadityaraj:
### Problem

Gmail's `messages.send` operation is non-idempotent. The existing `num_retries=3` setting could resend a message after an ambiguous transport or server response even when Gmail had already accepted the first request, resulting in duplicate delivery.

### Changes

- Disable googleapiclient retries for new-message and forward sends.
- Add regression assertions for both send paths.
- Preserve the existing retry behavior for draft creation and other write operations.

### User impact

An ambiguous Gmail send now returns its error to the caller instead of silently risking duplicate delivery. Successful sends are unchanged.

Fixes #997

## Type of Change
- [ ] Documentation update

## Testing

## Verification

- `uv run pytest tests/gmail/test_draft_gmail_message.py tests/gmail/test_forward_gmail_message.py -q` — 64 passed.
- `uv run pytest -q` — 1,772 passed and 2 skipped; four pre-existing Windows-specific permission/home-path assertions failed outside the changed Gmail code.
- `uvx ruff@0.15.2 check` — passed.
- `uvx ruff@0.15.2 format --check` — 193 files already formatted.

## Checklist

## Additional Notes

Draft creation remains retryable because it does not deliver a message to recipients.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when sending and forwarding Gmail messages by preventing automatic write retries.
  * Ensured each send operation is executed only once, reducing the risk of duplicate messages.

* **Tests**
  * Added coverage verifying the updated send behavior for new and forwarded messages.

### DIFF
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -2643,7 +2643,7 @@ async def send_gmail_message(
     # Send the message
     sent_message = await asyncio.to_thread(
         service.users().messages().send(userId="me", body=send_body).execute,
-        num_retries=GOOGLE_API_WRITE_RETRIES,
+        num_retries=0,
     )
     message_id = sent_message.get("id")
 
@@ -2767,7 +2767,7 @@ async def _forward_gmail_message_impl(
     # Send the message
     sent_message = await asyncio.to_thread(
         service.users().messages().send(userId="me", body=send_body).execute,
-        num_retries=GOOGLE_API_WRITE_RETRIES,
+        num_retries=0,
     )
     sent_message_id = sent_message.get("id")
 
--- a/tests/gmail/test_draft_gmail_message.py
+++ b/tests/gmail/test_draft_gmail_message.py
@@ -1665,3 +1665,6 @@ async def test_send_gmail_message_does_not_fetch_thread_for_a_new_message():
     )
 
     assert mock_service.users.return_value.threads.return_value.get.call_count == 0
+    mock_service.users().messages().send.return_value.execute.assert_called_once_with(
+        num_retries=0
+    )
--- a/tests/gmail/test_forward_gmail_message.py
+++ b/tests/gmail/test_forward_gmail_message.py
@@ -160,6 +160,9 @@ async def test_forward_simple_text_email():
     assert "Email forwarded" in result
     assert "fwd001" in result
     mock_service.users().messages().send.assert_called()
+    mock_service.users().messages().send.return_value.execute.assert_called_once_with(
+        num_retries=0
+    )
 
     sent = get_sent_mime_message(mock_service)
     assert sent["Subject"] == "Fwd: Hello"


taylorwilsdon:
Now that's an interesting one - in theory, I don't see how you could get a failure code back when a message was sent but in practice you never know (and a timeout that ultimately succeeds behind the scenes is still a failure, I suppose) 

Will get some form of this merged thank you!

justadityaraj:
yeah just a timeout path.

Thanks!

ConnorMoss02:
Checked the other call sites: `messages().send()` only appears twice in `gmail_tools.py`, at `:2828` and `:2975` on current main and it looks like you covered both. The only other write retry in that file is draft creation at `:3275` and that one never reaches a recipient.

justadityaraj:
@ConnorMoss02 thanks for checking, confirmed independently on `fb89c81`. `messages().send()` is exactly those two sites, `:3275` is `drafts().create()`, and there's no `drafts().send()` anywhere in the repo, so nothing else can put mail in front of a recipient.

@taylorwilsdon on the doubt: you're right that getting a *failure code* back for a sent message is odd, and the PR doesn't rest on that. The case is getting **no response at all**.

`num_retries` is handled by `googleapiclient.http._retry_request`, which retries when `http.request()` raises `ssl.SSLError`, `socket.timeout`, `ConnectionError`, an `OSError` with errno `ETIMEDOUT`/`EPIPE`/`ECONNABORTED`/`ECONNRESET`/`ECONNREFUSED`, or `ServerNotFoundError` — plus any 5xx/429 via `_should_retry_response`.

`socket.timeout` is the load-bearing one, and it is structurally post-acceptance: HTTP/1.1 writes the entire request before it reads, so a read timeout can only fire once the body is already with Google. `auth/google_auth.py:97` sets `httplib2.Http(timeout=30)`, so it's a live path here. Gmail has accepted and delivered, the client never sees the 200, and `_retry_request` sends it again — it can't distinguish that from a request that never landed, because all it sees is the exception type.

Not purely theoretical: googleapis/google-api-python-client#1474 is this exact call with `socket.timeout`, and the recipient received 4 identical copies. Being straight about it though — that issue was closed `not_planned` and Google never conceded a defect, so the load-bearing part is the reporter's observation, not a Google ruling.

There's also nothing to clean up afterwards: `_prepare_gmail_message` never sets a `Message-ID`, so Gmail mints a fresh one per accepted send. Two accepted sends are two distinct messages with nothing for any dedupe layer to collapse.

The repo already takes this position in two places:

- `gmail_tools.py:2422` declares `idempotentHint=False` on the send tool, while the transport underneath replays it
- `gdrive/drive_tools.py:2224` — your `0fed4f5` — already varies retries by whether a replay is safe: `num_retries=GOOGLE_API_WRITE_RETRIES if replacing_content else 0`

Honest about the trade: transient 5xx/429 are almost certainly more common than ambiguous post-accept failures, so some sends that used to self-heal invisibly will now surface as errors. The case is asymmetry rather than frequency — a surfaced error is visible and a human can retry it, while a delivered duplicate is silent, irreversible, and lands in someone else's inbox.

**One gap in my own diff, worth flagging.** With `num_retries=0` the transport failure now reaches `handle_http_errors`, and `send_gmail_message` is `is_read_only=False`, so it either falls through to the generic `except Exception` (`"An unexpected error occurred in send_gmail_message: timed out"`) or, for SSL, raises `TransientNetworkError(... "Please try again shortly.")`. Both tell the caller — an LLM, which will do as it's told — to retry a send that may already have gone out. So this stops the library double-sending and then invites the agent to do it instead.

Happy to add a send-specific message along the lines of "this may or may not have been delivered, check Sent before retrying". Say the word and I'll push it here, or leave this diff minimal and do it as a follow-up.

PR #1065, opened 2026-08-26
fix(gmail): default replies to latest eligible thread message

floriangaldo:
## Description

Fix Gmail replies and reply drafts sometimes targeting an older message or an existing draft instead of the latest real message in a thread.

This change:

- Defaults replies to the latest non-`DRAFT`, non-`TRASH` message containing an RFC `Message-ID`.
- Derives `In-Reply-To` and `References` from the selected parent's actual RFC ancestry, including branched threads.
- Preserves an explicitly selected older reply target.
- Keeps recipients, quoted content, and reply headers aligned with the same target.
- Treats `References` as ancestry rather than an implicit reply-target selector.
- Updates tool descriptions and agent guidance so normal replies provide `thread_id` and omit low-level reply headers.

## Type of Change

- [ ] Breaking change
- [x] Documentation update

## Testing

- [x] I have added tests that prove the fix is effective
- [x] New and existing unit tests pass locally

Added coverage for:

- Existing drafts appearing after the latest real message.
- References-only calls defaulting to the latest reply target.
- Branched RFC ancestry.
- Trashed ancestors.
- Explicit older and trashed reply targets.
- Messages without an RFC `Message-ID`.

Full result: `1,782 passed, 2 skipped`.

Live Gmail validation on the Inyo account created two disposable reply drafts in the same existing thread, with the second created while the first draft was present. Both retained the expected thread, reply parent, and RFC ancestry. Nothing was sent, and both test drafts were deleted afterward.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review
- [x] I have commented the non-obvious logic

## Summary by CodeRabbit

* **New Features**
  * Gmail replies and drafts now automatically target the latest eligible message in a thread.
  * Reply ancestry, including `In-Reply-To` and `References` headers, is derived automatically.
  * Specific older or trashed messages can still be selected explicitly.

* **Documentation**
  * Updated Gmail guidance and examples to reflect automatic reply targeting and threading.

* **Tests**
  * Expanded coverage for reply-parent selection, draft exclusion, missing headers, and preserved ancestry.

### DIFF
--- a/gmail/gmail_helpers.py
+++ b/gmail/gmail_helpers.py
@@ -213,7 +213,10 @@ def _derive_reply_headers(
     references: Optional[str],
     target: Optional[Mapping[str, Any]] = None,
 ) -> tuple[Optional[str], Optional[str]]:
-    """Fill reply headers, defaulting to the latest thread message."""
+    """Fill reply headers, defaulting to the latest automatically eligible message.
+
+    Automatic targets are non-draft, non-trash messages with an RFC Message-ID.
+    """
     derived_in_reply_to = in_reply_to
     derived_references = references
 
--- a/gmail/gmail_helpers.py
+++ b/gmail/gmail_helpers.py
@@ -199,37 +199,46 @@ def _parse_date_header(
 
 
 def _parse_message_id_chain(header_value: Optional[str]) -> list[str]:
-    """Extract Message-IDs from a reply header value."""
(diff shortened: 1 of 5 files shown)


taylorwilsdon:
Seems like this would only apply to very specific edge cases where the conversation branches to the point that target.message_id and thread_message_ids[-1] deviate but edge cases are worth addressing haha thank you!

PR #1070, opened 2026-08-27
feat(accounts): explicit multi-account routing, with no silent fallback

jonathanmoregard:
Lets one server hold two or more authenticated Google accounts that differ in
Developer Preview capability, and lets an agent route between them explicitly:
a preferred default, honest per-account capability, and errors that name the
alternatives without ever reaching for them.

**There is no automatic fallback, and that is the point.** Drive documents that
`404 notFound` covers both "no read access" and "the file does not exist", so a
retry trigger built on it is unreliable by construction — a typo'd document id
would walk every credential in the wallet, turning a mistake into an
enumeration sweep across identities. Instead the affordance is surfaced and the
action is not taken: a miss returns a structured error naming the other
configured accounts, and a second explicit call is required to use one. Writes
never auto-switch identity either; a comment landing under the wrong account is
visible to everyone in the document and is not revertable in their eyes.

With exactly one account authenticated, behaviour and the server instructions
string are byte-identical to today. Multi-account output appears only when the
credential store actually holds more than one account.

## What is here

- `core/account_directory.py` (new) — enumeration, capability lookup,
  instructions builder, candidate-account hints. Fork-owned.
- `core/email_identity.py` (new) — the single `fold_email` rule, a leaf module
  with no project imports.
- `core/env_flags.py` (new) — one strict parser for the boolean mode flags.
- `list_google_accounts` — read-only, zero network calls, never probes. Reports
  `GCSCredentialStore.list_users()` raising `NotImplementedError` honestly
  rather than crashing.
- The AC7 arbitrary-pick warning: `_find_any_credentials` still takes
  `users[0]`, but now says so, names every account it found, and says how to pin
  one.
- An HTTP origin/Host middleware rewrite — see below.

## Scope note: this grew an HTTP security fix

The review loop found a DNS-rebinding hole in `OriginValidationMiddleware`
while reviewing this branch, and it was fixed here rather than on its own
branch. Two further rounds found more in the same guard. The result is that
this PR contains a real behaviour change unrelated to multi-account routing:

**Reaching the server on any name other than loopback — a LAN IP, or a reverse
proxy passing the original hostname through — now requires declaring that name**
via `WORKSPACE_EXTERNAL_URL` or `OAUTH_ALLOWED_ORIGINS`. The `Host` header is
now checked on every request rather than only on requests carrying an `Origin`,
because per the Fetch standard a browser appends `Origin` only for methods other
than GET and HEAD, and a rebound page's GETs are same-origin to the browser —
so an Origin-gated check never ran on most of what rebinding actually produces.
`main.py` defaults the HTTP bind to `0.0.0.0`, which is the exposure this
defends. Documented in `HANDOVER.md` 2.4.

An OAuth 2.1 deployment already needs `WORKSPACE_EXTERNAL_URL` for correct
issuer metadata, so in practice only legacy HTTP mode behind a proxy is newly
affected.

`c7d5492` (one strict parser for the boolean mode flags) is independently
upstreamable as a standalone fix and does not depend on the rest of this branch.

## Review coverage, and one gap

Reviewed under the advice-refine-test-loop skill. Every finding was reproduced
before being fixed; nothing was fixed on an advisor's word alone.

- Phase 1 (Opus + narrow specialists): 4 rounds.
- **Phase 2 (Codex / GPT-5.5) never ran — quota exhausted** on the default model
  and on the one sanctioned fallback. Shipping without cross-vendor coverage was
  a conscious decision, not an oversight. The defect class Codex reliably catches
  that Opus misses is inference-by-absence: treating a coverage gap (no test, no
  branch, no error path) as evidence the case is handled. **That gap is real and
  this PR is not covered against it.**
- Phase 3 (adversarial): 1 round, which found the original rebinding hole.
- Phase 0 deterministic gate re-run every round: 0 findings, gitleaks clean,
  0 waivers.

Local verification on a settled tree at the head commit: 2750 passed, 84
skipped, 0 failed; `ruff 0.15.22 check` and `format --check` both clean.

## Surfaced during review and deliberately NOT fixed here

- `MCP_SINGLE_USER_MODE` is compared `== "1"` in `auth/google_auth.py:1099` and
  `:752`, `auth/oauth21_session_store.py:631` and
  `auth/oauth_callback_server.py:102`, while the startup banner parses it
  strictly. Sibling of `c7d5492`, upstream-bound.
- `auth/google_auth.py:68`'s loopback test is a substring match, so
  `https://localhost.example.com` sets `OAUTHLIB_INSECURE_TRANSPORT`
  process-wide and `[::1]` is missed.
- `OAUTHLIB_INSECURE_TRANSPORT=0` renders as "off" in the banner while oauthlib
  treats any non-empty value as set.
- `unavailable` means three different things across `store_status`,
  `docs_preview.availability` and `verification.source`.
- Instructions are per-process, so an account added later via
  `start_google_auth` leaves them stale while `list_google_accounts` is already
  correct.
- `suggestion_ledger.observe()` is fed only modelled cards, so an unmodelled
  OPEN suggestion id never enters `last_read_ids`, and `explain_missing`'s final
  rung answers "most likely the id is wrong" for an id the same complete read
  listed as OPEN.
- `auth/credential_store.py:547` reads `MCP_ENABLE_OAUTH21` from `os.getenv`
  rather than the config object. Deliberate — it runs pre-reload.
- The `ruff` dev-dependency bound stays `ruff>=0.12.4` while upstream is
  `>=0.15.22,<0.16`. Aligning forces a `uv.lock` regeneration, left for a
  standalone lock-bump PR.

## Still open

Whether Developer Preview enrollment is per-Cloud-project, per-account, or both
is genuinely undocumented for the two-accounts-one-OAuth-client case, and is
**not** resolved by assertion anywhere in this branch. Capability is modelled as
per-account and unknown-until-observed, which is correct under either answer.
`HANDOVER.md` 2.1 documents the two experiments that would settle it.

Authenticating a second Google account needs an interactive browser login, so
the feature is tested against a fake credential store holding two accounts
rather than live ones.

## Summary by CodeRabbit

- **New Features**
  - Added a Docs Preview service for reviewing, filtering, paginating, and resolving Google Docs suggestions and comments.
  - Added support for nested document tabs, headers, footers, footnotes, and scoped review ranges.
  - Added account discovery with clearer account-aware guidance and a read-only account listing.
  - Added mock and end-to-end review scenarios, including high-volume and concurrent-edit workflows.

- **Bug Fixes**
  - Improved error reporting so failed document operations are no longer shown as successful.
  - Strengthened OAuth callback host validation and local-only insecure transport handling.
  - Improved detection and reporting of unmodelled suggestions and degraded reads.

- **Documentation**
  - Added setup, API, review workflow, findings, and scenario documentation.


jonathanmoregard:
Opened against this repo by mistake — this is fork-specific work and belongs on my own fork. Closing immediately, sorry for the noise.

PR #1083, opened 2026-08-31
feat(gmail): add label colors to manage_gmail_label

### LINKED ISSUE #351: Add label color support to manage_gmail_label
## Feature Request

  The `manage_gmail_label` tool currently supports creating, updating, and deleting labels, but doesn't support setting label colors.

  ### Current parameters:
  - action (create/update/delete)
  - name
  - label_id
  - label_list_visibility
  - message_list_visibility

  ### Requested addition:
  - `background_color` - hex color code (e.g., "#ff0000")
  - `text_color` - hex color code (optional, defaults based on background)

  ### Use case:
  When automating inbox triage with many labels, colors help visually differentiate priority levels (e.g., red for urgent, gray for low-priority).

  The Gmail API supports this via the `color` field in [labels.update](https://developers.google.com/gmail/api/reference/rest/v1/users.labels#Label).

ConnorMoss02:
## Description

Adds two optional parameters to `manage_gmail_label`. Closes #351. Defaults are unchanged, so existing callers are unaffected until they opt in.

- **`background_color`** and **`text_color`** set a label's color on `create` and `update`, as hex strings like `"#fb4c2f"`.
- Validation lives in `gmail/gmail_helpers.py` as `build_label_color()`, backed by `GMAIL_LABEL_COLORS`.
- Both are required together, and passing neither on `update` keeps the label's current color.

### Why

Gmail accepts label colors only from its own palette and answers anything else with a bare 400 that does not name the offending field. Validating up front turns that into a message naming the value and linking the palette.

Three details shaped the implementation:

- **Colors come in pairs.** Gmail requires `backgroundColor` and `textColor` together whenever `color` is set, so a half-specified color is rejected before the request rather than sent and refused.
- **`update` does a read-modify-write** and Gmail replaces the whole `color` object, so a rename that does not mention color re-sends the label's existing color instead of dropping it.
- **Colors apply to `type: user` labels.** System labels reject them, noted in the parameter docs rather than enforced client-side, since the create path does not otherwise fetch label type.

The palette is generated from the `LabelColor` schema in the [Gmail v1 discovery document](https://gmail.googleapis.com/$discovery/rest?version=v1) rather than transcribed from the reference page: 113 colors, with `backgroundColor` and `textColor` publishing an identical set.

## Type of Change
- [ ] Documentation update

## Testing

15 tests in `tests/gmail/test_manage_gmail_label_color.py`: create and update with color, preservation when no color is passed, rejection of off-palette values and of half-specified colors before any request is made, and the unchanged default bodies for both create and update. Full Gmail suite passes (264). `ruff@0.15.22 check` and `format --check` clean.

Not exercised against a live account, so the request shapes are asserted against a mocked service rather than Gmail itself.

## Checklist

## One thing I noticed but left alone

The `update` branch rebuilds `labelListVisibility` and `messageListVisibility` from the parameter defaults rather than from the fetched label, so updating only a label's name resets a `labelHide` label back to `labelShow`. It predates this change and is a behavior fix rather than a feature, so I have not touched it here. Happy to send it separately if useful.

## Summary by CodeRabbit

- **New Features**
  - Set Gmail label background and text colors using the complete supported color palette.
  - Clear an existing label color during updates with the new color-clearing option.

- **Bug Fixes**
  - Invalid or incomplete color combinations are rejected before changes are made.
  - Updating without color options preserves the existing color.
  - Color options are ignored when deleting labels.

- **Documentation**
  - Updated guidance for setting, preserving, and clearing Gmail label colors.


ConnorMoss02:
Both addressed in 6a6308b.

**Incomplete color objects.** Correct, and it resolves the open question in my description. A half-specified color is now rejected in `build_label_color()` before any request, on create and update alike. I dropped the merge-with-current behaviour rather than using it to complete a partial color: it made the same call succeed or fail depending on whether the label already had a color. The contract is now both or neither. Passing neither on update still keeps the label's existing color.

**Palette size.** Neither 102 nor 103. I pulled the `LabelColor` schema from the Gmail v1 discovery document (`https://gmail.googleapis.com/$discovery/rest?version=v1`) and diffed it against what I had shipped: my list was missing 11 valid colors, including `#d93025`, `#202124` and `#757575`. Anyone passing one of those would have hit a wrongly-rejected error from my validator. The constant is now generated from that schema at 113 colors, and `backgroundColor` and `textColor` publish an identical set.

Tests are 15 now, including the two cases you asked for: a half-specified color on create, and a half-specified color updating a label with no existing color. Full Gmail suite passes at 264. `ruff@0.15.22 check` and `format --check` clean.

ConnorMoss02:
Fixed in cce49e1. The color is now built once before the action branches, so a half-specified color raises without any Gmail request on either path. The update test asserts `get` is not called, not just `update`.

That also removed the duplicate `build_label_color` call the two branches had.

Gmail suite passes at 264, ruff clean.

taylorwilsdon:
Huh, would have sworn we had that in place, the colors must have missed the main place it's useful in manage haha - good add thank you!!

Issue #1086, opened 2026-09-01
Docs read tools cannot target a tab: tab_id is on the write tools but not on get_doc_content / get_doc_as_markdown


DrFaust92:
Six gdocs tools already accept `tab_id` — `modify_doc_text`, `find_and_replace_doc`, `inspect_doc_structure`, `update_paragraph_style`, `create_table_with_data`, `manage_doc_tab` — and `batch_update_doc` accepts it per operation. The two read tools do not:

- `get_doc_content` ([`docs_tools.py:149`](https://github.com/taylorwilsdon/google_workspace_mcp/blob/main/gdocs/docs_tools.py#L149))
- `get_doc_as_markdown` ([`docs_tools.py:2445`](https://github.com/taylorwilsdon/google_workspace_mcp/blob/main/gdocs/docs_tools.py#L2445))

Both call `documents.get(..., includeTabsContent=True)` and concatenate every tab into a single string with `--- TAB: ... ---` markers.

So you can *write* to a specific tab, but you cannot *read* that tab back. The only way to see one tab's content is to pull the whole document and slice it apart client-side.

#113 added multi-tab support to `get_doc_content`, which was the right first step — it fixed "only the first tab is visible". This is the follow-on: tab *selection*.

### Why it matters

Tabs are a natural fit for documents that accumulate an entry per week or month — a running report, a log, a review doc. That is exactly the shape where reading one tab is the common operation and reading all of them almost never is.

I hit this on a ~40-tab document where the tab I needed was ~10k characters. Getting those 10k characters meant fetching every tab in the document. The cost lands twice: once in API latency (#1052) and once in the client's context window, which is the same argument #638 makes for section-level reads.

The asymmetry is also just surprising in use — an agent that has been happily passing `tab_id` to `inspect_doc_structure` and `batch_update_doc` gets a validation error the moment it passes the same argument to `get_doc_content`.

### Suggested fix

Add `tab_id: str | None = None` to both read tools, matching the parameter that already exists on the write tools. The lookup helpers are already in the codebase — `find_tab` inside `inspect_doc_structure` and `_find_tab_end_index` at `docs_tools.py:2562` — so this is mostly wiring plus a shared helper.

Parameters on existing tools, no new tools, consistent with CONTRIBUTING's guidance on tool count.

Worth noting this composes with #1052: once the caller has named a tab, the `fields` mask can narrow the fetch itself rather than only the output.

[...]

Issue #1084, opened 2026-09-01
inspect_doc_structure caps text_preview at 100 characters, so the index workflow it documents cannot be completed


DrFaust92:
The docstring positions this tool as *the* way to get indices for formatting:

> **FORMATTING WORKFLOW:** After inserting all text via batch_update_doc with end_of_segment=true, call this tool with detailed=true to get exact start_index and end_index for every paragraph. Use those indices directly in format_text and update_paragraph_style operations in a second batch_update_doc call.

That holds when the unit you want to style is a whole paragraph. It breaks for anything *inside* a paragraph, because `text_preview` is hard-truncated to 100 characters with no override:

- [`gdocs/docs_tools.py:1513`](https://github.com/taylorwilsdon/google_workspace_mcp/blob/main/gdocs/docs_tools.py#L1513) — `elem_summary["text_preview"] = element.get("text", "")[:100]`
- [`gdocs/docs_structure.py:195`](https://github.com/taylorwilsdon/google_workspace_mcp/blob/main/gdocs/docs_structure.py#L195) — `"text_preview": "".join(text_parts)[:100]`

For any paragraph longer than 100 characters there is no way to compute the offset of a token within it. The obvious fallback is to take the text from `get_doc_content` and search it — but that output is not index-aligned with the document (#1085), so the offsets it yields are wrong. That leaves no reliable path to a sub-paragraph index.

**Concrete case.** Colouring every week-over-week delta token (`+18.7%`, `-37.8%`, …) in a generated report: a few dozen short tokens scattered through paragraphs of 300–1000 characters, each needing its own `format_text` range. Paragraph boundaries alone are not enough. The tool advertises the workflow but stops one step short of enabling it, and the gap is one slice expression.

### Suggested fix

Add `preview_chars: int | None = None` to `inspect_doc_structure`, defaulting to the current 100 so existing callers see no change, with `0` or `None` meaning no truncation. A parameter on an existing tool rather than a new tool, per CONTRIBUTING's guidance on tool count.

Happy to implement this — small change plus a test. Let me know if the default and parameter name look right to you and I'll open a PR.

PR #1087, opened 2026-09-01
fix(gdocs): keep get_doc_content output aligned with document indices

### LINKED ISSUE #1085: get_doc_content drops empty paragraphs and non-text runs, so its output cannot be used to compute document indices
`get_doc_content` builds its return string with:

```python
if current_line_text.strip():
    text_lines.append(current_line_text)
```

([`gdocs/docs_tools.py:231`](https://github.com/taylorwilsdon/google_workspace_mcp/blob/main/gdocs/docs_tools.py#L231))

An empty paragraph's content is `"\n"`, so `.strip()` is falsy and the paragraph is dropped entirely. The same loop accumulates only `textRun` content, so `inlineObjectElement`, `pageBreak`, `columnBreak`, `horizontalRule`, `footnoteReference`, `person` and `richLink` are skipped as well — each of which occupies index positions in the document.

The consequence is that offsets computed from `get_doc_content` output do not correspond to the document's real indices. Silently, and cumulatively.

### Reproduction

Create a document whose body is `Alpha`, an empty paragraph, `Bravo`, an empty paragraph.

`inspect_doc_structure(detailed=True)` reports it correctly — body spans indices 1–15, i.e. 14 characters:

| paragraph | range | text |
|---|---|---|
| 1 | 1–7 | `Alpha\n` |
| 2 | 7–8 | `\n` |
| 3 | 8–14 | `Bravo\n` |
| 4 | 14–15 | `\n` |

`get_doc_content` on the same document returns:

```
Alpha
Bravo
```

12 characters. Both empty paragraphs are gone.

Now insert a page break. `total_length` goes 15 → 17, and `get_doc_content` returns the identical 12 characters — the non-`textRun` element is invisible to it too.

### Why it matters

Index drift. An agent that reads the text, locates a substring in it, and passes that offset to `format_text` / `delete_text` / `replace_text` in `batch_update_doc` will act on the wrong range — off by one index per preceding empty paragraph, plus one per non-text element. Nothing errors; the edit just lands in the wrong place.

`inspect_doc_structure` does give correct paragraph boundaries, but its `text_preview` is capped at 100 characters (#1084), so it cannot be used to locate anything *inside* a paragraph. Between the two tools there is currently no reliable way to compute a sub-paragraph index.

Measured on a real document: a 105-paragraph tab whose body spans 10,789 characters came back as 10,740 — exactly the count of its 49 empty paragraphs. Reconstructing the dropped paragraphs from `inspect_doc_structure`'s boundaries was the only way to get offsets that matched.

### Suggested fix

Append `current_line_text` unconditionally (drop the `.strip()` guard), and emit the single-character placeholder for non-`textRun` elements, so the returned string is index-aligned with the document.

If collapsing blank lines is deliberate for readability, then an opt-in `preserve_structure: bool = False` parameter on the existing tool preserves both behaviours without adding a tool. My preference would be index-alignment as the default, since the current output is silently lossy rather than merely tidy, but I'll follow whichever you prefer.

Happy to implement — small change in `docs_tools.py` plus a test. Say the word and I'll open a PR.

DrFaust92:
## Description

Fixes #1085.

`get_doc_content` dropped any paragraph that was empty after `.strip()`, and accumulated only `textRun` content. An empty paragraph is `"\n"` and still occupies an index in the document, as do `inlineObjectElement`, `pageBreak`, `columnBreak`, `horizontalRule`, `footnoteReference`, `person` and `richLink`. Offsets computed by searching the returned text therefore drifted from the document's real indices — silently, and cumulatively — so a subsequent `format_text` / `delete_text` / `replace_text` landed on the wrong range.

This preserves empty paragraphs and represents non-`textRun` elements with U+FFFC (OBJECT REPLACEMENT CHARACTER), sized from each element's own `startIndex`/`endIndex` span so the extracted text keeps the same length as the range it came from.

The extraction also moves out of the nested closure inside `get_doc_content` into `docs_helpers.extract_text_from_elements`, so it is unit-testable and the tool file stays thin per CONTRIBUTING.

**Known limitation, called out in the docstring:** table cell text is included but is not index-aligned — the table and row structure occupies indices that are not represented here. Fixing that is a larger change and I left it out deliberately rather than half-doing it. Happy to follow up if you want it.

**Behaviour change worth flagging for review:** callers that treated the output as prose will now see blank lines that were previously swallowed, and U+FFFC where an inline object sits. That is the point of the fix, but it is a visible difference.

## Type of Change
- [ ] Documentation update

## Testing

`tests/gdocs/test_doc_content_index_alignment.py` — 11 tests. The core one asserts the extracted length equals the document's own span, and that `text.index("Bravo") + 1` equals the index `inspect_doc_structure` reports (8) for the repro in #1085.

Local run with the CI-pinned toolchain:

```
uvx ruff@0.15.22 check          -> All checks passed!
uvx ruff@0.15.22 format --check -> 201 files already formatted
uv run pytest                   -> 1897 passed, 2 skipped
```

Manually verified against a real ~10k-character document tab: extracted length now matches the API's reported span exactly, where it was previously short by the count of empty paragraphs.

## Checklist

## Additional Notes

Two related issues are open and I am happy to take either: #1084 (`inspect_doc_structure` truncates `text_preview` at 100 chars, so sub-paragraph offsets cannot be located even once this fix lands) and #1086 (read tools cannot target a tab). Both are parameter additions to existing tools rather than new tools.

## Summary by CodeRabbit

* **Improvements**
  * Improved document text extraction to preserve character positions accurately, including page breaks, inline objects, and other non-text elements.
  * Preserved empty paragraphs and whitespace during content extraction.
  * Added support for extracting text from tables and document tabs, including tab identification.
  * Added safeguards for deeply nested table content.
* **Tests**
  * Expanded coverage for extraction alignment, special elements, tables, tabs, and nested content.

### DIFF
--- a/gdocs/docs_helpers.py
+++ b/gdocs/docs_helpers.py
@@ -1955,3 +1955,88 @@ def validate_operation(operation: Dict[str, Any]) -> tuple[bool, str]:
             )
 
     return True, ""
+
+
+# Unicode OBJECT REPLACEMENT CHARACTER. Stands in for a non-text paragraph
+# element (inline image, page break, footnote reference, ...) so that the
+# extracted text keeps the same length as the document range it came from.
+OBJECT_REPLACEMENT_CHAR = "￼"
+
+# Paragraph elements that are not textRun but still occupy index positions.
+NON_TEXT_PARAGRAPH_ELEMENTS = (
+    "inlineObjectElement",
+    "pageBreak",
+    "columnBreak",
+    "horizontalRule",
+    "footnoteReference",
+    "person",
+    "richLink",
+    "equation",
+    "autoText",
+)
+
+TAB_HEADER_FORMAT = "\n--- TAB: {tab_name} (ID: {tab_id}) ---\n"
+
+
+def _paragraph_element_text(element: Dict[str, Any]) -> str:
+    """Return the text a single paragraph element contributes.
+
+    A textRun contributes its content verbatim. Any other element type
+    contributes OBJECT_REPLACEMENT_CHAR repeated to match the element's own
+    index span, so offsets computed from the result stay aligned with the
+    document. Falls back to a single character when the API omits the span.
+    """
+    text_run = element.get("textRun", {})
+    if text_run and "content" in text_run:
+        return text_run["content"]
+
+    if any(key in element for key in NON_TEXT_PARAGRAPH_ELEMENTS):
+        start = element.get("startIndex")
+        end = element.get("endIndex")
+        if isinstance(start, int) and isinstance(end, int) and end > start:
+            return OBJECT_REPLACEMENT_CHAR * (end - start)
+        return OBJECT_REPLACEMENT_CHAR
+
+    return ""
+
+
+def extract_text_from_elements(
+    elements: List[Dict[str, Any]],
+    tab_name: Optional[str] = None,
+    tab_id: Optional[str] = None,
+    depth: int = 0,
+) -> str:
+    """Extract text from Google Docs structural elements.
+
+    Empty paragraphs are preserved: a paragraph whose only content is a
+    newline still occupies an index in the document, so dropping it shifts
+    every subsequent offset. Non-textRun paragraph elements are represented
+    by OBJECT_REPLACEMENT_CHAR for the same reason.
+
+    Table cell text is included but is not index-aligned; the table and row
+    structure itself occupies indices that are not represented here.
+    """
+    if depth > 5:
+        return ""
+
+    text_lines = []
+    if tab_name:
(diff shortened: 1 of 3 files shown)


taylorwilsdon:
Shoot, I totally missed this when working through https://github.com/taylorwilsdon/google_workspace_mcp/pull/1098 

Appreciate it regardless, will give you a shoutout in the release notes haha

DrFaust92:
> Shoot, I totally missed this when working through #1098
> 
> Appreciate it regardless, will give you a shoutout in the release notes haha

No worries, I have more work to do

PR #1090, opened 2026-09-04
feat(calendar): paginate get_events and list_calendars

### LINKED ISSUE #1089: gcalendar has no pagination: get_events and list_calendars silently truncate
**Describe the bug**

`gcalendar` is the only package in the repo with no pagination. A grep for `pageToken|nextPageToken` finds it in `gtasks`, `gmail`, `gdrive`, `gforms` and `gcontacts`, and in zero files under `gcalendar/`.

Two consequences.

`get_events` makes one `events().list()` call (`gcalendar/calendar_tools.py:507`), reads `items` (`:509`), and discards `nextPageToken`. With `max_results` defaulting to 25, a caller gets 25 events and no signal that a 26th exists. When the result is empty it returns:

```
No events found in calendar 'primary' for <email> for the specified time range.
```

That sentence is stated as fact, and the Calendar API can return an empty page alongside a `nextPageToken`, so it can be false.

`list_calendars` (`:377`) calls `service.calendarList().list()` with no arguments at all: no `maxResults`, no `pageToken`. It then reports `Successfully listed {len(items)} calendars` (`:388`). On an account past the API's default page size, that number is the page, not the total, and nothing says so.

**To Reproduce**

1. Use an account with more than 25 events in a window, or more than one page of calendars.
2. Call `get_events` without `max_results`, or `list_calendars`.
3. The response gives no indication that further pages exist.

**Expected behavior**

The same treatment `gtasks` already gives this. `list_tasks` (`gtasks/tasks_tools.py:131,150-151,156,166-167`) takes a `page_token`, forwards it as `pageToken`, reads `nextPageToken` off the result, and appends `Next page token: …` to the response when there is one.

Applying that to `get_events` and `list_calendars` would make truncation visible without changing what a caller who ignores the parameter sees today.

**Platform**
- Version: 1.25.2 (`main`)
- Not platform-specific; it is in the request construction.

ConnorMoss02:
Closes #1089

## Description

Adds pagination to `get_events` and `list_calendars`. Closes #1089. Defaults are unchanged, so a caller who passes neither parameter sends the same request as before.

`gcalendar` was the only package in the repo without it. A grep for `pageToken|nextPageToken` finds it in `gtasks`, `gmail`, `gdrive`, `gforms` and `gcontacts`, and in zero files under `gcalendar/`.

### Why

`get_events` made one `events().list()` call (`calendar_tools.py:507`) and dropped `nextPageToken`. With `max_results` defaulting to 25, a caller received 25 events and nothing indicating a 26th existed. When the page came back empty it returned:

```
No events found in calendar 'primary' for <email> for the specified time range.
```

The Calendar API can return an empty page alongside a `nextPageToken`, so that sentence could be false.

`list_calendars` called `service.calendarList().list()` with no arguments at all, then reported `Successfully listed {len(items)} calendars`. Past the API's default page size that number is the page, not the total.

## Changes

- `gcalendar/calendar_tools.py`: both tools take `page_token`, forward it as `pageToken`, read `nextPageToken` off the result, and append `Next page token: …` when one is present. `list_calendars` also gains `max_results`.
- An empty page carrying a token now says so instead of claiming there is nothing there.
- `tests/gcalendar/test_pagination.py`: 11 tests.
- `skills/managing-google-workspace/references/calendar.md`: parameter rows and a note on how to continue.

The shape follows `list_tasks` (`gtasks/tasks_tools.py:131`, `:150-151`, `:156`, `:166-167`) rather than inventing one — same parameter name, same forwarding, same `Next page token: …` line on the response.

## Backward compatibility

Additive. Both parameters are optional and omitted by default, so the request built for a caller who ignores them is byte-identical to today's. `test_omits_page_token_when_not_given` and `test_sends_no_paging_arguments_by_default` pin that.

The one changed message is the empty-page-with-more-pages case, which previously reported "No events found" for a range that had events.

## Validation

```bash
uv run pytest tests/gcalendar/
uvx ruff@0.15.22 check && uvx ruff@0.15.22 format --check
```

210 passed. Ruff clean on the pinned version.

Confirmed the tests are load-bearing: with `calendar_tools.py` reverted, 6 of the 11 fail.

## Summary by CodeRabbit

* **New Features**
  * Added pagination support for listing calendars and retrieving events.
  * Added page-size and continuation-token controls.
  * Responses now include continuation details, including the effective start time when needed.

* **Bug Fixes**
  * Improved handling of empty intermediate pages and continuation requests.
  * Expanded event searches now require a start time for continuation, while non-expanded searches may continue without one.

* **Documentation**
  * Updated calendar guidance for pagination, event expansion, continuation requirements, and preserving query parameters.

### DIFF
--- a/gcalendar/calendar_tools.py
+++ b/gcalendar/calendar_tools.py
@@ -449,7 +449,7 @@ async def get_events(
         time_min (Optional[str]): The start of the time range (inclusive) in RFC3339 format (e.g., '2024-05-12T10:00:00Z' or '2024-05-12'). If omitted, defaults to the current time when single_events=True. It is omitted from unexpanded queries so recurring masters that began in the past but still have future occurrences remain discoverable. Ignored if event_id is provided.
         time_max (Optional[str]): The end of the time range (exclusive) in RFC3339 format. If omitted, events starting from `time_min` onwards are considered (up to `max_results`). Ignored if event_id is provided.
         max_results (int): The maximum number of events to return in one page. Defaults to 25. Ignored if event_id is provided.
-        page_token (Optional[str]): Token for the next page, taken from a previous response. Ignored if event_id is provided.
+        page_token (Optional[str]): Token for the next page, taken from a previous response. When single_events=True, also pass the response's Pagination time_min as time_min, even if omitted on the first call. Keep all other query parameters unchanged. Ignored if event_id is provided.
         query (Optional[str]): A keyword to search for within event fields (summary, description, location). Ignored if event_id is provided.
         detailed (bool): Whether to return detailed event information including description, location, colour (colorId), attendees, and attendee details (response status, organizer, optional flags). Recurring instances also report the parent series ID needed to edit the whole series; recurring masters report their raw RFC5545 recurrence rules; and events that are not ordinary confirmed meetings report their event type (outOfOffice, workingLocation, focusTime) and status. Defaults to False.
         include_attachments (bool): Whether to include attachment information in detailed event output. When True, shows attachment details (fileId, fileUrl, mimeType, title) for events that have attachments. Only applies when detailed=True. Set this to True when you need to view or access files that have been attached to calendar events, such as meeting documents, presentations, or other shared files. Defaults to False.
@@ -463,6 +463,7 @@ async def get_events(
     )
 
     next_page_token: Optional[str] = None
+    pagination_info = ""
 
     # Handle single event retrieval
     if event_id:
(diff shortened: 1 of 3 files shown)


ConnorMoss02:
Confirmed and fixed in 1838f25.

`effective_time_min` falls back to `datetime.datetime.now(...)` when `time_min` is omitted and `single_events` is on (`calendar_tools.py:482-484`), so it is recomputed per call. A continuation would have sent a different `timeMin` than the request the token came from, and nothing in the result would have said so.

Rather than silently paging a drifted query, `get_events` now raises before the request when `page_token` is given without `time_min`:

```
[get_events] page_token requires time_min. Pass the same time_min used for the
first page; without it the range starts at the current time, which changes
between calls and does not match the token.
```

Unexpanded queries are unaffected: `single_events=False` omits `timeMin` entirely, so nothing drifts and continuation still works. There is a test for that case so the guard cannot widen to it later.

Three tests added: the rejection (asserting `events().list` is never reached), a continuation with `time_min` forwarding both `pageToken` and `timeMin`, and the unexpanded case. Documented on the `page_token` row and in a note under the table.

213 passed. Ruff clean on 0.15.22. Confirmed load-bearing with the source change reverted.

taylorwilsdon:
Thanks!

PR #1095, opened 2026-09-06
fix(gmail): keep a label's visibility when updating it

### LINKED ISSUE #1094: manage_gmail_label update resets a hidden label to visible
Renaming a hidden Gmail label makes it visible again.

`manage_gmail_label(action="update")` types both visibility fields with defaults (`gmail/gmail_tools.py:3748-3749`):

```python
label_list_visibility: Literal["labelShow", "labelHide"] = "labelShow",
message_list_visibility: Literal["show", "hide"] = "show",
```

`users.labels.update` is a PUT, so the body replaces the label outright. Any update that does not mention visibility sends those defaults, and they overwrite what Gmail has stored.

The update branch already fetches the label at `:3810`, and `name` and `color` both fall back to it. The two visibility fields do not (`:3814-3820`):

```python
label_object = {
    "id": label_id,
    "name": name if name is not None else current_label["name"],   # falls back
    "labelListVisibility": label_list_visibility,                  # parameter default
    "messageListVisibility": message_list_visibility,              # parameter default
}
label_color = None if clear_color else color or current_label.get("color")  # falls back
```

**Repro.** A label stored as `labelListVisibility: labelHide`, `messageListVisibility: hide`:

```
manage_gmail_label(action="update", label_id="Label_1", name="Receipts 2026")

PUT body: {'id': 'Label_1', 'name': 'Receipts 2026',
           'labelListVisibility': 'labelShow', 'messageListVisibility': 'show'}
```

The label reappears in the sidebar and its messages reappear in the message list. The tool returns `Label updated successfully!` with the name and ID, and says nothing about the visibility change, so nothing downstream can tell.

Reachable from any update that changes only the name or only the color.

**Expected.** Omitting a visibility field on update should keep the label's current setting, as `name` and `color` already do.

**Note on `patch`.** `users.labels.patch` exists, but switching to it would break `clear_color`, which relies on PUT semantics where omitting `color` removes it. The fix belongs in how the body is built.

ConnorMoss02:
Closes #1094

`users.labels.update` is a PUT, so the body replaces the label outright. Both visibility fields were typed with defaults, so any update that did not mention them wrote those defaults over what Gmail had stored:

```
stored:   labelListVisibility=labelHide, messageListVisibility=hide
call:     manage_gmail_label(action="update", label_id="Label_1", name="Receipts 2026")
PUT body: {'id': 'Label_1', 'name': 'Receipts 2026',
           'labelListVisibility': 'labelShow', 'messageListVisibility': 'show'}
```

The label returns to the sidebar and its messages return to the message list. The receipt reports the rename and says nothing about the visibility change.

`name` and `color` in that same request body already fall back to the fetched label. Visibility now does too.

### Why not `users.labels.patch`

It exists and would avoid the problem generally, but `clear_color` depends on PUT semantics — omitting `color` is what removes it. Switching would break that, so the fix is in how the body is built.

### Changes

- `build_label_visibility` in `gmail_helpers.py`, beside `build_label_color`.
- Both parameters become `Optional[...] = None`, so "not specified" is distinguishable from "labelShow".
- Create is unchanged and still sends `labelShow`/`show`.
- Docstrings match the wording already used for `text_color`.

### Tests

In `tests/gmail/test_manage_gmail_label_color.py`:

- three stored-visibility combinations preserved across a rename — these fail with the source change reverted
- an explicit visibility change still applies
- a label Gmail returns without visibility fields keeps the previous behaviour
- create sends the same body as before

280 passed in `tests/gmail`. `ruff check` and `ruff format --check` clean on 0.15.22.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updating a Gmail label without specifying visibility settings now preserves its existing visibility.
  - Explicit visibility changes continue to apply correctly.
  - Visibility uses standard defaults when no stored values are available.
  - Creating labels retains the existing default visibility behavior.

- **Documentation**
  - Clarified that visibility defaults apply during label creation, while omitted settings during updates preserve current values.

- **Tests**
  - Added coverage for preserving, overriding, and defaulting label visibility settings.

### DIFF
--- a/gmail/gmail_helpers.py
+++ b/gmail/gmail_helpers.py
@@ -206,6 +206,24 @@ def build_label_color(
     }
 
 
+def build_label_visibility(
+    requested: Optional[str],
+    stored: Optional[str],
+    fallback: str,
+) -> str:
+    """Resolve one visibility field for a label update.
+
+    `users.labels.update` is a PUT, so the body it receives replaces the label
+    outright: any field taken from a parameter default overwrites what Gmail
+    has stored. Visibility is therefore carried over from the fetched label
+    whenever the caller did not ask to change it, which is what `name` and
+    `color` already do in the same request body.
+    """
+    if requested is not None:
+        return requested
+    return stored if stored is not None else fallback
+
+
 def _normalize_email(address: str) -> str:
     """Lowercase an email address and strip plus-addressing so that
     e.g. 'Alex <alex+foo@scopestack.io>' normalizes to 'alex@scopestack.io'.
--- a/gmail/gmail_tools.py
+++ b/gmail/gmail_tools.py
@@ -77,6 +77,7 @@ from gmail.gmail_helpers import (
     _retryable_result_ids,
     _signature_html_to_text,
     build_label_color,
+    build_label_visibility,
     html_to_text_preserving_breaks,
 )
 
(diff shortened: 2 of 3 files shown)


ConnorMoss02:
Still green and conflict-free against main. Closes #1094.

ConnorMoss02:
Done in `aa6d4fc`. The docstring already drew the create/update distinction; the reference table still showed the defaults as unconditional.

taylorwilsdon:
Thanks!

PR #1101, opened 2026-09-08
fix(utils): distinguish an unreadable Office file from an empty one

warnes:
Replaces #1060, which had drifted 114 commits behind `main` and could no longer be merged. Same change, rebuilt on current `main` (v1.26.0). **Please close #1060 in favour of this.**

## The problem

`extract_office_xml_text` returns `None` for two different things: a file that is genuinely empty, and a file that could not be read at all. Callers cannot tell them apart, so `get_drive_file_content` reports a damaged `.docx` as `[Binary or unsupported text encoding]` — which sends the reader looking at encoding rather than at a corrupt file.

Measured on `main` before the change:

| input | result |
|---|---|
| not a zip at all | `None` |
| truncated zip | `None` |
| malformed `word/document.xml` | `None` |
| readable, genuinely no text | `None` |

Four different situations, one answer.

## The change

A new `OfficeXmlExtractionError`. Damaged input **raises**; readable-but-empty still returns `None`. Same four inputs after:

```
not a zip at all          -> RAISED  not a valid Office file
truncated zip             -> RAISED  not a valid Office file
malformed document.xml    -> RAISED  member 'word/document.xml' is not parseable XML
READABLE but no text      -> returned None
```

The two call sites (`gdocs/docs_tools.py`, `gdrive/drive_tools.py`) catch it and report `[Could not read '<mime>' file - it appears damaged…]`.

Deliberately preserved: a file that is readable and simply has no extractable text keeps returning `None`, and a *missing* optional part stays a non-error. Only "present and unreadable" raises. Distinguishing those is the whole point — a change that raised for both would trade one conflation for another.

## What differs from #1060

Beyond the rebase, this covers two paths that version missed, both found by re-deriving against current `main` rather than replaying the old diff:

- `word/document.xml` is now parsed **ahead of** the member loop, so its failure never reached the member-level handler.
- `_word_related_text_targets` returned `[]` on a **malformed** relationships part, silently dropping every header, footer and note — a different door onto the same defect.

## Testing

`2133 passed, 2 skipped` on this branch. `ruff check .` and `ruff format --check .` both clean. 19 new tests in `tests/core/test_office_extraction_errors.py`, covering all three states.

**Negative control:** with all 9 raise sites neutered, **12 of 19 fail** with `DID NOT RAISE`. The 7 that still pass are the readable-but-empty and success-path tests, which must pass either way — so the suite distinguishes "guard works" from "guard absent" rather than merely being green.

## Note on #1059

#1059's work is already on `main` — it landed via #1071. The comment I left on both PRs saying they were not independent and could not merge in either order **is now stale**; it was written when both were open, and there is nothing left to order. This PR stands alone against current `main`.

Scoped to Office/OOXML only. The equivalent PDF conflation is real but is a separate change and not included here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of damaged or invalid Office documents during text extraction.
  - Users now receive a clear message when an Office file cannot be read, instead of an inaccurate unsupported-encoding message.
  - Readable Office files containing no text continue to be handled as empty.
  - Non-Office files, including text, CSV, Markdown, JSON, and PDF files, continue through their normal fallback handling.
  - Existing extraction behavior for valid documents and UTF-8 files remains unchanged.

### DIFF
--- a/core/utils.py
+++ b/core/utils.py
@@ -47,6 +47,20 @@ _OFFICE_RELATIONSHIP_BASES = {
     "https://schemas.openxmlformats.org/officeDocument/2006/relationships",
     "http://purl.oclc.org/ooxml/officeDocument/relationships",
 }
+# The MIME types this extractor understands. Anything else is not an Office
+# document, so its bytes failing to open as a ZIP says nothing about damage.
+#
+# Do not remove this set and let every payload reach the ZIP open below: callers
+# distinguish "raised" from "returned None" to pick a branch, and their fallback
+# decoding lives in the branch an exception skips. Raising for a readable .txt
+# therefore does not merely mislabel it — it discards the content.
+_OFFICE_XML_MIME_TYPES = frozenset(
+    {
+        "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+        "application/vnd.openxmlformats-officedocument.presentationml.presentation",
+    }
+)
 _WORD_TEXT_RELATIONSHIP_KINDS = {"header", "footer", "footnotes", "endnotes"}
 _WORD_TEXT_RELATIONSHIP_TYPES = {
     f"{base}/{kind}": kind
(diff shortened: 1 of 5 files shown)


taylorwilsdon:
PR submitted incorrectly without maintainer edit rights, moved to https://github.com/taylorwilsdon/google_workspace_mcp/pull/1125

PR #1102, opened 2026-09-08
fix(comments): comment tools emit literal \n instead of newlines

warnes:
## The problem

Every return in `core/comments.py` is assembled with `"\\n"` in the source — a backslash followed by the letter `n`, not a newline. The comment tools return **one physical line** with the escape sequence in it.

Observed on current `main` (v1.26.0), `read_doc_comments` with a single comment:

```
Found 1 comments in Google Docs 1abc:\n\nComment ID: c1\nAuthor: Alice\nContent: First note\n...
```

`len(result.splitlines()) == 1`. Whatever renders that output shows the reader `\n` rather than a line break.

## The change

Five sites, not one. Grepping for the `"\\n".join` shape finds only the read path — the three write confirmations and the list header are **f-strings**, so that search does not reach them:

| line | site |
|---|---|
| 265 | read: list header |
| 299 | read: body `.join` |
| 329 | `_create_comment_impl` confirmation |
| 357 | `_reply_to_comment_impl` confirmation |
| 385 | `_resolve_comment_impl` confirmation |

After: the same call returns 6 real lines, 0 literal escapes.

## Testing

`2118 passed, 2 skipped`. `ruff check .` and `ruff format --check .` clean.

`tests/core/test_comments.py:66` split on the literal escape to reach the fields — that only worked *because* the output was one line. It now uses `splitlines()`, which reads the same either way.

New `tests/core/test_comments_line_breaks.py` covers all four entry points. The assertions pin the **property** (`len(out.splitlines()) > 1`) rather than the absence of one spelling of the bug, so they do not need rewriting if the format changes again.

**Negative control:** with the literals restored, **5 of 21 fail**; with the fix, 21 pass. The tests were watched failing before being trusted.

## Note on scope

This is display formatting only — no API call, argument or return type changes. Any caller that was string-matching on the literal `\n` in these outputs would be matching on the defect.

## Summary by CodeRabbit

- **Bug Fixes**
  - Comment tool output now uses actual line breaks instead of displaying literal `\n` sequences.
  - Improved formatting for reading, creating, replying to, and resolving comments.

- **Tests**
  - Added coverage to verify multi-line comment output is rendered correctly.

### DIFF
--- a/core/comments.py
+++ b/core/comments.py
@@ -262,7 +262,7 @@ async def _read_comments_impl(
     if not comments:
         return f"No comments found in {app_name} {file_id}"
 
-    output = [f"Found {len(comments)} comments in {app_name} {file_id}:\\n"]
+    output = [f"Found {len(comments)} comments in {app_name} {file_id}:\n"]
 
     for comment in comments:
         author = comment.get("author", {}).get("displayName", "Unknown")
@@ -296,7 +296,7 @@ async def _read_comments_impl(
 
         output.append("")  # Empty line between comments
 
-    return "\\n".join(output)
+    return "\n".join(output)
 
 
 async def _create_comment_impl(
@@ -326,7 +326,7 @@ async def _create_comment_impl(
     author = comment.get("author", {}).get("displayName", "Unknown")
     created = comment.get("createdTime", "")
 
-    return f"Comment created successfully!\\nComment ID: {comment_id}\\nAuthor: {author}\\nCreated: {created}\\nContent: {comment_content}"
+    return f"Comment created successfully!\nComment ID: {comment_id}\nAuthor: {author}\nCreated: {created}\nContent: {comment_content}"
 
 
 async def _reply_to_comment_impl(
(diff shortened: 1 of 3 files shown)


taylorwilsdon:
Thanks!

taylorwilsdon:
➜  google_workspace_mcp git:(fix/comments-real-newline) ✗ git push
remote: Permission to Warnes-Innovations/google_workspace_mcp.git denied to taylorwilsdon.
fatal: unable to access 'https://github.com/Warnes-Innovations/google_workspace_mcp.git/': The requested URL returned error: 403

taylorwilsdon:
I don't know why it skipped [Check Maintainer Edits Enabled / check-maintainer-edits-internal (pull_request)](https://github.com/taylorwilsdon/google_workspace_mcp/actions/runs/34173065533/job/101896998073?pr=1102)

Spun off as https://github.com/taylorwilsdon/google_workspace_mcp/pull/1124

PR #1105, opened 2026-09-08
Add opt-in OpenClaw HTTP deployment preset with coordinated credential refresh

### LINKED ISSUE #1104: Feature: opt-in OpenClaw managed HTTP deployment
### Problem

OpenClaw's normal stdio MCP lifecycle can create multiple Google Workspace MCP processes across agent sessions. In a local single-user setup, those processes may share the same Google OAuth refresh-token file. Concurrent access-token refreshes�w^~)�tespecially Calendar write operationr��y��y�can rotate or overwrite the refresh token and lead to repeated, disruptive reauthorization prompts.

This was observed with OpenClaw 2026.9.x and google_workspace_mcp v1.26.0. The MCP's token file was persistent and permissioned correctly; the failure mode came from several concurrent stdio server instances sharing it.

### Requested feature

Please add an **opt-in OpenClaw deployment preset** for one durable, localhost-only Streamable HTTP Workspace MCP service with coordinated Google credential refresh.

Possible interface:

```bash
workspace-mcp serve --preset openclaw
```

or:

```bash
WORKSPACE_MCP_DEPLOYMENT_PRESET=openclaw
```

The preset should be optional and leave existing stdio behavior unchanged.

### Proposed behavior

1. **Require Streamable HTTP**
   - Bind to `127.0.0.1` by default.
   - Expose the MCP endpoint at `/mcp`.
   - Document registration in OpenClaw as a shared Streamable HTTP endpoint.
   - Keep the service private by default; reverse-proxy exposure should be an explicit deployment choice.

2. **Use one durable credential identity**
   - Use the existing credential-store abstraction and a dedicated durable credential directory/backend.
   - Key credentials by Google account and effective permission set.
   - Avoid session/request-specific token files.

3. **Coordinate refresh-token operations**
   - Add a per-credential async/process lock around load+�u����R validity check"��y��y� refresh+�u����R persist.
   - Reload credentials after lock acquisition, so a waiting request observes a refresh performed by another request.
   - Use atomic persistence (temporary file, fsync, rename) for filesystem-backed stores.

Illustrative flow:

```python
async with credential_store.lock(account_id):
    credentials = credential_store.load(account_id)
    if credentials.expired and credentials.refresh_token:
        credentials.refresh(request)
        credential_store.atomic_save(account_id, credentials)
```

This belongs naturally in the credential-store/auth layer rather than in individual Gmail, Calendar, Drive, or Sheets tools.

4. **Expose non-secret health status**
   - Service running
   - account authorized
   - refresh capability available
   - enabled Google services/scopes
   - no token material in logs, config, or tool output

5. **Provide OpenClaw documentation**
   - A systemd unit example for one persistent local instance.
   - An OpenClaw Streamable HTTP registration example.
   - One Google consent flow that survives OpenClaw session churn and service restarts.
   - Clear ownership boundary: OpenClaw authorizes agents to the local MCP; Workspace MCP retains and refreshes the Google OAuth credentials.

### Acceptance tests

- Force token expiry, then run 10 concurrent Calendar read/write operations through one HTTP service.
- Verify one refresh request and no reauthorization prompt.
- Restart OpenClaw repeatedly; consent remains valid.
- Restart Workspace MCP; consent remains valid.
- Verify tokens never appear in logs, configuration, or tool output.

### Why this seems feasible

Workspace MCP already documents Streamable HTTP and credential-store modes; OpenClaw supports shared Streamable HTTP MCP endpoints. This looks primarily like a supported lifecycle/deployment preset plus safe refresh coordination�w^~)�tnot a protocol redesign or a request to expose Google tokens to OpenClaw.

Happy to test an implementation against OpenClaw and report results.

kip-claw:
Fixes #1104

Adds `--preset openclaw` / `WORKSPACE_MCP_DEPLOYMENT_PRESET=openclaw`, an opt-in
preset that forces streamable-http transport bound to 127.0.0.1, for running one
durable local MCP service instead of one stdio process per OpenClaw agent
session. Multiple concurrent stdio processes sharing the local OAuth
refresh-token file could otherwise race on refresh and rotate/corrupt it,
forcing repeated reauthorization.

- Per-account `asyncio.Lock` serializes the credential load -> check -> refresh ->
  persist cycle (`auth/credential_store.py`, `auth/google_auth.py`); a caller that
  waited behind a refresh observes the already-refreshed credentials.
- Atomic credential persistence (temp file + fsync + `os.replace`) in
  `LocalDirectoryCredentialStore`.
- New `GET /status` endpoint: running/account-authorized/refresh-capable/enabled
  services & scopes, with no token material ever included.
- `docs/openclaw-deployment.md`: systemd unit + OpenClaw HTTP registration
  example + ownership-boundary notes.
- Default stdio behavior is unchanged unless the preset is explicitly selected.

## Test plan

- `ruff check` / `ruff format --check` clean
- Full pytest suite green aside from 20 pre-existing, unrelated failures in
  `tests/gcontacts/test_contacts_integration.py` (a Python 3.13 event-loop issue,
  identical before and after this change)
- 24 new tests added, all passing: preset resolution and its effect on
  transport/host selection, per-account lock serialization under concurrency
  (10 concurrent callers -> exactly 1 refresh), atomic-save semantics
  (temp file + `os.replace`, no debris on failure, no partial content on
  overwrite), and `/status` field correctness with no token leakage.

## Manual road test against a live Google account

Beyond the unit tests, I installed this branch into an isolated environment on
the machine that filed #1104 (separate venv, a *copy* of the real credentials
directory, port 8765) and exercised it against a real, live Google account —
the production OpenClaw MCP process was never touched.

1. Started with `WORKSPACE_MCP_DEPLOYMENT_PRESET=openclaw`. Confirmed in the
   logs the preset forced `streamable-http` and bound `127.0.0.1`, serving at
   `http://127.0.0.1:8765/mcp`.
2. `GET /status` returned:
   ```json
   {
     "status": "running",
     "transport": "streamable-http",
     "account_authorized": true,
     "refresh_capable": true,
     "enabled_services": ["calendar", "contacts", "docs", "drive", "gmail", "sheets"],
     "enabled_scopes": ["...", "..."]
   }
   ```
   No token or refresh-token material present.
3. Real tool calls worked end-to-end over the HTTP transport, e.g.
   `list_calendars` successfully returned the account's real calendar.
4. **Concurrency / refresh-coordination test (mirrors the issue's acceptance
   test):** hand-edited the stored credential's `expiry` to a date in the past
   to force expiry, restarted the service so it loaded the now-expired
   credential fresh, then fired 10 concurrent `list_calendars` calls at once
   via `asyncio.gather` against the single running HTTP service.
   - **10/10 calls succeeded.**
   - The credential file's `expiry` was updated to a new valid future
     timestamp (proving a refresh actually occurred).
   - The credential JSON remained fully valid and readable afterward (no
     partial/corrupt content from concurrent writers).
   - No reauthorization prompt was triggered at any point.

This is the exact failure mode described in #1104 (concurrent processes
racing on a shared refresh token) reproduced and shown fixed under this
change, on top of the unit-level lock/atomic-save tests.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in OpenClaw deployment preset for durable local Streamable HTTP deployments.
  - Added a `/status` endpoint showing non-sensitive deployment, authorization, service, and scope information.

- **Bug Fixes**
  - Improved credential handling to prevent concurrent refresh conflicts for the same account.
  - Credential updates are now saved atomically and remain durable across crashes.
  - Status reporting avoids exposing tokens or other sensitive credential data.
  - Single-user mode now consistently uses the configured account.

- **Documentation**
  - Added guidance for configuring the OpenClaw preset, secure deployment, and the health endpoint.


kip-claw:
@coderabbitai Both actionable findings addressed in acc6117:

- `auth/credential_store.py`: `store_credential()` now fsyncs the containing directory after `os.replace()`, so the rename itself is durable across a crash immediately following it — not just the temp file's own contents.
- `docs/openclaw-deployment.md`: the systemd example no longer puts `GOOGLE_OAUTH_CLIENT_SECRET` in `Environment=` (readable via `systemctl show`/`/proc/<pid>/environ`). It now references a permission-restricted `EnvironmentFile` instead, with a note on why and how to lock it down.

Full test suite and `ruff check`/`ruff format --check` (against the pinned 0.15.22 used in CI) are clean.

PR #1106, opened 2026-09-08
Add draft deletion to draft_gmail_message

guidupuy:
Gmail drafts can be created through `draft_gmail_message`, but they cannot currently be discarded through the tool surface. Trying to delete one with message-label tools fails because Gmail does not allow labels on draft messages.

This extends the existing tool, following CONTRIBUTING.md's preference against adding tools:

- `action="create"` remains the default, preserving existing calls, including explicitly empty subjects and bodies.
- `action="delete", draft_id="…"` permanently deletes the specified draft through `users.drafts.delete`; subject and body can be omitted. The identifier is the Draft ID returned by creation, not a message or thread ID.
- The deletion API call lives in `gmail_helpers.py`. Missing IDs and API failures cannot return a success message. Creation rejects a supplied `draft_id` so it cannot silently create a replacement when deletion was intended.
- The existing `gmail.compose` scope suffices; no tool, permission tier, dependency, or OAuth scope is added. The tool's destructive annotation is updated to reflect deletion.

Example:

```python
draft_gmail_message(
    user_google_email="user@example.com",
    action="delete",
    draft_id="r123456789",
)
```

Validation: 101 targeted tests passed across `test_delete_gmail_draft.py`, `test_draft_gmail_message.py`, and `test_permissions.py`. Ruff lint and format checks pass for the changed files. Comprehensive pytest validation is left to CI. No live mailbox mutations were performed.

API reference: https://developers.google.com/workspace/gmail/api/reference/rest/v1/users.drafts/delete

## Summary by CodeRabbit

* **New Features**
  * Gmail draft management now supports permanently deleting drafts in addition to creating them.
  * Draft deletion requires a valid draft ID and clearly indicates its destructive nature.
  * Draft creation validates required subject and body content and prevents conflicting inputs.

* **Bug Fixes**
  * Gmail API errors during draft deletion are surfaced without reporting a false success.

* **Tests**
  * Added coverage for successful deletion, invalid inputs, API errors, and tool discoverability.


ConnorMoss02:
Checked the surrounding surface: `drafts().create()` at `gmail_tools.py:3363` is the only drafts call in the tool today, so this adds the second of Gmail's six. Worth noting the ID problem that creates — your error text says to use the ID returned by creation, and since there's no drafts list or get, that's the only place it can come from, so delete only works if the caller still has the ID from the same session.

PR #1109, opened 2026-09-09
fix(docs): stop inspect_doc_structure 400ing on every call

### LINKED ISSUE #1108: inspect_doc_structure always 400s: field mask mixes legacy body/headers/footers with tabs(...) under includeTabsContent=True
## Summary

`inspect_doc_structure` fails on every call with:

```
HttpError 400 when requesting https://docs.googleapis.com/v1/documents/{id}?includeTabsContent=true&fields=...&alt=json
returned "Field mask may not contain legacy text-level Document resource fields while requesting tabs content".
```

This reproduces on any Google Doc (in practice, virtually all Docs now, since Google auto-migrated every document to the tabs model — including single-tab docs with just a default "Tab 1").

## Version tested

- Commit: `main` as of 2026-09-09 (same field mask present in the current HEAD)

## Root cause

In `gdocs/docs_tools.py`:

```python
_STRUCTURE_FIELDS = (
    f"title,documentStyle,namedRanges,headers,footers,body({_STRUCTURE_CONTENT_FIELDS}),"
    f"tabs(tabProperties,childTabs,documentTab("
    f"documentStyle,namedRanges,headers,footers,body({_STRUCTURE_CONTENT_FIELDS})))"
)
```

`inspect_doc_structure` then calls:

```python
doc = await asyncio.to_thread(
    service.documents()
    .get(
        documentId=document_id,
        includeTabsContent=True,
        fields=_STRUCTURE_FIELDS,
    )
    .execute
)
```

unconditionally — regardless of `tab_id`, `detailed`, or `preview_chars`. `_STRUCTURE_FIELDS` requests both the legacy top-level `body`/`headers`/`footers` fields *and* `tabs(...documentTab(body,headers,footers)...)` in the same field mask. Once `includeTabsContent=True` is set, Google's Docs API rejects any field mask that still references the legacy top-level fields alongside `tabs(...)` — hence the 400 on literally every call.

I confirmed this isn't caller misuse: calling `inspect_doc_structure` with different combinations of `detailed` (true/false), `tab_id` (set/unset), and `preview_chars` produced a byte-for-byte identical outgoing request URL every time. Nothing a caller passes can avoid the bug.

## Reproduction

```json
{
  "document_id": "<any Google Doc ID>",
  "detailed": true
}
```

Result: the 400 above, every time, for every document.

## Relation to prior issues

This looks like a regression introduced while fixing #975 ("inspect_doc_structure returns zero elements for tabbed Google Docs without tab_id"). That fix added `tabs(...)` to the field mask and started passing `includeTabsContent=True` unconditionally, but didn't drop the legacy top-level fields (`headers,footers,body(...)`) that conflict with it.

#1052 ("inspect_doc_structure costs the same as reading the whole document") independently suggested: *"Pass `includeTabsContent=True` only when `tab_id` is set."* That suggestion wasn't implemented, but it would also fix this bug as a side effect.

## Suggested fix

Branch the request on whether `tab_id` is provided:

- No `tab_id`: omit `includeTabsContent` and use only the legacy top-level fields (`title,documentStyle,namedRanges,headers,footers,body(...)`) — this also addresses #1052's cost complaint.
- `tab_id` given: pass `includeTabsContent=True` and use only the `tabs(...)` branch of the field mask, dropping the legacy top-level `body`/`headers`/`footers` from the mask entirely.

## Impact

`inspect_doc_structure` is documented as the primary tool for finding safe insertion indices before `batch_update_doc`/`modify_doc_text` calls (table creation, formatting, header/footer segment edits). With it completely non-functional, there's no supported way to get real document indices — callers are forced to guess offsets and rely on trial-and-error against `insertText`'s bounds validation, which is exactly the workflow the tool exists to prevent.

mcarans:
## Summary

- Fixes #1108: `inspect_doc_structure` failed on every call with `HttpError 400 ... "Field mask may not contain legacy text-level Document resource fields while requesting tabs content"`.
- Root cause: `_STRUCTURE_FIELDS` requested the legacy top-level `body`/`headers`/`footers` fields alongside `tabs(...)` while the request always sets `includeTabsContent=True`. Google's Docs API rejects that combination outright, regardless of `tab_id`/`detailed`/`preview_chars` — there was no way for a caller to avoid it.
- Per [Google's tabs guide](https://developers.google.com/workspace/docs/api/how-tos/tabs), when `includeTabsContent=true` the top-level `body`/`headers`/`footers` are left **empty** anyway, so dropping them from the top-level portion of the mask loses no data. The function already has fallback logic that sources `body`/`headers`/`footers`/`documentStyle` from the requested tab, or the first tab when none is specified — that logic is unaffected by this change since it never actually depended on the top-level values (they were always empty once `includeTabsContent=True` was set, which is exactly why the first-tab fallback was added for #975 in the first place).
- Kept `title`, `documentStyle`, `namedRanges` at the top level, since those are document-level metadata rather than "text-level" content and aren't implicated by the API's rejection.

## Test plan

- [x] Added `test_inspect_doc_structure_field_mask_omits_legacy_top_level_content` in `tests/gdocs/test_read_tab_selection.py`, which asserts the field mask never names `body(`/`headers`/`footers` outside the `tabs(...)` branch. Verified it fails against the pre-fix code (reproducing the exact field mask from the bug report) and passes with the fix.
- [x] `pytest tests/gdocs/` — 393 passed (up from 392 with the new test).
- [x] `ruff check gdocs/docs_tools.py tests/gdocs/test_read_tab_selection.py` — clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved document structure inspection for tabbed documents by retrieving tab content through the current tab-based format.
  - Preserved access to document titles, headers, footers, body content, and named ranges within individual tabs.
  - Reduced unnecessary metadata retrieval after batch operations, improving reliability and efficiency.
  - Removed reliance on legacy top-level content fields when inspecting documents.

- **Tests**
  - Added regression coverage for tab selection, field requests, and per-tab metadata handling.


kryan007 (commented):
Reviewed, and reproduced the bug this fixes against a production deployment of 1.26.0 before we rolled back.

The change looks right to me. With `includeTabsContent=True` the API rejects the mask outright — that part we saw live — and per the Docs API tabs guide the legacy top-level `body`/`headers`/`footers` are left empty in that mode anyway, so dropping them from `_STRUCTURE_FIELDS` loses nothing. Keeping `title`/`documentStyle`/`namedRanges` is safe: those are document-level rather than text-level fields, which is why they are allowed alongside `tabs(...)`. The regression test pins the actual invariant — no legacy content fields at the top level of the mask — rather than the current string, which is the part that will keep this from coming back.

One thing worth folding in here or filing separately: this fixes `inspect_doc_structure`, but the post-batch read in `gdocs/managers/batch_operation_manager.py` (around line 167) has the same mask shape under `includeTabsContent=True` — `fields="revisionId,tabs,body,headers,footers,footnotes"`. It is wrapped in `except Exception` that sets `document_length = None`, so it fails silently rather than 400ing; details and evidence are in #1108.

`inspect_doc_structure` is unusable on every document on 1.26.0, so a patch release once this lands would be very welcome — happy to verify both call sites on our side as soon as it ships.

mcarans:
@kryan007 I'm unclear from the review if I'm blocking anything. Please let me know what is the next step (if any).

axsaucedo:
This is an issue on our side, would be great if this can get merged. Thanks for the PR!

taylorwilsdon:
Will get this released today, thanks!

axsaucedo:
This fixed it - thanks!

PR #1110, opened 2026-09-09
fix(drive): return the file description in search and list results

vaultastorage:
## Description

**Symptom.** There is no way to read a Drive file's `description` through this
server. `search_drive_files` and `list_drive_items` never show it, and no tool
parameter turns it on. A caller can *set* a description via `update_drive_file`,
but the moment that call ends the value is unreadable again — the only way to see
it back is to call `update_drive_file` on the same file and read the old-vs-new
diff it prints, which means writing to a file just to read one field.

**Cause.** `build_drive_list_params` in `gdrive/drive_helpers.py` hardcodes the
Drive `files.list` field projection, and `description` is in neither branch:

```python
if detailed:
    fields = ("nextPageToken, files(id, name, mimeType, webViewLink, iconLink,"
              " modifiedTime, createdTime, size, driveId,"
              " lastModifyingUser(displayName, emailAddress)"
              f"{permission_fields})")
else:
    fields = "nextPageToken, files(id, name, mimeType)"
```

Drive returns only the fields you ask for, so the data never reaches the
formatters. Repo-wide, `description` appears in a Drive field projection exactly
once: in `update_drive_file`, the write path.

**Change.**

1. `gdrive/drive_helpers.py` — add `description` to the **detailed branch only**.
2. `gdrive/drive_tools.py` — render it in the `search_drive_files` and
   `list_drive_items` formatters, guarded (see below).
3. Docstrings for both tools and the helper updated, since those are the tool
   contract clients read.

**Why the minimal branch was left alone.** `detailed=False` exists to make cheap
listings cheap. Adding the field there would let one long description dominate a
compact listing, so that branch stays `id, name, mimeType`. Callers who want
descriptions use `detailed=True`, which is the default.

### The description is untrusted input, and is treated as such

A file description is free text that **anyone with edit rights on the file can
set**, and it is a field almost nobody looks at. Before this change it never
reached a model through this server; now it does. Two guards, both covered by
tests:

- **Control characters are flattened.** The output format is one line per file,
  so a raw newline in a description could otherwise forge an entire additional
  result row. `\r`, `\n` and `\t` become spaces.
- **Length is capped** at `DESCRIPTION_RENDER_LIMIT` (1000 chars), and truncation
  is explicit: `... [truncated, N chars total]`. Drive's discovery document
  defines `description` as a plain string with **no documented maximum**, and
  `page_size` has no upper bound, so an uncapped field is an unbounded response.
  Silent truncation would be worse than either — a caller could not tell — hence
  the explicit marker.

Empty, whitespace-only, and absent descriptions produce no `Description:`
segment at all, so output for those files is byte-identical to before.

This does not make the row format injection-proof — `name` is already
interpolated raw into the same line, so a filename can already do the same thing.
It does mean this change adds no new avenue.

No new tools, no new dependencies, no signature changes. The field is a normal
part of the `files.list` response and needs no extra scope or API call.

## Type of Change

## Testing

Nine assertions in `tests/gdrive/test_drive_tools.py`, extending the existing
`build_drive_list_params` block and adding a formatter section:

| Test | Checks |
| --- | --- |
| `test_build_params_detailed_true_includes_extra_fields` (extended) | `detailed=True` requests `description` |
| `test_build_params_detailed_false_omits_extra_fields` (extended) | `detailed=False` still does not |
| `test_search_drive_files_includes_description_when_set` | reaches the output |
| `test_search_drive_files_omits_description_when_absent` | no empty segment |
| `test_list_drive_items_includes_description_when_set` | reaches the output |
| `test_list_drive_items_omits_description_when_absent` | no empty segment |
| `test_search_drive_files_flattens_newlines_in_description` | a newline cannot forge a second row |
| `test_search_drive_files_truncates_overlong_description` | 5000 chars capped at 1000, marker present |
| `test_search_drive_files_omits_whitespace_only_description` | `"   "` renders nothing |

Locally, on `main` at v1.26.0:

- `uv run pytest` — 2121 passed, 2 skipped
- `uvx ruff@0.15.22 check` — all checks passed
- `uvx ruff@0.15.22 format --check` — all files formatted

With the source change reverted and the tests kept, the suite fails — the tests
depend on the fix rather than passing either way.

## Checklist

- [ ] **I have enabled "Allow edits from maintainers" for this pull request**

## Note for the reviewer

`description_str` is built inline in the tool functions rather than in
`*_helpers.py`. That is deliberate: `size_str`, `created_str`,
`last_edited_by_str`, `drive_id_str` and `anyone_role_str` are all already built
inline at that same spot, so this matches its neighbors. Happy to move all of
them into a helper if you would prefer that as a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Detailed Google Drive search and listing results now include file descriptions.
  * Descriptions are formatted for readability, with line breaks normalized and long text truncated with an indicator.

* **Documentation**
  * Updated Drive tool documentation to describe the available description field and detailed-result behavior.

### DIFF
--- a/gdrive/drive_tools.py
+++ b/gdrive/drive_tools.py
@@ -73,6 +73,11 @@ from gdrive.drive_helpers import (
 
 logger = logging.getLogger(__name__)
 
+# Cap for a rendered Drive file description. The field is free text set by
+# anyone who can edit the file, and Drive documents no maximum length, so an
+# uncapped value could dominate a page of up to 1000 results.
+DESCRIPTION_RENDER_LIMIT = 1000
+
 SHARED_DRIVE_ORGANIZER_CONCURRENCY_LIMIT = 10
 
 IMPORT_FORMATS_BY_GOOGLE_MIME_TYPE = {
@@ -206,7 +211,7 @@ async def search_drive_files(
                                    'presentation'/'slides', 'form', 'drawing', 'pdf', 'shortcut',
                                    'script', 'site', 'jam'/'jamboard') or any raw MIME type
                                    string (e.g. 'application/pdf'). Defaults to None (all types).
-        detailed (bool): Whether to include size, modified time, and link in results. Defaults to True.
+        detailed (bool): Whether to include size, modified time, description and link in results. Defaults to True.
         order_by (Optional[str]): Sort order. Comma-separated list of sort keys with optional 'desc' modifier.
                                   Valid keys: 'createdTime', 'folder', 'modifiedByMeTime', 'modifiedTime',
                                   'name', 'name_natural', 'quotaBytesUsed', 'recency', 'sharedWithMeTime',
@@ -217,7 +222,7 @@ async def search_drive_files(
                                 contains its own `trashed` clause (`=` or `!=`), which always wins.
 
     Returns:
-        str: A formatted list of found files/folders with their details (ID, name, type, and optionally size, modified time, link).
+        str: A formatted list of found files/folders with their details (ID, name, type, and optionally size, modified time, description, link).
              Includes a nextPageToken line when more results are available.
     """
     logger.info(
@@ -310,10 +315,26 @@ async def search_drive_files(
             # owns the file.  True creator attribution requires fetching revision 1
             # via files/{id}/revisions and reading its lastModifyingUser.  That adds
             # one API call per file and should be a separate follow-up.
+            _desc = (item.get("description") or "").strip()
+            if _desc:
+                # A description is untrusted text that any editor of the file
+                # can set, and it is rendered into a line-per-file listing.
+                # Newlines and tabs are flattened so it cannot forge an extra
+                # row, and the length is capped so one file cannot dominate a
+                # 100-file page. Drive documents no maximum for this field.
+                _desc = _desc.replace("\r", " ").replace("\n", " ").replace("\t", " ")
+                if len(_desc) > DESCRIPTION_RENDER_LIMIT:
+                    _desc = (
+                        _desc[:DESCRIPTION_RENDER_LIMIT]
+                        + f"... [truncated, {len(_desc)} chars total]"
+                    )
+                description_str = f", Description: {_desc}"
+            else:
+                description_str = ""
             formatted_files_text_parts.append(
                 f'- Name: "{item["name"]}" (ID: {item["id"]}, Type: {item["mimeType"]}{size_str}'
                 f"{created_str}, Modified: {item.get('modifiedTime', 'N/A')}"
-                f"{last_edited_by_str}{anyone_role_str})"
+                f"{last_edited_by_str}{anyone_role_str}{description_str})"
                 f" Link: {item.get('webViewLink', '#')}"
             )
         else:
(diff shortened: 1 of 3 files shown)


gdrive/drive_tools.py
```diff
@@ -310,10 +315,26 @@ async def search_drive_files(
...
+                # Newlines and tabs are flattened so it cannot forge an extra
+                # row, and the length is capped so one file cannot dominate a
+                # 100-file page. Drive documents no maximum for this field.
+                _desc = _desc.replace("\r", " ").replace("\n", " ").replace("\t", " ")
+                if len(_desc) > DESCRIPTION_RENDER_LIMIT:
+                    _desc = (
+                        _desc[:DESCRIPTION_RENDER_LIMIT]
+                        + f"... [truncated, {len(_desc)} chars total]"
+                    )
+                description_str = f", Description: {_desc}"
```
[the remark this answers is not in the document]
vaultastorage: Fair point, and worth being precise about what the sanitizing does and does not claim.

Flattening CR/LF/tab and capping length address **structural** forgery: without them a description can emit what reads as an additional result row in a line-per-file listing. They are not claimed to stop instruction-like text from reaching a consuming agent, and the PR body says as much.

On the suggested remedy, two things:

1. `item["name"]` is already interpolated raw into the same line, and has been all along. Anyone who can rename a shared file can already place arbitrary text into these tool results. This change widens an existing surface rather than opening a new one, which is why I did not treat it as a blocker for returning a standard metadata field.
2. Omitting descriptions from default results returns both tools to today's behavior, where the field cannot be read at all through this server — which is the bug being fixed.

If you would prefer belt-and-braces, I am happy to do either: label the segment explicitly as user-supplied text, or put it behind an opt-in parameter. Both are small. I did not want to add API surface unprompted, and an opt-in field that defaults to off tends to read as "absent" to callers who do not know it exists — the same way the missing field reads today.

Also pushed a commit for the second comment: `list_drive_items` does return early for `resource_type="shared_drives"`, so the docstring now scopes descriptions to `resource_type="items"`. Good catch.

gdrive/drive_tools.py
```diff
@@ -310,10 +315,26 @@ async def search_drive_files(
             # owns the file.  True creator attribution requires fetching revision 1
             # via files/{id}/revisions and reading its lastModifyingUser.  That adds
             # one API call per file and should be a separate follow-up.
+            _desc = (item.get("description") or "").strip()
+            if _desc:
+                # A description is untrusted text that any editor of the file
+                # can set, and it is rendered into a line-per-file listing.
+                # Newlines and tabs are flattened so it cannot forge an extra
+                # row, and the length is capped so one file cannot dominate a
+                # 100-file page. Drive documents no maximum for this field.
```
taylorwilsdon, line 324: Unnecessary

gdrive/drive_tools.py
```diff
@@ -780,10 +805,26 @@ async def list_drive_items(
                     last_edited_by_str = ""
             else:
                 last_edited_by_str = ""
+            _desc = (item.get("description") or "").strip()
+            if _desc:
+                # A description is untrusted text that any editor of the file
+                # can set, and it is rendered into a line-per-file listing.
+                # Newlines and tabs are flattened so it cannot forge an extra
+                # row, and the length is capped so one file cannot dominate a
+                # 100-file page. Drive documents no maximum for this field.
```
taylorwilsdon, line 814: Please limit comments to useful / actionable info.

taylorwilsdon (changes requested):
Thanks! Few cleanup items

PR #1114, opened 2026-09-11
test: contain OAuthConfig's environment writes within the test that makes them

warnes:
Split out of #1103, where it did not belong — `CONTRIBUTING.md` asks for one change per PR, and a suite-root autouse fixture is not part of "support multiple OAuth clients". It stands alone: the leak predates that feature and needs none of its context to review.

## The problem

`OAuthConfig._apply_fastmcp_google_env` writes `FASTMCP_SERVER_AUTH*` into `os.environ` when they are absent and a `client_id` is configured. `monkeypatch` restores only what it set itself, so those writes outlive the test that caused them.

**Nothing fails as a result** — the leak is silent, which is why it lasted. Measured on `main` with credentials absent (the CI condition), running `tests/auth/` and `tests/core/`:

```
573 passed, 0 failed
left behind: FASTMCP_SERVER_AUTH, ..._GOOGLE_BASE_URL, ..._GOOGLE_CLIENT_ID,
             ..._GOOGLE_CLIENT_SECRET, ..._GOOGLE_REDIRECT_PATH
```

Every test after the first one to build a configured `OAuthConfig` runs against configuration it did not set.

## Two paths reach that write, and they mask each other

Worth stating, because it makes this easy to measure wrongly in both directions:

| path | fires when | reachable by a fixture |
|---|---|---|
| **import time** — `core/config.py` builds the config at module level | a credential is already resolvable (exported var, or the default repo-root `client_secret.json`) | **no** — it happens during collection |
| **test time** — a test sets `GOOGLE_OAUTH_CLIENT_ID` and builds a config | no ambient credential, so the names are still absent | yes |

On a developer machine with a credential present, the first fires during collection and every later `_set_if_absent` is a no-op — so measuring there shows a fixture apparently doing nothing. CI has no credentials, so the second is what runs.

**This PR contains only the second.** The first needs the constructor to stop mutating process state, which is a production change and deliberately not attempted here.

## The fixture

Autouse, at the suite root, holding its own `MonkeyPatch`. Each choice is load-bearing:

- **autouse** — opt-in is what let this spread; only three tests had asked for the protection while four modules leaked.
- **suite root** — the leak follows `OAuthConfig`, not a directory. Scoped to `tests/auth/`, that directory reports clean while `tests/core/test_http_oauth_scope_config.py` still leaks.
- **its own `MonkeyPatch`** — `tests/auth/conftest.py`'s `isolated_gateway_config` calls `monkeypatch.undo()` in teardown then rebuilds the singleton. Sharing one releases the pins first and lets the rebuild write the credentials back, after every in-body assertion has passed.

## Testing

`2123 passed, 2 skipped`. `ruff check .` and `ruff format --check .` clean.

Every property is pinned by a mutation that fails a **different** test, because one mutation cannot distinguish them:

| mutation | tests failing |
|---|---|
| `autouse=False` | 8 |
| drop a name from the pinned tuple | 3 |
| demote the fixture to `tests/auth/` | 1 (genuine placement detection) |
| restore the shared `MonkeyPatch` | 1 |
| **delete the production write path entirely** | 1 |
| **leak under a name nobody enumerated** | 1 |

The last two were invisible to an earlier version of these tests. A positive control now proves the write was *attempted*, so "contained" and "never ran" stop producing the same observation; and the session probe diffs the **whole environment** rather than filtering by a known prefix, so a write under an unlisted name cannot hide.

That diff also records what this PR does **not** fix: six other names leak from the same class in `auth/google_auth.py` and `auth/port_resolver.py`, three of which `OAuthConfig` reads back (`WORKSPACE_MCP_RESOLVED_PORT` feeds `base_url`). They are listed with their writers in `KNOWN_UNFIXED_LEAKS` rather than filtered out, so the condition is reviewable instead of invisible.

## Cost and safety notes

The two subprocess probes add about 14s (the session probe is 12.2s by `--durations`), on a suite that runs in 70–95s here. For calibration, `tests/test_log_hygiene.py` already has a 13.9s subprocess test, so this is in line with existing practice rather than an outlier.

Assertions compare a boolean, never the value, and no child-process output is spliced into an assertion message — a guard over a credential-bearing name that renders its operands publishes the secret at exactly the moment it stops being protected.

Nothing reads the five pinned names: not this repo (only `..._JWT_SIGNING_KEY` is read, which is deliberately not pinned), and not `fastmcp` 3.4.7. Related, and filed separately: #1112 stops the client **secret** being written at all.

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive coverage for authentication-related environment variable isolation.
  - Verified environment settings are applied before tests, restored afterward, and remain protected when other fixtures modify the environment.
  - Added checks to detect unexpected environment leaks in subprocess-based test sessions.
- **Chores**
  - Added an automatic test fixture that pins authentication environment variables and cleans them up reliably after each test.

### DIFF
--- a/tests/auth/test_fastmcp_env_isolation.py
+++ b/tests/auth/test_fastmcp_env_isolation.py
@@ -1,0 +1,336 @@
+"""OAuthConfig's environment writes must not outlive the test that makes them.
+
+`OAuthConfig._apply_fastmcp_google_env` writes FASTMCP_SERVER_AUTH* into
+`os.environ` when they are absent and a client_id is configured. `monkeypatch`
+restores only what it set itself, so those writes survive the test that caused
+them, and every later test runs against configuration it did not set.
+
+Nothing fails as a result today -- the leak is silent, which is why it lasted.
+
+TWO PATHS REACH THAT WRITE, and they mask each other, so measuring this is easy
+to get wrong in both directions:
+
+  1. IMPORT TIME. `core/config.py` evaluates `is_oauth21_enabled()` at module
+     level, building the memoised config during collection. It fires only when
+     a credential is already resolvable -- an exported GOOGLE_OAUTH_CLIENT_ID,
+     or the default repo-root client_secret.json. No FIXTURE can reach it,
+     because none exists yet. (A module-level statement in tests/conftest.py
+     could; that is deliberately not done here, since it would hide the
+     production defect rather than contain a test-suite one.)
+
+  2. TEST TIME. A test sets GOOGLE_OAUTH_CLIENT_ID and builds an OAuthConfig.
+     It fires when path 1 did NOT, because the names are then still absent.
+
+CI has no credentials, so CI runs path 2, and path 2 is what the fixture
+contains. On a developer machine with a credential present, path 1 fires first
+and every later `_set_if_absent` is a no-op -- so measuring there shows a
+fixture apparently doing nothing. The probes below therefore neutralise BOTH
+credential sources in the child environment, variables and file alike.
+"""
+
+import os
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+from auth.oauth_config import OAuthConfig
+from tests.conftest import _FASTMCP_ENV_VARS
+
+PINNED = "pinned-by-test"
+REPO_ROOT = Path(__file__).resolve().parents[2]
+
+# Environment names this suite is KNOWN to leak, that this change does not fix.
+# Listed rather than filtered out, so the condition is recorded and reviewable
+# instead of invisible: a probe that matched only FASTMCP_* names would be
+# structurally unable to see any of these, or any future leak under a name
+# nobody has enumerated yet.
+#
+#   OAUTHLIB_INSECURE_TRANSPORT   auth/google_auth.py:705,900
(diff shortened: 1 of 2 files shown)


taylorwilsdon:
Missing maintainer perms - ➜  google_workspace_mcp git:(fix/contain-oauthconfig-env-writes) ✗ git push
remote: Permission to Warnes-Innovations/google_workspace_mcp.git denied to taylorwilsdon.
fatal: unable to access 'https://github.com/Warnes-Innovations/google_workspace_mcp.git/': The requested URL returned error: 403

Spun out as https://github.com/taylorwilsdon/google_workspace_mcp/pull/1137

taylorwilsdon:
Hey Gregory, thanks for digging into this - you were right that the leak is real, but when I went to verify it turned out nothing actually reads those FASTMCP_SERVER_AUTH* vars anymore (fastmcp 3.x doesn't touch them and we pass everything to GoogleProvider directly in core/server.py) so rather than add a fixture to contain the writes I just deleted the write path entirely in #1137, plus a small test that building an OAuthConfig doesn't add anything to os.environ. The OAUTHLIB_* and port resolver leaks you called out are still there and definitely worth their own PR if you're up for it!

PR #1119, opened 2026-09-11
fix: run_script_function must use a deployment ID, not the project ID

### LINKED ISSUE #1118: run_script_function fails for every call: passes project scriptId instead of deployment ID to scripts.run
## Bug

`run_script_function` (and the underlying `_run_script_function_impl` in `gappsscript/apps_script_tools.py`) fails for every call, on every Apps Script project we tested, with either:

```
Execution failed
Function: <fn>
Error: We're sorry, a server error occurred while reading from storage. Error code NOT_FOUND.
```

or, on one project, a clearer:

```
HttpError 403 when requesting https://script.googleapis.com/v1/scripts/{scriptId}:run?alt=json returned "The caller does not have permission"
```

## Root cause

`apps_script_tools.py:512`:

```python
response = await asyncio.to_thread(
    service.scripts().run(scriptId=script_id, body=request_body).execute
)
```

`script_id` here is the Apps Script **project ID** (what `list_script_projects` / the editor URL show). But Google's Apps Script API `scripts.run` method requires that path parameter to be the **API Executable deployment ID** (the `AKfycb...` string returned by `list_deployments` / `create_deployment`), not the bare project ID. Google's own reference is explicit about this: https://developers.google.com/apps-script/api/reference/rest/v1/scripts/run

Passing the project ID instead means Google can't resolve the target as a deployment, hence the generic `NOT_FOUND` (most cases) or a permission-flavored 403 depending on how far resolution gets before failing.

## What we ruled out before finding this (in case useful to others hitting the same symptom)

- The personal "Google Apps Script API" toggle at `script.google.com/home/usersettings` — confirmed already ON.
- Stale/broken deployment — recreated a fresh version + deployment, identical error persisted.
- OAuth scope — re-ran the full OAuth consent flow; the consent screen already requested every Apps Script scope (`script.projects`, `script.scriptapp`, `script.processes`, `script.deployments`, `script.external_request`, etc.). Re-authenticated, identical errors persisted on every project.
- Time-based triggers on the exact same scripts run and complete successfully on schedule throughout — this is specific to the Execution API path, not the scripts themselves or their permissions in general.

## Suggested fix

`run_script_function` should resolve (or require/accept) an API-Executable deployment ID before calling `.run()` — e.g. look one up via `list_deployments` for the given `script_id` (using the most recent API Executable deployment), or add an explicit `deployment_id` parameter to the tool so callers who already have one can pass it directly instead of the project ID.

## Environment

- `workspace-mcp` 1.26.0 (also reproduced on the version installed via `uvx workspace-mcp`)
- Tested against 3 separate Apps Script projects, multiple functions each, consistent failure every time via the Execution API path; all 3 projects' own time-based triggers work fine independently.

ptulin:
Fixes #1118.

## Bug

`run_script_function` fails for every call, on every Apps Script project, with either:

```
Execution failed
Error: We're sorry, a server error occurred while reading from storage. Error code NOT_FOUND.
```

or, depending on how far resolution gets:

```
HttpError 403 ... "The caller does not have permission"
```

## Root cause

`_run_script_function_impl` calls `service.scripts().run(scriptId=script_id, ...)` using the bare Apps Script **project ID**. Google's `scripts.run` reference (https://developers.google.com/apps-script/api/reference/rest/v1/scripts/run) is explicit that this path parameter must actually be the ID of an **API Executable deployment**, not the project ID — the client library's `scriptId` kwarg name is just confusingly reused for that value.

Confirmed this is specific to the Execution API path, not a permissions problem with the scripts themselves: the exact same scripts' own time-based triggers ran successfully throughout testing.

## Fix

- When `deployment_id` isn't supplied, look one up automatically: list deployments for the script, filter out HEAD-only entries (`deploymentConfig.versionNumber` is `None` for those — they can't be executed via this API), and use the one with the highest `versionNumber`.
- Added an optional `deployment_id` parameter so a caller who already has one can skip the lookup entirely.
- If no real (pinned) deployment exists yet, return a clear, actionable error instead of attempting execution with a value that will fail anyway — points at `manage_deployment(action='create')`.
- Fully backward compatible: no change to any existing required argument.

## Testing

- Added 3 new unit tests covering: normal auto-resolution, HEAD-deployment skipping / highest-version selection, explicit `deployment_id` bypassing the lookup, and the no-deployment error path.
- `uv run pytest` — full suite: 2117 passed, 2 skipped.
- `uv run ruff check` — clean.

## Summary by CodeRabbit

* **New Features**
  * Script functions can run against a specified deployment.
  * Automatic discovery searches all deployment pages and considers only versioned API-executable deployments.
  * When multiple eligible deployments exist, the highest-version deployment is selected automatically.
  * Clear guidance is provided when no suitable deployment is available.

* **Bug Fixes**
  * Prevents execution against unpublished, unpinned, or non-API-executable deployments.


taylorwilsdon:
Thanks!

PR #1121, opened 2026-09-13
fix(gappsscript): list_script_processes fails when script_id is passed

CapitalX:
## Summary
- `_list_script_processes_impl` passes `script_id` to `processes().list()` as a top-level `scriptId` kwarg, but the Apps Script API's `processes.list` method has no such parameter — script-scoped filtering is exposed as the nested `userProcessFilter.scriptId` query param (confirmed against the live discovery doc at `https://script.googleapis.com/$discovery/rest?version=v1`).
- Calling `list_script_processes(..., script_id="...")` therefore raises `TypeError: ... got an unexpected keyword argument 'scriptId'` from the generated googleapiclient method every time.
- Fix: build the request with `userProcessFilter.scriptId` instead. No need to switch to `processes.listScriptProcesses()` — `list()` already supports script-scoped filtering once the parameter name is correct.

## Test plan
- [x] Added `test_list_script_processes_with_script_id`, asserting the mock is called with `userProcessFilter.scriptId` (and never a bare `scriptId`).
- [x] `pytest tests/gappsscript -q` — 39 passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed script process filtering so requests using a script ID are accepted correctly.
  - Preserved page size settings when retrieving filtered script processes.
  - Ensured requests without a script ID continue to retrieve processes through standard listing behavior.

- **Tests**
  - Added coverage for automatic deployment discovery, version selection, explicit deployment handling, and missing-deployment errors.
  - Added validation for filtered and unfiltered process listings against the API schema.


taylorwilsdon:
Thanks!

PR #1132, opened 2026-09-14
fix(gmail): render bare newlines in HTML bodies as <br>

midna-agent:
## Summary

`send_gmail_message` / `draft_gmail_message` (and the forward path) take `body_format="html"` and put the body into the `text/html` part verbatim. LLM callers very often pass paragraphs separated by bare `\n` with no block markup at all — an HTML renderer collapses that whitespace, so the recipient gets one run-on paragraph. In our deployment this was the single most common way an outgoing email came out wrong, and prompt instructions ("always use `<p>`, never `\n`") did not reliably prevent it.

This PR makes the server tolerant instead of relying on caller discipline:

- New helper `gmail_helpers.html_newlines_to_br()` turns bare newlines that separate *content* (text, or inline tags such as `<b>`/`<a>`) into `<br>`. Newlines that merely sit next to a block-level tag (`<p>`, `<li>`, `<div>`, `<blockquote>`, …) are formatting whitespace and are left alone, so a well-formed HTML body — including the Gmail signature that `include_signature` appends — comes back byte-identical. Runs of newlines cap at `<br><br>`. Bodies containing `<pre>` are never touched.
- `_prepare_gmail_message()` applies it for `body_format="html"` **before** deriving the `text/plain` alternative, so send, draft and forward all benefit and the plain-text fallback keeps real line breaks.
- `body_format="plain"` is unchanged.

Helper logic lives in `gmail_helpers.py` per CONTRIBUTING; the tool file only gains the call.

## Test plan

- [x] New `tests/gmail/test_html_newlines_to_br.py`: helper edge cases (bare paragraphs, well-formed/compact HTML untouched, newline inside `<p>`, inline tags, appended signature, newline runs, `<pre>`, no-op inputs) and MIME-level checks of `_prepare_gmail_message` for html and plain bodies.
- [x] `uv run ruff check && uv run ruff format --check` clean.
- [x] `uv run pytest`: 2161 passed, 2 skipped.
- [x] Verified end-to-end against a live Gmail account: an html body with bare newlines and no tags arrives with the paragraphs intact and the signature unchanged.

## Summary by CodeRabbit

* **New Features**
  * Improved Gmail HTML formatting by converting meaningful line breaks into visible breaks in sent messages, drafts, and forwarded notes.
  * Preserved existing HTML structure, preformatted content, and plain-text line breaks.
* **Bug Fixes**
  * Prevented formatting whitespace around block elements, indentation, and raw-text content from creating unintended visible breaks.
  * Limited consecutive converted line breaks to avoid excessive spacing.
  * Ensured signatures and quoted original messages retain their existing formatting.


taylorwilsdon:
Thanks!

PR #1144, opened 2026-09-19
Add optional TrustModel AgentCert verify-gate (off by default, shadow mode)

karlmehta:
## Description

**The problem.** MCP servers increasingly receive tool calls from autonomous agents, but there's no standard way for a server to know *who* is calling or whether that agent is trustworthy. As the number of agents grows, that gap widens — today a tool call carries no verifiable identity or trust posture.

**What this PR adds.** An optional hook to verify a calling agent's AgentCert + TrustScore on each tool call:
- Registered alongside the existing `AuthInfoMiddleware` / `CamelCaseArgumentsMiddleware`.
- No-op unless `TRUSTMODEL_VERIFY=1` (existing users see zero change). Shadow mode by default (logs a verdict, never blocks); `TRUSTMODEL_MODE=enforce` rejects unverified/revoked callers.
- Reads request headers only — never tool arguments or Workspace data.
- Verification client is an **optional dependency** (`workspace-mcp[trustmodel]`), not installed for existing users. Fails closed at startup in enforce mode if the dep is missing.
- One small module (`core/trustmodel_verify.py`) + one registration line.

**Validation.** The verifier is open source (MIT, [pdxlab/agentcert-tag](https://github.com/pdxlab/agentcert-tag)) and published on npm + PyPI. Per TrustModel's website it already assigns a TrustScore to 19,400+ agents (growing toward 215,000) — a live, populated system, not a placeholder.

## Type of Change

## Testing
- Verified the middleware attaches against `fastmcp>=3.4.7` and is a no-op when the flag is unset.
- Verified enforce mode raises at startup when the optional dependency is absent (fails closed).

## Checklist

## Additional Notes
Placement/naming is easy to change to match your conventions.


karlmehta:
Thanks for the review. Addressed the enforce-mode finding: if `TRUSTMODEL_MODE=enforce` and the optional `trustmodel-agentcert-tag` dependency is missing, the server now **fails closed at startup** (raises) instead of starting unprotected; shadow mode keeps the warn-and-skip behavior. Also added docstrings and reformatted the description to the repo template. Small, opt-in, and off by default — glad to adjust anything to fit your conventions.

